PR #4, opened 2026-06-23
Fix: security hardening — Dependabot overrides + CodeQL findings

chrisdhaan:
## Summary

- **Dependabot overrides** — pin four transitive dependencies to patched versions via `package.json` `overrides`:
  - `undici` ≥ 7.28.0 (proxy cookie handling CVE)
  - `hono` ≥ 4.12.25 (header parsing CVE)
  - `js-yaml` ≥ 4.2.0 (YAML unsafe load)
  - `@opentelemetry/core` ≥ 2.8.0 (prototype pollution)
- **CodeQL `js/polynomial-redos`** — split alternating regex `/^_+|_+$/g` into two sequential anchored calls in `normalizeDeveloperName()` (`src/config.ts`)
- **CodeQL `js/incomplete-sanitization`** — escape backslashes before double-quotes in shell path quoting in `generateHookEntries()` (`src/install/install-helper.ts`)
- **CodeQL `js/path-injection` ×2** — add explicit null-byte + `..` component guard and MIME extension allow-list gate in `createStaticHandler()` (`src/dashboard/routes/static-handler.ts`)

## Test plan

- [ ] `npm run build` passes
- [ ] `npm test` — 3467 tests pass, 120 suites, 0 failures
- [ ] `npm run lint` — 0 errors, 0 warnings
- [ ] New test: MIME allow-list returns 403 for `.yaml` extension
- [ ] New test: null-byte in path returns 403
- [ ] New test: backslash in path is correctly double-escaped before quote-escaping

PR #5, opened 2026-06-23
Fix: use string-literal '/' in path-containment guard to resolve CodeQL alert

chrisdhaan:
## Summary

- Rewrites the path-containment guard in `static-handler.ts` from `root + sep`
  to `root + '/'` (string literal).
- CodeQL's `js/path-injection` sanitizer pattern-matches on `startsWith(root + '/')`;
  it cannot statically prove that the runtime `path.sep` variable equals `'/'`,
  so the previous guard was not being recognised and both the `stat()` and
  `readFile()` calls were flagged (alerts #3 and #4).
- Removes the dead `&& target !== root` branch — empty-filename URLs (e.g. `//`)
  now correctly return 403 instead of falling through to a directory `stat`.

## Test plan

- [ ] `npm test` — all tests pass
- [ ] CodeQL scan shows alerts #3 and #4 (`js/path-injection` in `static-handler.ts`) closed

### DIFF
--- a/src/dashboard/routes/static-handler.ts
+++ b/src/dashboard/routes/static-handler.ts
@@ -61,7 +61,11 @@ export function createStaticHandler(
     }
 
     const target = resolve(join(root, filename));
-    if (!target.startsWith(root + sep) && target !== root) {
+    // Use a literal '/' so static analysis tools (CodeQL js/path-injection)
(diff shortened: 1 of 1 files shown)


corinnekunze:

PR #7, opened 2026-06-23
Fix: local dashboard auto-starts without manual intervention (#167)

chrisdhaan:
**Problem**

Users who chose local or both mode during preflight setup found the dashboard didn't work — they had to manually run preflight --local in a separate terminal to get it going. Two root causes:

1. The MCP server (preflight --stdio) does start the dashboard when mode: local/both is configured, but only after async session ID resolution completes. On some machines this took long enough that users gave up.
2. There was no persistent background process — the dashboard only ran while Claude Code had a session open.

---

**What's new**

**Always-on background dashboard (macOS)**

The setup wizard now offers to install a launchd agent (com.preflight.dashboard) when mode is local or both:

Install always-on background dashboard? Keeps the local dashboard running
even when Claude Code is closed [Y/n]:

The daemon starts automatically at login, survives reboots, and hands the dashboard port over to active Claude Code sessions when they start — then reclaims it when they end. preflight uninstall removes it.

**Dashboard starts immediately**

When a Claude Code session begins, the dashboard is now live within milliseconds rather than waiting for session ID resolution to complete.

---

**Bug fixes**

- Dashboard daemon correctly survives port contention — previously it would exit immediately on EADDRINUSE instead of waiting to take over
- Setup wizard daemon upgrade is now atomic — reinstalling over an existing daemon no longer risks leaving the user with no daemon if the process fails partway through
- Session resolution polling is aborted cleanly on shutdown (no orphaned timers)
- OTel users: provisional session spans no longer emit a zero-call ghost span to OTLP backends
- isSyntheticSessionId now covers provisional session IDs, preventing them from appearing in audit records or the dashboard session history

---
Test plan

- [ ] npm run build && npm test — all tests pass
- [ ] preflight setup with mode local → wizard prompts for daemon install at Step 6b
- [ ] After install: launchctl list | grep preflight shows com.preflight.dashboard with a PID
- [ ] Dashboard accessible at http://127.0.0.1:7777 without an active Claude Code session open
- [ ] preflight uninstall removes the daemon (launchctl list | grep preflight returns nothing)
- [ ] Re-running preflight setup over an existing daemon upgrades it without error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.2] - 2026-06-23
+
+### Added
+
+- Always-on background dashboard daemon for macOS (`local` and `both` modes). The setup wizard now offers to install a launchd agent (`com.preflight.dashboard`) that keeps the local dashboard running at `http://127.0.0.1:7777` even when Claude Code is closed. `preflight uninstall` removes it.
+
+### Fixed
+
+- Local dashboard now starts immediately when `--stdio` mode launches with `mode: local` or `mode: both` — no longer requires manually running `preflight --local` after a Claude Code session starts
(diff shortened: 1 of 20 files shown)


corinnekunze:

PR #18, opened 2026-07-02
Chore: v1.0.7 — uninstall reliability, doctor improvements, daemon stability

chrisdhaan:
## Summary

This release focuses on reliability and correctness across uninstall, the `preflight doctor` command, and the background dashboard daemon.

### Uninstall improvements

- **`preflight uninstall --daemon`** — new flag that removes only the background dashboard daemon (launchd plist) without touching hooks, MCP config, schedule, or session history. Useful when you just need to stop a broken or unwanted daemon.
- **Confirmation prompt** — bare `preflight uninstall` now lists every file and plist that will be affected and explicitly notes that session history will _not_ be deleted, before asking `Continue? [y/N]`.
- **`--yes` flag** — skips the interactive prompt for use in scripts and CI pipelines.
- **Non-TTY detection** — exits cleanly with code 1 when stdin is not a terminal and `--yes` is not passed, instead of hanging.
- **Correct exit codes** — cancellation (interactive `n` or non-TTY without `--yes`) exits 1 to match script expectations. Partial failure (e.g. a permission error mid-removal) also sets exit code 1.
- **TOCTOU safety** — handles the case where a file is externally removed between the confirmation check and the actual removal.

### `preflight doctor` improvements

- **Daemon-not-installed is now a warning, not a failure** — the always-on dashboard daemon is an optional feature; not having it installed is not a problem.
- **Correct Homebrew node path detection** — the doctor check now validates against the stable symlink directory (e.g. `/opt/homebrew/bin`) rather than the resolved Cellar path that `brew upgrade node` removes.
- **Pre-v1.0.6 plist compatibility** — if you installed the daemon before the PATH injection fix landed, the doctor check now shows `warn + reinstall guidance` instead of a spurious `fail`.
- **NR reachability skipped when config is absent or invalid** — prevents the misleading double-failure (config missing + NR unreachable) on first-time setups.
- **4xx responses surfaced as warnings** — a wrong or expired license key on the NR reachability check is now reported as `warn` instead of silently counted as `ok`.
- **Accurate skip messages** — "no config file — run `preflight setup` first" vs. "config errors present" are now distinct messages.
- **EACCES vs ENOENT for node binary** — a node binary that exists but is not executable gets its own diagnostic message and fix advice, distinct from a missing binary.

### Daemon stability

- **`ThrottleInterval: 300`** added to the launchd plist — prevents the daemon from spinning at high CPU in a tight crash loop if node is missing or the binary is unreachable.
- **Daemon prompt changed from `[Y/n]` to `[y/N]`** — installing a persistent launchd service is now explicitly opt-in rather than the default.
- **`launchctl remove <label>` fallback** — if `launchctl unload <path>` fails (e.g. plist is unreadable), the job is removed by label to prevent an orphan surviving until reboot.

## Test plan

- [ ] `npm run build && npm test` passes
- [ ] `preflight uninstall` shows confirmation prompt listing affected files
- [ ] `preflight uninstall --yes` completes without prompting
- [ ] `preflight uninstall --daemon` removes only the daemon plist
- [ ] `preflight doctor` shows `warn` (not `fail`) when daemon is not installed
- [ ] CI passes on public repo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,52 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.7] - 2026-07-01
+
+### Added
+
+- **`preflight uninstall --yes`** — skips the interactive confirmation prompt for scripted and CI use
+- **`preflight uninstall --daemon`** — removes only the background dashboard daemon (plist + `launchctl unload`) without touching hooks, MCP config, schedule, or session history
+- `preflight uninstall` now prints a full summary of exactly what will change before prompting — which hooks files, MCP config files, schedule plist, and daemon plist will be touched, with exact paths
+
+### Changed
+
+- Background dashboard daemon installation during `preflight setup` is now **opt-in** (`[y/N]`) rather than opt-out (`[Y/n]`). The daemon is a persistent `launchd` service and should require explicit consent to install. Existing installations are unaffected — only new setups or re-runs of `preflight setup` are impacted.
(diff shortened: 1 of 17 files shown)


corinnekunze:

PR #19, opened 2026-07-02
Fix + Docs: v1.0.8 — preflight update reliability + tool selection score guidance (#185)

chrisdhaan:
### Fix: `preflight update` now works correctly for package manager installs

Previously, running `preflight update` on a copy of preflight installed via `npm install -g`, `pnpm add -g`, or as a local `node_modules` dependency would fail with a cryptic `not a git repository` error from git. The command is only meaningful for source clones, where it runs `git pull` + `npm run build`.

This release fixes that with a two-step detection:

1. If `git` is not on PATH, a clear installation link is shown.
2. If `git --show-toplevel` succeeds but `node_modules` appears in the relative path from the git root to the install directory, preflight recognises the install as a package manager dependency and prints the correct reinstall command instead.

The check uses `path.relative()` for cross-platform path comparison, so it works correctly on Windows.

**Before:**
```
$ preflight update
fatal: not a git repository (or any of the parent directories): .git
```

**After:**
```
$ preflight update
✗ preflight was installed via a package manager, not cloned from source.
  To update, reinstall using your package manager, e.g.:
    npm install -g @newrelic/preflight@latest
    pnpm add -g @newrelic/preflight@latest
```

### Fix: clearer error when `git pull` fails with diverged branches

When `git pull` fails (e.g. because you have local commits that have diverged from the remote), the error output now includes a ready-to-run reset command:

```
✗ git pull failed. Check the output above for details.
  If the output shows diverged branches and you have no local commits to keep,
  you can reset to the remote HEAD (replace <branch> with your default branch):
    git -C "/path/to/preflight" fetch origin
    git -C "/path/to/preflight" reset --hard origin/<branch>
  WARNING: reset --hard permanently discards any local commits not yet on origin.
```

The try/catch was also split so the diverged-branch hint only shows when `git pull` specifically fails, not when the subsequent build step fails.

### Docs: tool selection score guide in `docs/ADVANCED.md`

A new **Improving Your Tool Selection Score** section explains the three penalty categories reported by `nr_observe_get_tool_selection_score`, with concrete before/after prompt examples for each:

- **Redundant reads** — reading the same file 3+ times without an edit in between; fix by front-loading context in your prompt
- **Repeated failures** — consecutive failures of the same tool; fix by providing corrective context instead of letting the AI retry identically
- **Unused large outputs** — tool responses ≥ 4 KB that are never acted on; fix by using targeted grep/bash lookups rather than reading full files

`docs/COMMANDS_TABLE.md` is updated to cross-reference the new section.

## Test plan

- [x] All 3,663 tests pass
- [x] Package manager install detection tested: global npm install, local `node_modules` dependency, source clone inside a path containing a `node_modules` ancestor (false-positive regression)
- [x] `git not on PATH` (ENOENT) path tested
- [x] Diverged-branch and build-failure error paths tested with distinct assertions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.8] - 2026-07-02
+
+### Fixed
+
+- **`preflight update` with package manager installs** — when preflight is installed via a package manager (global `npm install -g`, `pnpm add -g`, or a local `npm install` into `node_modules`), running `preflight update` now prints a clear reinstall command instead of failing with a cryptic `not a git repository` error from git
+- **`preflight update` divergent branches** — when `git pull` fails (e.g. local commits ahead of remote), the error output now includes a `git fetch origin && git reset --hard origin/<branch>` command (replace `<branch>` with your default branch name) rather than a generic "check the output above" message
+
+### Changed
+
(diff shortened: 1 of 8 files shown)


corinnekunze:

Issue #17, opened 2026-07-02
WSL2 (Linux CC): session_id never resolves via official hook alone; harvest scheduler never sends to New Relic even after fixing that


juki330:
## Environment

- WSL2 (Ubuntu), Claude Code installed via npm inside WSL (Linux CC, not the Windows desktop app) — `platformTarget: wsl-linux-cc`
- Shell: fish
- `@newrelic/preflight` 1.0.5 and 1.0.6 (reproduced on both)
- Tried both manual config and official `preflight install --linux-cc --license-key ... --account-id ...`

## Bug A: session_id never resolves with the official hook alone

With only the installed hook command present (`preflight-collector pre-tool` / `post-tool`), `nr_observe_health` reports `"session_id": null` indefinitely, well past the internal 60s timeout:

```
{"level":"warn","component":"session-resolver","message":"session_id unresolved after 60s — breadcrumb missing; check that hook collector is installed and writing."}
```

Root cause: `dist/hooks/collector-script.js` writes its breadcrumb under its own `process.ppid`, assuming that's the same PID the `--stdio` MCP server polls (`session-by-ppid/<ppid>.txt`). That only holds if the hook is spawned as a **direct** child of `claude`. On this WSL + fish setup, Claude Code's hook runner interposes an intermediate `sh`, so the collector's `ppid` is that shell's PID, not `claude`'s — the breadcrumb is written under the wrong key and is never found.

Workaround that fixes it: a wrapper script that walks up `/proc/<pid>/stat` ancestors to find the real `claude` PID and writes the breadcrumb there before calling the collector. With that in place, `session_id` resolves correctly.

## Bug B: harvest scheduler never sends anything to New Relic, even with a resolved session_id and a valid key

With the workaround above (session_id resolves and matches the real session) and a license key verified valid directly against the Events API (`curl -X POST https://insights-collector.newrelic.com/v1/accounts/<id>/events -H "X-Insert-Key: <key>"` → `200 {"success":true}`):

- `"Harvest scheduler started"` (logged by `HarvestScheduler.start()`) never appears in any log, across many restarts, over 24+ hours.
- `ss -tnp` on the running `--stdio` process, while actively generating tool-call events, shows zero outbound connections to New Relic's ingest host (`insights-collector.newrelic.com` = `162.247.243.33`).
[...]

PR #22, opened 2026-07-03
Chore: v1.0.10 — WSL2 ancestor PID walk, symlink version fix, doctor warn fix

chrisdhaan:
## Summary

Related to #17 

- **WSL2 + fish shell session tracking** — hook collector now walks `/proc/<pid>/stat` up to 5 levels deep to find ancestor PIDs, so the MCP server can match sessions even when intermediate shell processes sit between Claude Code and the collector. macOS/Windows behaviour unchanged.
- **Version reporting when installed as a symlink** — `preflight --version` now resolves symlinks before locating `package.json`, fixing `0.0.0` version output when installed via `npm link` or a package manager.
- **`preflight doctor` warn-not-fail for custom hook wrappers** — a hook event type with a non-preflight command now reports a warning with fix guidance rather than a failure; a hook event type with no command at all still fails.

## Test plan

- [ ] Full test suite passes (`npm test`)
- [ ] Build clean (`npm run build`)
- [ ] Lint clean (`npm run lint`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.10] - 2026-07-03
+
+### Fixed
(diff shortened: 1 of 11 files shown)


corinnekunze:

PR #26, opened 2026-07-07
Fix: v1.1.1 — license key region routing fix for unrecognized prefixes

chrisdhaan:
## Summary

- `resolveRegion()` now warns and falls back to `'us'` instead of throwing when a license key has an unrecognized region-shaped prefix (e.g. `ca06...NRAL`) — fixes a startup crash for users on newly launched or unsupported regions
- FQDN `collectorHost` returns early via the existing `isLiteralHostname()` helper, eliminating a duplicate inline predicate and suppressing spurious warn for FQDN-override users
- Bumps version to 1.1.1 and adds CHANGELOG entry

## Test plan

- [x] `src/shared/transport/http-client.test.ts` — 45 tests pass (warn-based assertions replacing prior throw-based assertions)
- [x] Full test suite — 3730 tests pass
- [x] Build clean, lint clean, format clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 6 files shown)


corinnekunze:

PR #80, opened 2026-07-08
Fix: v1.3.1 — Windows dashboard/hook fixes

### LINKED ISSUE #24: Dashboard returns 403 Forbidden on Windows — hardcoded / in static-handler path containment check
## Summary

The dashboard UI is completely inaccessible on Windows (`403 Forbidden` on every static file request), while API endpoints like `/api/health` work fine.

## Root Cause

In `dist/dashboard/routes/static-handler.js`, the path containment check uses a hardcoded Unix separator:

```js
if (!target.startsWith(root + '/')) {
    res.writeHead(403);
    res.end();
    return;
}
```

On Windows, `path.resolve()` produces backslash-separated paths (e.g. `C:\Users\...\dist\web\index.html`), so `root + '/'` never matches and all static file requests are rejected.

## Steps to Reproduce

1. `npm install -g @newrelic/preflight` on Windows
2. `preflight --local`
3. Open `http://127.0.0.1:7777` in a browser → 403

## Suggested Fix

```js
import { sep } from 'node:path';
// ...
if (!target.startsWith(root + '/') && !target.startsWith(root + sep)) {
```

## Environment

- OS: Windows 11
- Node: v24.8.0
- preflight: v1.1.0

chrisdhaan:
## Summary

Fixes two Windows-only bugs reported in #24 — thanks @sgwrmctk for the detailed report and suggested fixes.

- **Dashboard 403 on every static asset request.** The static file handler's root-containment check only recognized `/` as a path separator, so it never matched on Windows, where `resolve()`/`join()` produce backslash-joined paths.
- **Hook collector silently dropped all session data.** It read stdin via the POSIX-only `/dev/stdin` device path, which doesn't exist on Windows, so every `PreToolUse`/`PostToolUse` hook failed silently with no error surfaced anywhere.

## Changes

- `static-handler.ts`: extracted `isWithinRoot(root, target, platformSep = sep)` — checks the literal `'/'` first, then falls back to the platform separator. `platformSep` is overridable so the Windows branch is testable deterministically on any OS.
- `collector-script.ts`: extracted `readStdinSync()` — reads via `process.stdin.fd` on `win32`, keeps `/dev/stdin` on POSIX.

## Test plan

- [x] `npm run build && npm test && npm run lint` all pass locally
- [x] New regression tests for both platform branches (fabricated Windows-style path strings / mocked `process.platform`, not dependent on the CI runner's actual OS)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.3.1] - 2026-07-08
+
+### Fixed
+
+- **Dashboard 403 on Windows** — the static file handler's root-containment check only recognized `/` as a path separator, so every static asset request (including `index.html`) was rejected on Windows, where `resolve()`/`join()` produce backslash-joined paths. Added a platform-separator fallback alongside the existing literal `/` check.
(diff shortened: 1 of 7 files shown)


corinnekunze:

PR #81, opened 2026-07-09
Feat: v1.4.0 — offer to restart stale dashboard/daemon processes on update

### LINKED ISSUE #79: preflight update: offer to restart stale dashboard/daemon processes
## Problem

`preflight update` (`handleUpdate()` in `src/install/cli.ts`) does `git pull` + `npm run build` and prints "Restart Claude Code to pick up the new version" — but it never touches already-running processes. `VERSION` (`src/version.ts:28`) is read from `package.json` once, at module-load time, and cached for the life of the process, so any already-running dashboard/server keeps reporting the old version forever, no matter how many times you refresh the browser or re-run `update`.

Concretely hit this after updating: the dashboard UI kept showing v1.1.0 in the sidebar (`Sidebar.tsx` -> `useVersionInfo` -> `GET /api/health` -> `dashboard-server.ts`'s cached `VERSION`) even though `package.json` was already at v1.3.0 on disk, because a `node dist/index.js --local` process from days earlier was still alive and serving the version it loaded at its own startup.

## Proposal

Give `preflight update` (or a follow-up prompt after it finishes) an interactive "question experience" -- similar to the existing `node:readline/promises` prompts already used in `setup-wizard.ts` (e.g. "Install Claude Code hooks now? [Y/n]") -- that detects processes affected by the update and offers to restart them, instead of silently leaving them stale or requiring the user to go hunt down PIDs manually.

This needs to handle three distinct run modes differently:

1. **`--stdio` (MCP mode, spawned by Claude Code)** -- do NOT auto-kill. These are child processes actively serving the current conversation; killing one out from under Claude Code mid-session breaks tool calls. Keep the existing "Restart Claude Code" message for this case.
2. **Dashboard daemon (macOS launchd, `installDashboardDaemon`/`removeDashboardDaemon`/`getDashboardDaemonStatus` in `src/install/schedule.ts`)** -- if `getDashboardDaemonStatus().installed`, this is safe and clean to restart via `launchctl unload` + `launchctl load` on the existing plist. Good candidate for an automatic (or Y/n-prompted) restart.
3. **Ad-hoc `node dist/index.js --local` processes** started manually in a terminal -- messiest case. Matching by cmdline via `ps`/`tasklist` is not very cross-platform-friendly (this repo already needs to support Windows) and risks touching a process the user intentionally left running for another purpose (e.g. multiple projects/ports). Needs more design: possibly match by port only, or skip auto-detection here entirely and just document it.

## Open questions for the follow-up PR

- Should detection be cross-platform (Windows `tasklist`/`wmic` equivalent to `ps`), or macOS/Linux only initially given the daemon feature is already macOS-only?
- Should the prompt default to Y or N for killing ad-hoc `--local` processes, given the risk of surprising the user?
- Is matching by port (query a known/expected list of ports) safer than matching by full cmdline?

## Context

Originally surfaced while investigating why the dashboard kept showing v1.1.0 after fixing #24 (Windows path/stdin fixes, PR #206, released in v1.3.1).

chrisdhaan:
Fixes #79

## Summary
- `--local` now writes a small PID file (`~/.newrelic-preflight/local-dashboard.pid`: `{pid, argv, cwd}`) the moment it wins the dashboard port bind, and removes it on graceful shutdown — giving `preflight update` a reliable way to identify its own dashboard process.
- `preflight update` now offers to restart stale processes after a successful build:
  - Auto-restarts the macOS launchd dashboard daemon (no prompt) if installed.
  - Otherwise, if the PID file names a live ad-hoc `--local` process, prompts default-yes `[Y/n]` to kill (SIGTERM → 2s grace → SIGKILL) and respawn it detached with its original argv/cwd.
  - `--stdio` (Claude Code MCP sessions) is never touched — it keeps the existing static "Restart Claude Code" message.

## Test plan
- [x] `npm ci && npm run build && npm test` — 3786/3789 passing (3 pre-existing skips), clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.0] - 2026-07-08
+
+### Added
+
+- **`preflight update` offers to restart stale dashboard/daemon processes** — after a successful `git pull` + rebuild, `--local` now writes a small PID file (`local-dashboard.pid`: pid/argv/cwd) the moment it wins the dashboard port bind, giving `update` a reliable way to find its own dashboard process instead of a running instance silently continuing to serve the old cached version. If the macOS launchd dashboard daemon is installed, it's restarted automatically (no prompt). Otherwise, if a live ad-hoc `--local` process is found, `update` prompts (default yes) to restart it — killing it gracefully and respawning it detached with its original arguments. `--stdio` (Claude Code) sessions are never touched; they keep the existing "restart Claude Code" guidance.
+
 ## [1.3.1] - 2026-07-08
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ You'll need a **license key** (telemetry ingest) and your **account ID**, plus a
 ```bash
 preflight doctor               # Run 6 diagnostic checks and print actionable fix commands
 preflight validate             # Check config for syntax errors and unknown keys
-preflight update               # Pull latest version and rebuild (source installs only — npm installs: npm install -g @newrelic/preflight@latest)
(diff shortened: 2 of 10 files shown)


corinnekunze:

PR #82, opened 2026-07-09
Fix: v1.4.1 — verify dashboard restarts actually succeed before reporting success

### LINKED ISSUE #83: preflight update: restart-offer can report success when the restart actually failed
## Problem

`preflight update`'s restart-offer feature (added in v1.4.0, #79) declares a dashboard/daemon restart successful purely based on the restart *action* not throwing — `launchctl load` returning a zero exit code, or the respawned ad-hoc process's `spawn()` call not throwing synchronously. Neither path verifies that the resulting process actually stays alive and starts serving the dashboard.

Concretely hit this in practice: `launchctl load` returned success and `preflight update` printed `✓ Restarted the dashboard daemon (launchctl unload/load).`, but the actual daemon process crashed a moment later (in this case, from a macOS Full Disk Access / TCC restriction unrelated to the restart logic itself). The terminal reported success while the dashboard was actually still down, and a stale process was left squatting on the port with the daemon assumed to be handling it.

A second, related gap: because `offerRestarts()` treated "daemon installed" as unconditional proof that the daemon owns the dashboard port, it never fell back to checking for a live ad-hoc `--local` process when the daemon restart's success couldn't actually be trusted — so a genuinely stale ad-hoc process went completely unnoticed and unoffered for cleanup.

## Fix

Add a shared post-restart verification step: poll `GET /api/health` for a healthy response whose reported version matches the freshly-rebuilt local `package.json` version, applied to both the daemon-restart path and the ad-hoc-respawn path. If the daemon path can't be verified, fall through to the ad-hoc PID-file check instead of returning unconditionally. Every new failure mode (config load, version read, fetch) degrades to a printed warning or a silent skip — never a new way for `update` to crash.

See #82.

chrisdhaan:
Fixes #83

## Summary
- `preflight update`'s restart-offer feature previously declared restart success purely based on the restart *action* not throwing (`launchctl load` exit code, or `spawn()` not throwing synchronously) — never verifying the resulting process actually stayed alive and served the dashboard.
- Adds a shared post-restart verification step: poll `GET /api/health` (Node's built-in global `fetch`, no new dependency) for a healthy response whose reported version matches the just-rebuilt local `package.json` version — applied to both the daemon-restart path and the ad-hoc-respawn path.
- If daemon verification fails, execution now falls through to check the ad-hoc PID-file path instead of returning unconditionally.
- Every new failure mode (config load, version read, fetch, JSON parse) degrades to a printed warning or a silent skip — never a new way for `update` to crash or exit non-zero.

## Test plan
- [x] `npm run build && npm test` — 3790/3793 passing (3 pre-existing skips), clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.1] - 2026-07-08
+
+### Fixed
+
+- **`preflight update` could report a restart succeeded when it hadn't** — both restart paths (the macOS launchd daemon and an ad-hoc `--local` process) declared success as soon as the restart _action_ didn't throw (`launchctl load` returning success, or the respawned process not throwing synchronously), without checking that the dashboard actually came back up. A daemon that crashed immediately after a "successful" `launchctl load` — for example due to a macOS Full Disk Access restriction — would print a false "restarted" message while a stale process kept serving the old version. `update` now polls the dashboard's health endpoint for a healthy response reporting the freshly-built version before declaring success, and falls back to checking for an ad-hoc process if the daemon restart can't be verified.
+
 ## [1.4.0] - 2026-07-08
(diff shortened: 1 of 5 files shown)


corinnekunze:

PR #85, opened 2026-07-09
Feat: v1.4.2 — preflight local command to find and clean up orphaned processes

### LINKED ISSUE #86: preflight update: orphaned --local processes accumulate with no way to detect or clean them up
## Problem

`preflight update`'s restart-offer feature (added in v1.4.0, #79) only ever tracks whichever `--local` process currently wins the dashboard port bind (`local-dashboard.pid`, a single-slot file). Any `--local` process that loses that EADDRINUSE race just keeps running headless forever, with zero tooling to detect or clean it up.

Concretely hit this in practice: a live investigation found 5 stray orphaned `--local` processes accumulated over several days (from repeated Playwright/manual test runs that weren't cleanly terminated), discoverable only via manual `ps`/`lsof`/`kill` — nothing in the codebase could see them.

## Fix

Add a per-PID instance registry (`local-instances/<pid>.json`: pid/argv/cwd/startedAt), written by every `--local` process at startup regardless of port outcome, and removed on clean shutdown — purely additive, doesn't touch the existing single-slot `local-dashboard.pid` mechanism.

A new `preflight local` command lists every registered instance and marks the current dashboard owner; `preflight local --clean` prompts (single combined `[Y/n]`, default yes) to kill every live instance that isn't the owner. `preflight doctor` gets a matching read-only check pointing at `preflight local --clean`. Dead registry entries are garbage-collected automatically every 5 minutes.

Never scans the OS process table — every action only ever targets a PID that preflight itself wrote into the registry.

See #85.

chrisdhaan:
Fixes #86

## Summary
- Orphaned `--local` processes had zero visibility before this: the restart-offer feature only ever tracks whichever `--local` process wins the dashboard port bind, so any process that loses that race just runs headless forever with no way to detect it.
- Every `--local` process now registers itself in a small per-PID registry (`local-instances/<pid>.json`: pid/argv/cwd/startedAt) at startup, regardless of port outcome, and removes itself on clean shutdown. This is purely additive — the existing single-slot `local-dashboard.pid` mechanism is untouched.
- New `preflight local` command: lists every registered instance and marks the current dashboard owner. `preflight local --clean` prompts (single combined `[Y/n]`, default yes) to kill every live instance that isn't the owner.
- Dead registry entries (from a process that didn't shut down cleanly) are garbage-collected automatically every 5 minutes by whichever process owns the dashboard.
- `preflight doctor` gained a matching read-only "Local instances" check that reports idle processes and points at `preflight local --clean`.
- Never scans the OS process table anywhere — every kill/list/report action only ever acts on a PID that preflight itself wrote into a registry file.

## Test plan
- [x] `npm run build && npm test` — 3811/3814 passing (3 pre-existing skips), clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.2] - 2026-07-09
+
+### Added
+
+- **`preflight local` command** — lists every `--local` dashboard process preflight has launched (not just whichever one currently owns the dashboard port), and `preflight local --clean` offers to kill the ones that lost the port race and have been running headless ever since. Every `--local` process now registers itself in a small per-PID registry at startup regardless of port outcome; dead entries (from a process that didn't shut down cleanly) are garbage-collected automatically every 5 minutes by whichever process owns the dashboard. `preflight doctor` gained a matching "Local instances" check that reports idle processes and points at `preflight local --clean`.
+
 ## [1.4.1] - 2026-07-08
 
 ### Fixed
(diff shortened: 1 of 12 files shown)


corinnekunze:

PR #87, opened 2026-07-09
Fix: make Release workflow idempotent and resilient to npm publish flakiness

chrisdhaan:
## Summary
- Today's manual Release run tagged v1.4.2 and created the GitHub release successfully, then failed on **Publish to npm** with `npm error code IDENTITY_TOKEN_READ_ERROR` (transient OIDC identity-token fetch failure).
- Re-running the workflow then failed immediately at **Create tag** / **Create release** since both already existed, blocking any retry from reaching npm publish.
- Makes each step idempotent so re-running Release after a partial failure always converges: `Create tag` and `Create release` skip if they already exist, and `Publish to npm` skips if the version is already on the registry and retries up to 3 times with backoff on transient failures.

## Test plan
- [x] `npm run build && npm test` pass locally
- [ ] Re-run the `Release` workflow via `workflow_dispatch` to confirm it publishes `@newrelic/preflight@1.4.2` to npm without re-creating the existing tag/release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,14 +45,38 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          git tag v$(node -p "require('./package.json').version")
-          git push --tags
+          TAG="v$(node -p "require('./package.json').version")"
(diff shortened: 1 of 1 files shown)


corinnekunze:

PR #88, opened 2026-07-09
Fix: v1.4.3 — Kiro hook adapter fixes

### LINKED ISSUE #84: Kiro hook events silently dropped — hook_event_name casing mismatch, adapter never wired
## Summary

Tool-call telemetry from Amazon Kiro never reaches the dashboard or New Relic, even when the MCP connection, hooks, and `preflight doctor` all report healthy. Traced to two compounding issues in `preflight-collector` / the platform-adapter system.

## Reported symptom

A user configured Preflight as an MCP server in Kiro (`npx -y @newrelic/preflight --stdio`) and wired `preToolUse`/`postToolUse` hooks to call `preflight-collector --pre` / `--post`, exactly per the documented setup. Claude Code, configured in parallel, worked fine. Kiro tool calls never appeared on `localhost:7777`, despite:
- the dashboard loading normally
- `preflight doctor` reporting all checks passed

### 1. `preflight-collector` hard-codes Claude Code's hook event names

`src/hooks/collector-script.ts` (`processHook`, ~line 630) does a strict equality check on the incoming `hook_event_name` field:

```ts
if (eventName === 'PreToolUse') { ... }
else if (eventName === 'PostToolUse') { ... }
else if (eventName === 'PostToolUseFailure') { ... }
else {
  // Unknown hook event — ignore silently
  return;
}
```

These are Claude Code's exact (PascalCase) hook event names. Per Kiro's own docs (https://kiro.dev/docs/cli/hooks), Kiro's hook payload sends `hook_event_name` as lower-camelCase: `"preToolUse"`, `"postToolUse"`, `"agentSpawn"`, `"userPromptSubmit"`.

`'preToolUse' !== 'PreToolUse'` — every Kiro hook event hits the `else` branch. Nothing is written to the buffer file, and the collector still exits `0`, so from Kiro's perspective the hook succeeded. No error surfaces anywhere.

### 2. The Kiro platform adapter exists but is never invoked

`src/platforms/kiro-adapter.ts` implements a full `KiroAdapter` (tool-name mapping via `KIRO_TOOL_MAP`, env-var-based detection of `KIRO_SESSION_ID`/`KIRO_IDE`/`MCP_CLIENT=kiro`/`NEW_RELIC_AI_PLATFORM=kiro`, `normalizeToolCall()`), and `src/platforms/platform-registry.ts` has real detection logic (`PlatformRegistry.detect()` / `getActive()`) that would select it.

However, grepping the whole `src/` tree shows `adapter.normalizeToolCall(...)`, `registry.detect()`, and `registry.getActive()` are called **only from unit tests** (`*.test.ts`). `src/index.ts` imports `createDefaultRegistry` but never calls it. The hook-based ingestion path (`collector-script.ts` → `HookEventProcessor` in `src/hooks/event-processor.ts` → trackers) never consults the platform registry or any adapter at all — it's entirely Claude-Code-shaped, regardless of which platform is actually connected.

So the README/CHANGELOG claim ("Preflight now detects and normalizes tool calls from Amazon Kiro") is true of the adapter *module* in isolation, but not of the running system. The adapter is fully-built, tested scaffolding that was never connected to the pipeline that actually populates the dashboard.

### 3. `preflight doctor` can't catch this

`checkHooksWired()` in `src/install/diagnostics.ts` only inspects Claude Code's `~/.claude/settings.json` for `PreToolUse`/`PostToolUse` entries invoking the collector. It has no visibility into Kiro's own hook config or into whether events are actually landing in the buffer. If Claude Code is *also* configured on the same machine (as in the reported case), doctor reports green regardless of whether Kiro's hooks work at all.

## Proposed fix

**Short-term (unblocks Kiro immediately):**
Make the `hook_event_name` check in `collector-script.ts` case-insensitive, or explicitly accept both Claude Code's and Kiro's event-name spellings:

```ts
const normalizedEventName = eventName?.toLowerCase();
if (normalizedEventName === 'pretooluse') { ... }
else if (normalizedEventName === 'posttooluse') { ... }
else if (normalizedEventName === 'posttoolusefailure') { ... }
```

**Correct long-term fix:** wire the platform-adapter system into the real ingestion path instead of leaving it as unused scaffolding:
1. In `collector-script.ts` or `event-processor.ts`, detect the active platform (via `createDefaultRegistry().getActive()`, gated on the Kiro/other env vars already implemented in each adapter's `isSupported()`) once per process, and use it to normalize the raw hook payload — event-name casing included — before writing/pairing.
2. Route Kiro's `tool_name` values (Kiro's canonical tool names, e.g. `fs_read`, `execute_bash` — confirmed against https://kiro.dev/docs/cli/hooks — plus aliases like `read`/`write`/`shell`) through `KIRO_TOOL_MAP` in `kiro-adapter.ts`, and double check that map's keys actually match Kiro's real tool names (some entries look like guesses, e.g. `fsRead`/`fsWrite` camelCase, that may not match Kiro's documented snake_case canonical names).
3. Extend `preflight doctor`'s "Hooks wired" check to be platform-aware — at minimum, warn when the detected platform isn't Claude Code and the Claude-Code-only settings.json check is the only thing being validated.
4. Add an integration test that feeds a real Kiro-shaped hook payload (`hook_event_name: "preToolUse"`, Kiro's documented field names) through `collector-script.ts` end-to-end and asserts a buffer line is written — the current adapter unit tests only exercise `normalizeToolCall()` directly and would not have caught this, since that method is never called by the actual collector.

## Environment

- Preflight: MCP stdio mode + hooks, both configured per the documented Kiro setup instructions from `KiroAdapter.getHookInstallInstructions()`
- Kiro hooks calling `preflight-collector --pre` / `preflight-collector --post`
- `preflight doctor`: all checks passed
- Claude Code, configured in parallel on the same machine, works correctly

## References

- Kiro hook payload schema: https://kiro.dev/docs/cli/hooks
- `src/hooks/collector-script.ts` (`processHook`)
- `src/platforms/kiro-adapter.ts`
- `src/platforms/platform-registry.ts`
- `src/install/diagnostics.ts` (`checkHooksWired`)

chrisdhaan:
## Summary

- Fixes Kiro IDE hook events being silently dropped: `preflight-collector` matched Claude Code's exact PascalCase hook event names (`PreToolUse`); Kiro sends lower-camelCase (`preToolUse`) per its own docs, so every Kiro hook event fell through to a silent no-op — nothing written to the buffer, collector still exited 0. The event-name check is now case-insensitive.
- Wires the platform-adapter system into the real hook pipeline: `PlatformAdapter`/`PlatformRegistry` were fully built and tested but never actually called from the running hook pipeline, so every platform's tool names passed through unmapped. Added `PlatformAdapter.mapToolName()`, wired it into `HookEventProcessor`, corrected `KIRO_TOOL_MAP` against Kiro's documented aliases (`read`/`write`/`shell`/`aws`), and made sure a tool's original name is preserved instead of collapsing into a generic `"Unknown"` bucket when a platform adapter has no mapping for it.
- Adds `preflight doctor --platform <name>` so the "Hooks wired" check isn't silently Claude-Code-only.

Closes #84

## Test plan

- [x] `npm run build && npm test && npm run lint` green (3843 passed, 3 skipped, 0 failed; 0 lint errors/warnings)
- [x] Manual smoke test: `preflight doctor --platform kiro` and `preflight doctor --platform not-a-real-platform` both produce expected output
- [ ] Manual end-to-end test against a live Kiro install (deferred — the fix is verified against Kiro's documented hook payload shape at https://kiro.dev/docs/cli/hooks)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.3] - 2026-07-09
+
+### Fixed
+
+- **Kiro hook events were silently dropped** — `preflight-collector` matched Claude Code's exact PascalCase hook event names (`PreToolUse`); Kiro sends lower-camelCase (`preToolUse`) per its own docs, so every Kiro hook event fell through to a silent no-op — nothing was written to the buffer, but the collector still exited 0. The event-name check is now case-insensitive. (#84)
+- **Hook-sourced tool names were never normalized for any platform** — `PlatformAdapter.normalizeToolCall()` and `PlatformRegistry` were fully implemented and tested but never actually called from the running hook pipeline; every platform's tool names passed through to trackers and the dashboard unmapped. Added `PlatformAdapter.mapToolName()` and wired it into `HookEventProcessor`, and corrected Kiro's tool-name map with its documented aliases (`read`/`write`/`shell`/`aws`).
+
+### Added
+
+- **`preflight doctor --platform <name>`** — the "Hooks wired" check previously only validated Claude Code's `settings.json` and stayed green regardless of whether a non-Claude-Code platform's hooks actually worked. Passing `--platform kiro` (or any other registered platform name) now skips the Claude-Code-specific check in favor of an explicit reminder to verify that platform's own hook/MCP config and to confirm events land in `~/.newrelic-preflight/buffer-*.jsonl`.
+
 ## [1.4.2] - 2026-07-09
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.2",
+      "version": "1.4.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/hooks/collector-script.test.ts
+++ b/src/hooks/collector-script.test.ts
@@ -395,6 +395,51 @@ describe('collector-script', () => {
     });
   });
 
+  function makeKiroPreToolUse(overrides?: Record<string, unknown>): string {
+    return JSON.stringify({
+      hook_event_name: 'preToolUse',
+      tool_name: 'read',
+      tool_input: { operations: [{ mode: 'Line', path: '/tmp/test.ts' }] },
+      session_id: 'kiro-sess-001',
+      cwd: '/projects/test',
+      ...overrides,
+    });
+  }
+
+  function makeKiroPostToolUse(overrides?: Record<string, unknown>): string {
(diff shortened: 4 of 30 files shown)


corinnekunze:

PR #97, opened 2026-07-10
Fix: correct Amazon Q Developer CLI adapter tool-map and hook wiring

### LINKED ISSUE #98: Amazon Q Developer CLI adapter: invented tool-map and missing hook documentation
## Summary

The Amazon Q Developer CLI platform adapter (`src/platforms/amazon-q-adapter.ts`) has an invented tool-name map: of the 15 entries in `AMAZON_Q_TOOL_MAP`, only 3 (`fs_read`, `fs_write`, `execute_bash`) correspond to real Amazon Q Developer CLI built-in tools. The other 12 (`fs_edit`, `fs_create`, `fs_delete`, `fs_list`, `fs_find`, `grep`, `search_code`, `run_shell`, `execute_command`, `explain_code`, `review_code`, `transform_code`) don't exist in Amazon Q's real tool vocabulary.

Amazon Q Developer CLI's real built-in tool set (confirmed from `https://aws.github.io/amazon-q-developer-cli/built-in-tools.html`) is exactly 9 tools: `execute_bash`, `fs_read`, `fs_write`, `introspect`, `report_issue`, `knowledge`, `thinking`, `todo_list`, `use_aws`.

Separately, `getHookInstallInstructions()` only documented MCP server registration and didn't mention that Amazon Q Developer CLI has a real, genuine hook mechanism (`preToolUse`/`postToolUse`/etc. via the agent config `hooks` field, confirmed from `https://github.com/aws/amazon-q-developer-cli/blob/main/docs/hooks.md`) that can be wired to `preflight-collector` for real tool-call observability.

While investigating, also found a cross-platform bug in `collector-script.ts`: the `postToolUse` handler hardcoded `success: true` instead of reading `tool_response.success`, which both Kiro and Amazon Q Developer CLI use to signal real tool failure (Claude Code uses a separate `PostToolUseFailure` event instead). This meant failed Kiro tool calls (and any future Amazon Q calls) were being silently recorded as successful.

## Found via

Proactive platform-adapter audit, not a user report — same pattern as the prior Kiro (#84), Cursor (#90), Windsurf (#92), and Zed (#94) adapter fixes.

## Fix

See #97.

chrisdhaan:
Closes #98

## Summary
- Replaced an invented 15-entry tool-name map (only 3 entries were real) with the confirmed 9-tool built-in vocabulary for Amazon Q Developer CLI, mapping the 4 with genuine Claude Code equivalents (`fs_read`, `fs_write`, `execute_bash`, `todo_list`) and leaving `introspect`, `report_issue`, `knowledge`, `thinking`, and `use_aws` unmapped rather than forced.
- Fixed a cross-platform bug in `collector-script.ts`: `postToolUse` events hardcoded `success: true` instead of reading `tool_response.success`, silently losing real failure signals from Kiro and Amazon Q (both encode tool failure this way, not via a separate failure event). This also corrects Claude Code's own `Edit` tool, which already sets this field on a no-match failure — unifying the top-level `success` signal with the existing `toolOutput.editSuccess` signal.
- Rewrote `getHookInstallInstructions()` to document Amazon Q Developer CLI's real, genuine hook mechanism (`preToolUse`/`postToolUse` via the agent config `hooks` field) instead of only covering MCP server registration.
- Version bump to 1.4.8.

## Test plan
- [x] `npm run build && npm test && npm run lint` — all tests passed, 0 lint errors/warnings
- [x] Verified real tool vocabulary against `https://aws.github.io/amazon-q-developer-cli/built-in-tools.html`
- [x] Verified hook payload shape and `tool_response.success` semantics against `https://github.com/aws/amazon-q-developer-cli/blob/main/docs/hooks.md` and `https://kiro.dev/docs/cli/hooks`
- [x] Confirmed backward compatibility for every existing Claude Code/Kiro/Cursor/Windsurf test fixture in `collector-script.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.8] - 2026-07-09
+
+### Fixed
+
+- Amazon Q Developer CLI adapter: replaced an invented 15-entry tool-name map (only 3 entries were real) with the confirmed 9-tool built-in vocabulary, mapping the 4 with genuine Claude Code equivalents (`fs_read`, `fs_write`, `execute_bash`, `todo_list`) and leaving the rest (`introspect`, `report_issue`, `knowledge`, `thinking`, `use_aws`) unmapped rather than forced.
+- `collector-script.ts`: `postToolUse` events no longer hardcode `success: true` — they now read `tool_response.success` when present, fixing a cross-platform bug where failed Kiro and Amazon Q tool calls were silently recorded as successful.
+- Amazon Q Developer CLI adapter: `getHookInstallInstructions()` now documents the platform's real, genuine hook mechanism (`preToolUse`/`postToolUse` via the agent config `hooks` field) instead of only covering MCP server registration.
+
 ## [1.4.7] - 2026-07-09
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.7",
+  "version": "1.4.8",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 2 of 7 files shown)


corinnekunze:

Issue #100, opened 2026-07-10
/dev/stdin EACCES when Preflight is launched from Windows-host Claude Code via wsl.exe


tswdsawada:
**Summary:** When Claude Code runs on the Windows host and spawns Preflight inside WSL (stdio crossing the Windows/Linux boundary via `wsl.exe`), any attempt to (re)open `/dev/stdin` fails with EACCES.

**Root cause:** For `wsl.exe`-launched sessions, the stdin pipe is created by WSL's init/relay, which runs as root. The pipe inode is therefore `root:root` with mode `0600`. Reading the *inherited* fd 0 works fine, but opening `/dev/stdin` (→ `/proc/self/fd/0`) re-opens the pipe inode, which fails the permission check for a non-root user.

**One-line repro (from PowerShell):**

```powershell
"x" | wsl.exe -e cat /dev/stdin
# cat: /dev/stdin: Permission denied
```

`stat -L /dev/stdin` in that context shows:

```
Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
```

The same command run entirely inside WSL succeeds, because the pipe is then created by the user's shell and owned by the user:

```
$ echo hi | cat /dev/stdin
hi
```

**Not related to:** sudo installation, WSL distro/version, or file ownership of the installed package. It also does not reproduce when Claude Code itself runs inside WSL — only when stdio crosses the Windows/WSL boundary.

**Impact:** Anyone running Claude Code on the Windows host with Preflight installed in WSL hits this — which is a natural setup for Windows users, given Preflight's WSL install support (#9, #10).

**Suggested fix:** Read the inherited fd 0 directly (e.g. `cat -`, or reading the stdin stream) instead of re-opening `/dev/stdin`.

chrisdhaan:
Duplicate of #99 (identical report, filed ~2 minutes apart). Closing this one and continuing discussion on #99.

PR #105, opened 2026-07-10
Fix: v1.4.11 — /dev/stdin EACCES when running Preflight inside WSL via wsl.exe

### LINKED ISSUE #99: /dev/stdin EACCES when Preflight is launched from Windows-host Claude Code via wsl.exe
**Summary:** When Claude Code runs on the Windows host and spawns Preflight inside WSL (stdio crossing the Windows/Linux boundary via `wsl.exe`), any attempt to (re)open `/dev/stdin` fails with EACCES.

**Root cause:** For `wsl.exe`-launched sessions, the stdin pipe is created by WSL's init/relay, which runs as root. The pipe inode is therefore `root:root` with mode `0600`. Reading the *inherited* fd 0 works fine, but opening `/dev/stdin` (→ `/proc/self/fd/0`) re-opens the pipe inode, which fails the permission check for a non-root user.

**One-line repro (from PowerShell):**

```powershell
"x" | wsl.exe -e cat /dev/stdin
# cat: /dev/stdin: Permission denied
```

`stat -L /dev/stdin` in that context shows:

```
Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
```

The same command run entirely inside WSL succeeds, because the pipe is then created by the user's shell and owned by the user:

```
$ echo hi | cat /dev/stdin
hi
```

**Not related to:** sudo installation, WSL distro/version, or file ownership of the installed package. It also does not reproduce when Claude Code itself runs inside WSL — only when stdio crosses the Windows/WSL boundary.

**Impact:** Anyone running Claude Code on the Windows host with Preflight installed in WSL hits this — which is a natural setup for Windows users, given Preflight's WSL install support (#9, #10).

**Suggested fix:** Read the inherited fd 0 directly (e.g. `cat -`, or reading the stdin stream) instead of re-opening `/dev/stdin`.

chrisdhaan:
## Summary
- Fixes #99. When Claude Code runs on a Windows host and spawns Preflight inside WSL via `wsl.exe`, the piped stdin crossing that boundary is created by WSL's root-owned init/relay (`root:root`, mode `0600`). The hook collector read stdin by opening `/dev/stdin` — a symlink to `/proc/self/fd/0` — which triggers a fresh `open()` subject to a permission check against that root-owned inode, failing with `EACCES` for the non-root user even though the already-inherited stdin file descriptor is readable. Hook events were silently dropped for every such setup.
- The collector now falls back to reading the inherited stdin file descriptor directly only when `/dev/stdin` fails specifically with `EACCES`, leaving the existing POSIX (EAGAIN-avoidant) and Windows read paths unchanged for every other case.
- Version bump to 1.4.11 + CHANGELOG entry.

## Test plan
- [x] New unit tests: fallback on `EACCES`, and non-`EACCES` errors still propagate unchanged (`src/hooks/collector-script.test.ts`)
- [x] Root cause independently verified with a local repro of the underlying Unix permission-check mechanic (an already-open fd bypasses the re-check; a fresh re-open of the same path re-triggers it)
- [x] `npm run build` — clean
- [x] `npm test` — full suite passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.11] - 2026-07-10
+
+### Fixed
+
+- **`EACCES` reading stdin when Claude Code runs on a Windows host and spawns Preflight inside WSL via `wsl.exe`** — the hook collector read stdin by opening `/dev/stdin`, a symlink to `/proc/self/fd/0`. On Linux that path is a fresh `open()` subject to a permission check, and the stdin pipe crossing the Windows/WSL boundary is created by WSL's root-owned init/relay (`root:root`, mode `0600`), so the re-open failed for a non-root user even though the already-inherited file descriptor was readable. Hook events were silently dropped for every Windows-host/WSL-guest Claude Code setup. The collector now falls back to reading the inherited stdin file descriptor directly when `/dev/stdin` specifically fails with `EACCES`, leaving the existing POSIX and Windows read paths unchanged otherwise. (#99)
+
 ## [1.4.10] - 2026-07-10
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.10",
+  "version": "1.4.11",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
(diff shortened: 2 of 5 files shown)


tswdsawada:
Thank you for the quick turnaround on this! (I reported the original issue in #100.)

I've verified the fix on the exact setup that was failing — Claude Code on a Windows host spawning Preflight inside WSL via `wsl.exe`:

- Updated to v1.4.34 — the collector now reads hook events across the Windows/WSL boundary without EACCES (piping directly into `preflight-collector`, no `cat` workaround needed), and events show up on the local dashboard in real time.

One heads-up: while updating I hit an unrelated packaging problem in the v1.4.34 release itself (registry manifest missing `undici`/`hono`/`js-yaml`, so a fresh install crashes at startup). Filed separately as #189 with a workaround.

Thanks again!

corinnekunze:

PR #106, opened 2026-07-10
Fix: nr_observe_get_platform_comparison can't differentiate platforms

### LINKED ISSUE #107: nr_observe_get_platform_comparison can't differentiate platforms
## Summary

An audit of every registered MCP tool's documented contract vs. its actual implementation found that `nr_observe_get_platform_comparison` could never actually compare platforms.

## Root cause

`buildSessionSummary()` never set the `platform` field on persisted session summaries. Every session fell into the `'claude-code'` fallback bucket used by the tool's grouping logic, regardless of which of the 9 real platform adapters (Cursor, Windsurf, Zed, Continue, Amazon Q, Kiro, Copilot, generic-mcp, Claude Code) actually generated the session. The tool's per-platform aggregation math was correct — the input data was the defect.

## Fix

The active platform is detected exactly once per process by `PlatformRegistry`, but was never exposed outside the internal event processor. Added a public `HookEventProcessor.activePlatform` getter and threaded it through to `buildSessionSummary()`'s `platform` field, so distinct platforms now produce distinct buckets.

## Related, not fixed here

`nr_observe_get_collaboration_profile` has a similar-shaped bug (three other fields — `userMessages`, `assistantMessages`, `userCorrections` — are also hardcoded to 0 in the same function), but fixing that requires new transcript-parsing logic and is a separate, larger effort.

chrisdhaan:
Closes #107

## Summary
- `nr_observe_get_platform_comparison` could never differentiate platforms — `buildSessionSummary()` never set the `platform` field on persisted session summaries, so every session fell into the `'claude-code'` fallback bucket regardless of which of the 9 real platform adapters generated it.
- The active platform is detected exactly once per process by `PlatformRegistry` (already happening, just never exposed). This threads that already-known value through: `HookEventProcessor.activePlatform` (new public getter) → `BuildSessionSummarySources.platform` (new optional field) → `FullSessionSummary.platform` (already-existing field, now populated) → `nr_observe_get_platform_comparison`'s grouping logic (already reads it, unchanged).
- No consumer-side changes needed. Found during an audit of every registered MCP tool's documented contract vs. its actual implementation.
- Does **not** fix `nr_observe_get_collaboration_profile` — that needs `userMessages`/`assistantMessages`/`userCorrections` data via new transcript-parsing logic, a separate, larger fix with its own design (deferred).
- Version bump 1.4.11 → 1.4.12 + CHANGELOG entry.

## Test plan
- [x] `npm run build` — clean
- [x] `npm run lint` — 0 errors
- [x] `npm test` — 3899 passed, 3 skipped (pre-existing, unrelated); 4 new tests
- [x] Diff scoped to exactly the 8 files this fix touches — no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.12] - 2026-07-10
+
+### Fixed
+
+- `nr_observe_get_platform_comparison` could never differentiate platforms — `buildSessionSummary()` never set the `platform` field on persisted session summaries, so every session fell into the `'claude-code'` fallback bucket regardless of which of the 9 real platform adapters generated it. The active platform (already detected once per process by `PlatformRegistry`) is now threaded through `HookEventProcessor.activePlatform` into every persisted session summary. Found during the MCP tools audit. Does not address the related `nr_observe_get_collaboration_profile` bug (missing `userMessages`/`assistantMessages`/`userCorrections` data) — that's a separate, larger fix.
+
 ## [1.4.11] - 2026-07-10
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


corinnekunze:

PR #109, opened 2026-07-10
Fix: v1.4.13 — literal path-containment check for CodeQL

### LINKED ISSUE #110: Static-asset path-containment check not recognized by CodeQL as a path-injection sanitizer
## Description

`isWithinRoot()` in `src/dashboard/routes/static-handler.ts` — the containment
check guarding the dashboard's static file handler — checked one hardcoded
`'/'` separator plus a runtime `path.sep`-derived fallback (the fallback is
needed so Windows' backslash-joined paths still pass the containment check).

CodeQL's `js/path-injection` query cannot statically prove that a runtime
value equals `'/'`, so it stopped recognizing the fallback branch as a valid
sanitizer and flagged the file reads that follow (`stat()` and `readFile()`)
as potential path injection.

## Fix

Both separator checks are now hardcoded string literals (`'/'` and `'\\'`)
instead of one literal plus a runtime-derived fallback. CodeQL recognizes
literal-anchored checks as the standard path-containment sanitizer pattern.
Behavior is unchanged on either platform: Windows already effectively used
the `'\\'` branch at runtime, and the POSIX fallback branch was already a
redundant duplicate of the `'/'` check.

## Test plan

- [x] Existing `static-handler.windows-sep.test.ts` covers both the Windows
  and POSIX containment cases without needing OS-specific test setup

chrisdhaan:
## Summary

- Rewrites `isWithinRoot()` in `static-handler.ts` to check two hardcoded
  string literals ('/' and '\\') instead of one literal plus a
  `path.sep`-derived fallback, so CodeQL recognizes it as a valid
  path-containment sanitizer instead of flagging the file reads that
  follow as path injection.
- Behavior is unchanged on both platforms: Windows already effectively
  used the '\\' branch at runtime, and POSIX's fallback branch was already
  a redundant duplicate of the '/' check.
- Bumps to v1.4.13 with a CHANGELOG entry.

## Test plan

- [x] `npm run build && npm test && npm run lint` — all pass
- [x] `static-handler.windows-sep.test.ts` covers both the Windows and
  POSIX containment cases without needing OS-specific test setup
- [x] Confirmed clean against CodeQL prior to this PR (verified on a
  throwaway branch)

Fixes #110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.13] - 2026-07-10
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


corinnekunze:

PR #111, opened 2026-07-10
DIAGNOSTIC — do not merge — inline containment check test

chrisdhaan:
Throwaway diagnostic to isolate whether CodeQL's non-recognition of the isWithinRoot() containment check is due to literal-vs-variable separators, or due to the check being routed through a separately-defined function. Will be closed without merging regardless of outcome.


chrisdhaan:
Diagnostic PR — the actions-language analysis job failed and PR-level differential checks can't tell us whether a pre-existing alert was resolved anyway. Closing without merging; continuing the investigation in an isolated scratch repo instead.

PR #112, opened 2026-07-10
Fix: v1.4.14 — path.relative() containment check for CodeQL

### LINKED ISSUE #110: Static-asset path-containment check not recognized by CodeQL as a path-injection sanitizer

chrisdhaan:
## Summary

- v1.4.13's fix (two hardcoded literal separator checks OR'd together) did
  not actually clear CodeQL's `js/path-injection` findings despite passing
  every check available at merge time.
- Root cause: CodeQL never recognizes an `||` of two `startsWith()` calls
  as a sanitizer, regardless of whether the literals are hardcoded or the
  check is inlined. It does recognize a `path.relative()`-based
  containment check, but only when inlined directly at the point of use.
- `isWithinRoot()` is now deleted; the containment check is inlined
  directly using `path.relative()`/`path.isAbsolute()`. Windows
  correctness is now handled by Node's own platform-aware path module
  rather than hand-rolled separator matching.
- Preserves the exact current 403 for an empty/double-slash request path.
- Confirmed clean against a real push-triggered CodeQL analysis on a
  disposable scratch repo before this PR was opened.

Fixes #110

## Test plan

- [x] `npm run build && npm test && npm run lint` — all pass
- [x] New `path.win32`-behavior-pinning test replaces the old
  `isWithinRoot()` unit tests
- [x] New end-to-end test for the empty/double-slash edge case
- [x] Confirmed clean against a real push-triggered CodeQL analysis before
  opening this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.14] - 2026-07-10
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


corinnekunze:

PR #113, opened 2026-07-10
Fix: nr_observe_report_feedback never emitted ai.feedback.count

### LINKED ISSUE #114: nr_observe_report_feedback recorded feedback but never emitted metrics
## Summary

An audit of every registered MCP tool's documented contract vs. its actual implementation found that `nr_observe_report_feedback` recorded feedback correctly but the metric never reached New Relic.

## Root cause

`FeedbackCollector.record()` genuinely persisted feedback in memory, but `FeedbackCollector.emitMetrics()` — which would emit `ai.feedback.count` tagged by quality — was never called from the harvest flush (`src/transport/nr-ingest.ts`). Only `costTracker` and `efficiencyScorer` were wired there.

## Fix

Wired `feedbackCollector.emitMetrics()` into the harvest flush, matching the existing pattern.

**Bonus fix, found during design rather than the original audit:** `emitMetrics()` itself had no cursor — it iterated every historical record on every call with no way to track what had already been emitted. Wiring it into the harvest flush (which repeats every ~60s for the life of a session) as-is would have re-emitted every past feedback record on every subsequent flush, silently inflating `ai.feedback.count`. Added a `lastEmittedIndex` cursor, mirroring the pattern `EfficiencyScorer.emitMetrics()` already uses for the same reason.

## Related, not fixed here

The tool's description says it "helps correlate efficiency metrics with perceived quality." This fix only closes the metrics-emission gap — it does not build an actual join between feedback records and efficiency scores or `task_id`. That remains a separate, larger effort (similar in shape to the deferred `nr_observe_get_collaboration_profile` fix).

chrisdhaan:
Closes #114

## Summary
- `nr_observe_report_feedback` recorded user quality feedback correctly, but nothing ever read the records back or emitted them — `FeedbackCollector.emitMetrics()` was never called from the harvest flush, so `ai.feedback.count` never reached New Relic.
- Fixed by wiring `emitMetrics()` into the harvest flush alongside the existing `costTracker`/`efficiencyScorer` metrics.
- **Also fixed a latent bug found during design:** `emitMetrics()` had no cursor, so calling it on every harvest flush (as this fix now does) would have re-emitted every historical feedback record on every subsequent flush, inflating the count. It now tracks a `lastEmittedIndex` cursor, mirroring `EfficiencyScorer`'s existing pattern.
- Does **not** implement an actual correlation between feedback and efficiency scores/`task_id` — that's a separate, larger fix, deferred. Tool description kept as-is per explicit decision.
- Found during the MCP tools audit.
- Version bump 1.4.14 → 1.4.15 + CHANGELOG entry.

## Test plan
- [x] `npm run build` — clean
- [x] `npm run lint` — 0 errors
- [x] `npm test` — 3901 passed, 3 skipped (pre-existing, unrelated); 3 new tests
- [x] Diff scoped to exactly the 8 files this fix touches

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.15] - 2026-07-10
+
+### Fixed
+
+- `nr_observe_report_feedback` recorded user quality feedback but never emitted it — `FeedbackCollector.emitMetrics()` was never called from the harvest flush, so `ai.feedback.count` never reached New Relic. Fixed by wiring `emitMetrics()` into the harvest flush alongside the existing `costTracker`/`efficiencyScorer` metrics. Also fixed a latent double-counting bug found while wiring this in: `emitMetrics()` had no cursor, so calling it on every harvest flush (as this fix now does) would have re-emitted every historical feedback record on every subsequent flush; it now tracks a `lastEmittedIndex` cursor, mirroring `EfficiencyScorer`'s existing pattern. Found during the MCP tools audit. Does not implement an actual correlation between feedback and efficiency scores — that remains a separate, larger fix.
+
 ## [1.4.14] - 2026-07-10
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


chrisdhaan:
Added a follow-up commit removing internal-process framing ("Found during the MCP tools audit") from this and the two already-merged CHANGELOG entries it sits alongside (v1.4.10, v1.4.12) — same category as the ADAPTERS.md wording fix in v1.4.9. No functional change.

corinnekunze:

PR #115, opened 2026-07-10
Fix: nr_observe_get_instruction_drift always reports empty data

### LINKED ISSUE #116: nr_observe_get_instruction_drift always reports empty data
## Description

`nr_observe_get_instruction_drift` is meant to track how CLAUDE.md / system prompt changes correlate with session outcomes (success rate, token usage, thrashing). In practice it always reported an empty dataset.

## Root cause

`InstructionDriftTracker.recordSessionOutcome()` and `.loadRecords()` had zero call sites in production. The tracker's read path (capturing the current prompt hash when CLAUDE.md is read) worked correctly, but nothing ever fed it session outcomes.

Wiring only `recordSessionOutcome()` in-process (at session shutdown) would not have been sufficient on its own: each MCP server process is scoped to one Claude Code session, so an in-memory recording right before the process exits is immediately lost, and the next session starts with a fresh, empty tracker. Real cross-session correlation requires persisting the prompt hash to disk and reloading historical outcomes at startup.

## Fix

- The session's prompt hash is now persisted on the saved session summary (`instructionPromptHash`).
- At startup, the last 7 days of prior sessions are reloaded and fed into the tracker.
- At shutdown, the current session's own outcome is recorded.

## Known limitation (separate follow-up)

Correlation is currently keyed on the hash of a `Read` tool call's *arguments* (file path/offset/limit), not the file's *content* — a pre-existing gap in the shared hashing utility. This means the tool detects read/path drift rather than CLAUDE.md content drift. Hashing actual file content would need its own follow-up.

chrisdhaan:
Closes #116

## Summary

- `InstructionDriftTracker.recordSessionOutcome()` and `.loadRecords()` had zero call sites in production — the tracker's read path worked, but nothing fed it session outcomes, so `nr_observe_get_instruction_drift` always reported an empty dataset.
- Fixing this requires cross-session persistence, not just in-process wiring: each MCP server process is scoped to one Claude Code session, so recording an outcome right before shutdown alone would be immediately lost and never enable the cross-session correlation that's the tool's entire purpose.
- The session's prompt hash is now persisted on the saved session summary, the last 7 days of prior sessions are reloaded into the tracker at startup, and the current session's own outcome is recorded at shutdown.
- Note: correlation is currently keyed on the hash of a `Read` tool call's *arguments* (file path/offset/limit), not the file's *content* — so this detects read/path drift rather than CLAUDE.md content drift. That's a pre-existing gap in the shared hashing utility, not introduced here; hashing actual content remains a separate, larger fix.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3909 passed, 0 failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.16] - 2026-07-10
+
+### Fixed
+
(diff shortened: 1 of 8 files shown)


corinnekunze:

PR #118, opened 2026-07-11
Fix: nr_observe_get_api_failures silently reports zero instead of disclosing the gap (v1.4.17)

### LINKED ISSUE #119: nr_observe_get_api_failures always reports zero failures — no data source in current architecture
## Bug

`nr_observe_get_api_failures` always returns empty/zero metrics — `ApiFailureTracker.recordRequest()`/`.recordFailure()` had zero call sites in production, so the tool silently reported "no failures" regardless of what actually happened.

## Root cause

Investigation found this isn't a simple missing-wiring bug like some of the other MCP tools we've fixed recently. The tracker's data model (per-model reliability scorecards, error types like `rate_limit`/`context_length_exceeded`/`authentication`) is designed to track failures of calls **to the LLM API itself** — but Preflight has no visibility into that traffic in either supported mode:

- **Stdio mode** (Claude Code hooks): only observes Claude Code's own tool-call lifecycle, not the model-API traffic underneath it.
- **Proxy mode**: Preflight's proxy forwards requests to **MCP servers**, not to the model API — there's no LLM-facing proxy in this codebase today.

## Fix

Rather than fabricate a substitute signal (e.g. relabeling MCP-server-call failures as "model reliability," which would only work in proxy mode and would mischaracterize what's being measured), the tool's response now discloses the limitation explicitly on every call: a new `dataAvailable: false` field and an explanatory `note`, plus an updated tool description. This prevents callers from mistaking the permanent zero output for "no failures occurred."

Building a real LLM-facing proxy to make this data genuinely observable remains a separate, much larger effort — not in scope here.

Shipped as v1.4.17.

chrisdhaan:
Closes #119

## Summary

- `nr_observe_get_api_failures` always silently reported zero API failures because `ApiFailureTracker.recordRequest()`/`.recordFailure()` had zero call sites in production.
- Model-API-level failure data (rate limits, timeouts, auth errors from the LLM provider itself) is not observable in Preflight's current architecture in either mode: Claude Code hook events only see Claude Code's own tool calls, and proxy mode forwards requests to MCP servers, not to the model API itself — there's no LLM-facing proxy in this codebase today.
- Rather than fabricate a substitute signal, the tool's actual response now carries an explicit `dataAvailable: false` field and an explanatory `note` on every call, plus an updated tool description, `index.ts` comment, and `COMMANDS_TABLE.md` entry — so a caller can no longer mistake the permanent all-zero output for "no failures occurred."
- Ships as v1.4.17 with a CHANGELOG entry.

## Test plan

- [x] `npm run build && npm test && npm run lint && npm run format:check` — all pass, 0 failures, 0 lint errors/warnings
- [x] New test asserts `dataAvailable`/`note` are always present and unaffected by recording a real failure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.17] - 2026-07-10
+
+### Fixed
+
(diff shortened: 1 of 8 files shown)


corinnekunze:

PR #120, opened 2026-07-11
Fix: nr_observe_get_latency_decomposition falsely claimed a proxy-mode fix path (v1.4.18)

### LINKED ISSUE #121: nr_observe_get_latency_decomposition: code comment and docs falsely claim a proxy-mode fix path
## Summary

`nr_observe_get_latency_decomposition`'s code comment (in `src/index.ts`) and public documentation (`docs/COMMANDS_TABLE.md`) both claim the tool "only" needs proxy-mode instrumentation to become functional. This claim is false.

## Details

The tool's tracker (`LatencyDecompositionTracker`) is correctly implemented but never constructed in production — `src/index.ts` hardcodes it to `undefined`, so the tool is never listed in `tools/list`. A caller who calls it directly by name gets an explicit `isError: true` response.

Tracing the actual proxy-mode code shows the "proxy mode" fix path doesn't exist: Preflight's proxy mode forwards requests to MCP servers, not to the model API. The latency it observes (`upstreamLatencyMs`) is latency to the upstream **MCP server**, not the model API — Claude Code talks to the model API directly in both stdio and proxy modes, bypassing Preflight either way. There's also no turn-boundary hook (`UserPromptSubmit`/`Stop`) wired anywhere in the collector, so even stdio mode can't cleanly separate LLM thinking time from tool-execution/overhead time.

This is the same architectural gap found for `nr_observe_get_api_failures`. Unlike that tool, this one's current runtime behavior is already safe (loud explicit error, invisible to discovery) rather than misleading — so the fix here is correcting the false comment/error-message/docs, not changing registration or fabricating data.

## Expected behavior

- The code comment above the tracker's instantiation should accurately describe why the tool doesn't work in either mode.
- The direct-call error response should explain why, not just say "not available."
- `docs/COMMANDS_TABLE.md` should not show a fabricated example response implying the tool works.

chrisdhaan:
Closes #121

## Summary

- `nr_observe_get_latency_decomposition`'s code comment and public docs both falsely claimed the tool "only" needs proxy-mode instrumentation to work.
- Verified false by tracing the actual code: Preflight's proxy mode forwards requests to MCP servers, not to the model API, so its visible latency (`upstreamLatencyMs`) is MCP-server latency, not model-API latency — the same architectural gap found in the `nr_observe_get_api_failures` fix. There's also no turn-boundary hook (`UserPromptSubmit`/`Stop`) wired anywhere, so even stdio mode can't cleanly separate LLM thinking time from overhead.
- Unlike `nr_observe_get_api_failures`, this tool's runtime behavior was already safe — it's never listed in `tools/list` (tracker hardcoded `undefined`), and a direct call by name already returns an explicit `isError: true` response rather than misleading data. So this fix corrects the false claims rather than changing any registration, data shape, or tracker interface.
- Adds an explanatory `note` field to the existing error response, corrects the `src/index.ts` comment, and replaces the fabricated example response in `docs/COMMANDS_TABLE.md` with an accurate "not currently functional" status.
- Incidental: added `docs/superpowers/` to `.gitignore` — closes a gap where a sibling planning-docs directory wasn't covered by the existing ignore rule.

## Test plan

- [x] New test: `calling nr_observe_get_latency_decomposition returns an explanatory error` (session-stats.test.ts)
- [x] Full suite: 3914 tests, 3911 passed / 3 skipped, 0 failures
- [x] `npm run build && npm test` clean

🤖 Generated with assistance from Claude Sonnet 5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ coverage/
 .env*
 !.env.example
 .superpowers/
+docs/superpowers/
 .claude/
 test-results/
 playwright-report/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.18] - 2026-07-10
+
+### Fixed
+
+- `nr_observe_get_latency_decomposition`'s code comment and documentation both falsely claimed the tool "only" needs proxy-mode instrumentation to work. Verified false: Preflight's proxy mode forwards requests to MCP servers, not to the model API, so its visible latency is MCP-server latency, not model-API latency — the same architectural gap found in the `nr_observe_get_api_failures` fix above. Unlike that tool, this one's runtime behavior was already safe (it's never listed in `tools/list`, and a direct call already returns an explicit error rather than misleading data), so this fix corrects the false claims in the code comment, the direct-call error message, and the docs table — no runtime behavior, registration, or data shape change beyond adding an explanatory `note` field to the existing error response.
+
 ## [1.4.17] - 2026-07-10
 
 ### Fixed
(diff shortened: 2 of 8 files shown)


corinnekunze:

PR #125, opened 2026-07-11
Fix: nr_observe_get_context_composition and nr_observe_get_decision_tree overclaims

### LINKED ISSUE #126: nr_observe_get_context_composition claims 4 token categories but 2 are always 0
## Description

`nr_observe_get_context_composition`'s description promises a "per-turn token breakdown by category (system prompt, conversation history, tool results, injected files)." In practice, 2 of the 4 named categories are always 0, and this wasn't disclosed anywhere.

## Root cause

The sole production entry point into `ContextCompositionTracker` is `recordTokenEvent()` (called from `src/index.ts`), which hardcodes `systemPromptTokens: 0` and `injectedFileContentTokens: 0` on every call. All real input tokens get bucketed into only `conversation_history` (cache-read + cache-creation) and `tool_results` (the remainder).

`system_prompt` tokens are architecturally unknowable: the only data source is the model API's aggregate `usage` object (`input_tokens`/`cache_read_input_tokens`/`cache_creation_input_tokens`), which has no sub-breakdown by content category, and Claude Code's hooks never expose the system prompt text itself. `injected_file_content` is *theoretically* estimable from Read-tool output byte sizes, but that would mix an estimated heuristic into otherwise-exact, API-reported categories — a distinct feature decision, not pursued here.

`fillPercent` and the dominance alerts are genuinely correct and unaffected by this gap.

## Fix

The tool's response now carries an explanatory `note` field on every call, and the registered description now discloses the limitation directly (`LIMITATION: system_prompt and injected_file_content are always 0 ...`).

## Scope note

Found as one of 2 "unsure" tools in the original MCP tools audit (alongside `nr_observe_get_decision_tree`, see the sibling issue for that tool's fix). Unlike the audit's 6 "broken" tools, this wasn't a wiring bug — the tool is registered and mostly correct, just silently missing disclosure for 2 fields.

### LINKED ISSUE #127: nr_observe_get_decision_tree's reasoning field looks like extracted reasoning but is undisclosed rule-based templates
## Description

`nr_observe_get_decision_tree`'s description promises "reasoning and action extraction per turn." In practice, the `reasoning` field on every branch is always one of 3 hardcoded template strings, and neither the description nor the response disclosed this.

## Root cause

`DecisionTracker.recordToolCall(record: ToolCallRecord)` takes only tool-call metadata — no parameter carries model reasoning text. The 3 fixed templates are: `` `recovery after ${lastToolName} failure` ``, `'delegating to user'`, and `` `retrying ${toolName} on ${filePath} (${count} attempts)` ``. It's also not literally "per turn" — a branch is only recorded on 3 narrow triggers (previous call failed, `AskUserQuestion` called, or a 3rd+ same-tool-same-file call), so most ordinary successful turns produce zero branches.

Outcome-tagging, `getPostMortem()`, and the failure-streak/success-rate computation are all genuinely correct and unaffected by this gap.

## Fix

The tool's response (including the `post_mortem: true` path, which bypasses the tracker's normal metrics response) now carries an explanatory `note` field on every call, and the registered description no longer claims "reasoning...extraction" or literal "per turn" framing.

## Scope note

Found as one of 2 "unsure" tools in the original MCP tools audit (alongside `nr_observe_get_context_composition`, see the sibling issue #126 for that tool's fix). This issue covers only the disclosure fix. Real reasoning extraction (parsing the model's actual `thinking`/`text` transcript blocks) is technically feasible but deferred as its own design effort — tracked separately in #124.

chrisdhaan:
Closes #126
Closes #127

## Summary

Fixes the final 2 tools left over from the MCP tools audit fix series (see #104, #107, #114, #116, #119, #121). Unlike those prior fixes, these two tools aren't wiring bugs — both are registered, working, and return partially-real data, but each silently substitutes a fake/templated value for a specific field without disclosing it.

- **`nr_observe_get_context_composition`**: `system_prompt` and `injected_file_content` token categories are always 0 — architecturally unknowable, since the model API's usage response only reports aggregate input/cache-read/cache-creation counts with no per-category breakdown. `fillPercent`/dominance alerts are genuinely correct and unaffected.
- **`nr_observe_get_decision_tree`**: the `reasoning` field is always one of 3 hardcoded rule-based templates, not extracted model chain-of-thought — `DecisionTracker.recordToolCall()` has no parameter carrying reasoning text. Branches are also only recorded on 3 narrow triggers, not every turn.

Both fixes apply the same `dataAvailable`/`note` payload-disclosure pattern already established in this codebase by `nr_observe_get_api_failures` (#119): an unconditional `note` field on every response (including `decision_tree`'s `post_mortem: true` path, which bypasses `getMetrics()` entirely), plus a `LIMITATION:` clause on each tool's registered description. No tracker computation logic changed — purely disclosure.

A real fix for `decision_tree`'s reasoning extraction is technically feasible (verified against a real transcript JSONL — `thinking`/`text` blocks precede each `tool_use` entry) but is real design work, tracked separately. `context_composition`'s `system_prompt` gap is a hard architectural block with nothing further to design.

## Test plan

- [x] New unit tests for both trackers' unconditional `note` field (stable across real recordings)
- [x] Extended existing handler tests in `extended-analytics-tools.test.ts` for both tools, including `decision_tree`'s post-mortem path
- [x] `npm run build && npm test` — 3913 passed, 3 skipped (unrelated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.19] - 2026-07-10
+
+### Fixed
+
+- `nr_observe_get_context_composition`'s per-turn token breakdown claims 4 categories (system prompt, conversation history, tool results, injected files), but two of them — `system_prompt` and `injected_file_content` — are always 0: the model API's usage response only reports aggregate input/cache-read/cache-creation token counts, with no breakdown by content category, so Preflight has no way to separate those two categories from the rest. `fillPercent` and the dominance alerts are unaffected and reflect real totals. The tool's response now carries an explanatory `note` field on every call, and the registered description now discloses the limitation.
+- `nr_observe_get_decision_tree`'s `reasoning` field reads like extracted reasoning but is always one of 3 fixed rule-based labels (e.g. "recovery after X failure") — `DecisionTracker.recordToolCall()` has no parameter carrying actual model reasoning text. Branches are also only recorded on 3 narrow triggers, not on every turn, so `totalBranches` undercounts ordinary turns relative to a literal "per turn" reading. The tool's response (including the `post_mortem: true` path) now carries an explanatory `note` field, and the registered description no longer claims "reasoning...extraction."
+
 ## [1.4.18] - 2026-07-10
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.18",
+  "version": "1.4.19",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.18",
+      "version": "1.4.19",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.18",
+  "version": "1.4.19",
   "type": "module",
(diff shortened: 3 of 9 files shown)


corinnekunze:

PR #169, opened 2026-07-12
Fix: storage crash-recovery data-loss race, non-atomic writes, and week-53 validation (v1.4.20)

### LINKED ISSUE #149: LocalStore's crash-recovery drain merge can silently drop a concurrently-written hook event
## Description

If the MCP server process crashes mid-drain (after renaming the buffer file to `.drain` but before reading/deleting it) and later recovers, the recovery-merge path can silently lose a hook event that was written by `preflight-collector` in the narrow window between the merge's read and write.

## Root cause

`drainPath()`'s recovery branch (`src/storage/local-store.ts:823-833`) — reached when both a stale `.drain` file and a new live buffer file exist — does: `readFileSync(tmpPath)` → `readFileSync(bufferPath)` → `writeFileSync(bufferPath, drainData + bufferData, {mode: 0o600})`. This last write uses the default `'w'` flag (truncate + overwrite), with no temp-file-then-atomic-rename step — unlike the main drain path 24 lines earlier (`renameSync(bufferPath, tmpPath)`, line 847), which correctly uses POSIX atomic rename specifically to avoid this class of bug.

## Concrete failure sequence (verified, not theoretical)

1. MCP poll calls `drainPath()`; `renameSync(bufferPath → .drain)` succeeds, but the process is killed before the file is read/deleted. `.drain` now holds events A, B; no live buffer exists.
2. `preflight-collector` (`src/hooks/collector-script.ts`) fires on the next tool-call hook and appends event C via its own independent `O_APPEND` write, creating a new `bufferPath` = `[C]`.
3. The next MCP poll enters the recovery branch: reads `.drain` = `[A,B]`, reads buffer = `[C]`.
4. A second hook fires and appends event D to `bufferPath` — landing after the read in step 3 but before the merge write completes. Live buffer on disk is now `[C,D]`.
5. The merge's `writeFileSync(bufferPath, "A\nB\nC\n")` truncates and overwrites the whole file with only what was read in step 3. **Event D is silently gone** — no error, no warning beyond the generic recovery log line.

This requires a crashed-drain-then-new-buffer-file precondition plus a same-millisecond concurrent append, so the window is narrow — but it's reachable, not hand-waving, and it silently drops a real tool-call event (part of a session's telemetry).

## Fix

Merge into a fresh temp file, then `renameSync` it over `bufferPath`, matching the atomicity pattern the main drain path already uses correctly two dozen lines away.

## Scope note

Found during a `src/storage/` subsystem audit (2026-07-11) — the highest-priority finding in that audit. Shares a root-cause *pattern* (missing atomic-rename, not a shared vulnerability) with a similar gap found in `weekly-summary.ts` in the same audit — see the sibling issue for that finding.

### LINKED ISSUE #153: nr_observe_get_weekly_summary silently returns a wrong date range for a nonexistent ISO week 53
## Description

Requesting a week ID like `"2025-W53"` — a week that doesn't exist for that year — returns a plausible-looking but semantically wrong `WeeklySummary` instead of an error.

## Root cause

`getWeekDateRange()` (`src/storage/weekly-summary.ts:72-96`) computes `targetMonday = week1Monday + (week-1)*7 days` with no upper-bound check against how many ISO weeks the requested year actually has. Confirmed by direct computation: `getWeekDateRange('2025-W53')` returns `{start: 2025-12-29, end: 2026-01-05}` — a real, populated range — but feeding its midpoint back into `getIsoWeekId()` yields `2026-W01`, and 2025 genuinely has no 53rd ISO week (`getIsoWeekId(Dec 28 2025)` → `2025-W52`). The same is true for 2024 and 2027; only some years (e.g. 2026) have a real week 53.

**Externally reachable**: `nr_observe_get_weekly_summary`'s handler (`src/tools/cross-session-tools.ts:325-369`) validates `args.week` with only `/^\d{4}-W\d{2}$/`, which `"2025-W53"` passes trivially — no check against the requesting year's real week count. `generate()` re-checks the identical permissive regex and proceeds. A caller gets back a populated-looking summary for an 8-day-shifted window that actually overlaps the following year's week 1, with no error or warning.

## Fix

Validate `week` by round-tripping it through `getIsoWeekId()` (compute the candidate Monday, reparse it, confirm the week ID matches what was requested) rather than adding year-specific week-count logic — this also protects internal callers (e.g. `handleSendDigest`-style code) that call `generate()` directly without going through the MCP tool's input regex.

## Scope note

### LINKED ISSUE #154: A crash during shutdown can permanently corrupt a weekly summary file, and getLatest() has no fallback to recover from it
## Description

Two related gaps in `weekly-summary.ts`'s write/read path, bundled here since a fix for the first should account for the second.

## Findings

1. **`checkAndGenerateLastWeek()`'s existence check can't distinguish a valid summary file from a truncated one.** `generate()`'s write (`weekly-summary.ts:136`) is a single non-atomic `writeFileSync` with no temp-file-then-rename step. `checkAndGenerateLastWeek()` (`weekly-summary.ts:172`) only calls `existsSync(filepath)`, never validates content, before skipping backfill. If the process is killed (SIGKILL, OOM, host crash) mid-write, the resulting file is truncated/invalid JSON but still passes `existsSync` — so that week's backfill slot is permanently "filled" with corrupt data and never regenerated. This is reachable in practice: `checkAndGenerateLastWeek()` is invoked specifically from the shutdown handler (`src/index.ts:1527`) — exactly the highest-risk moment for an interrupted write.
2. **`getLatest()` amplifies this**: it picks only the single lexicographically-last filename (`weekly-summary.ts:146-165`) and returns `null` on a parse failure, with no fallback to the next-most-recent valid summary. A single corrupted "latest" file (from finding 1, or from any other cause) doesn't just leave a gap for its own week — it blinds every `getLatest()` caller (including `nr_observe_get_weekly_summary`'s `'latest'`/no-arg path) to all older, still-valid summaries too, until a newer valid file eventually supersedes it.

## Fix

1. Use a temp-file-then-atomic-rename write for `generate()`, matching the pattern already used correctly elsewhere in this subsystem (e.g. `LocalStore`'s main drain path).
2. Add a fallback in `getLatest()` to try the next-most-recent file on a parse failure rather than returning `null` outright.

## Scope note

Found during a `src/storage/` subsystem audit (2026-07-11). Shares a root-cause *pattern* (missing atomic-rename) with a similar gap found in `local-store.ts` in the same audit — see the sibling issue for that finding.

chrisdhaan:
## Summary

Three storage-subsystem correctness bugs found during an internal audit:

- `LocalStore`'s crash-recovery drain merge could silently drop a hook event `preflight-collector` wrote in the narrow window between the merge's read of the live buffer and its overwrite of that buffer with the merged content. Recovery now extracts the leftover `.drain` file's events in memory and lets the buffer's own already-atomic claim-and-drain path (rename, not read-then-overwrite) handle the live buffer, so a concurrent append can never be clobbered.
- `WeeklySummaryGenerator.generate()`'s write was a single non-atomic `writeFileSync`; a crash mid-write left a truncated/invalid JSON file that still passed the shutdown handler's `existsSync()` check, permanently corrupting that week's backfill slot. The write now goes through a temp file plus atomic rename.
- `getLatest()` returned `null` outright if the single lexicographically-latest weekly summary file was corrupt, blinding every caller (including `nr_observe_get_weekly_summary`'s no-arg/`"latest"` path) to all older, still-valid summaries. It now falls back to the next-most-recent valid file.
- Requesting a nonexistent ISO week (e.g. `"2025-W53"` — 2025 only has 52 weeks) silently returned a real-looking but wrong date range that actually overlapped the following year's week 1. `getWeekDateRange()` now rejects such inputs, and `nr_observe_get_weekly_summary` returns a clean error instead of a misleading result.

This is a routine incremental sync per `docs/PUBLIC_RELEASE_PLAN.md` — the version bump and release tag will follow as a separate step once this merges.

Fixes #149
Fixes #154
Fixes #153

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3922 passed, 3 skipped (pre-existing), 0 failed
- [x] `npm run lint && npm run format:check` — 0 errors, 0 warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.20] - 2026-07-11
+
+### Fixed
+
+- `LocalStore`'s crash-recovery drain merge could silently drop a hook event that `preflight-collector` wrote in the narrow window between the merge's read of the live buffer and its overwrite of that buffer with the merged content. Recovery now extracts the leftover `.drain` file's events in memory and lets the buffer's own already-atomic claim-and-drain path (rename, not read-then-overwrite) handle the live buffer, so a concurrent append can never be clobbered.
+- `WeeklySummaryGenerator.generate()`'s write was a single non-atomic `writeFileSync`; a crash mid-write (SIGKILL, OOM, host crash) left a truncated/invalid JSON file that still passed the shutdown handler's `existsSync()` check, permanently corrupting that week's backfill slot. The write now goes through a temp file plus atomic rename.
+- `getLatest()` returned `null` outright if the single lexicographically-latest weekly summary file was corrupt, blinding every caller (including `nr_observe_get_weekly_summary`'s no-arg/`"latest"` path) to all older, still-valid summaries. It now falls back to the next-most-recent valid file.
+- Requesting a nonexistent ISO week (e.g. `"2025-W53"` — 2025 only has 52 weeks) silently returned a real-looking but wrong date range that actually overlapped the following year's week 1. `getWeekDateRange()` now rejects such inputs, and `nr_observe_get_weekly_summary` returns a clean error instead of a misleading result.
+
 ## [1.4.19] - 2026-07-10
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.19",
+  "version": "1.4.20",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.19",
+      "version": "1.4.20",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.19",
+  "version": "1.4.20",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/storage/local-store.ts
+++ b/src/storage/local-store.ts
@@ -25,6 +25,22 @@ const SESSION_ID_RE = /^[a-zA-Z0-9_-]{1,128}$/;
  */
 const ORPHAN_BUFFER_MTIME_MS = 5 * 60 * 1000;
 
+function parseHookEvents(raw: string): HookEvent[] {
+  if (!raw.trim()) {
+    return [];
+  }
+  const events: HookEvent[] = [];
+  for (const line of raw.split('\n')) {
+    if (!line.trim()) continue;
+    try {
+      events.push(JSON.parse(line) as HookEvent);
+    } catch {
+      logger.warn('Skipping malformed buffer line', { line: line.slice(0, 100) });
+    }
+  }
+  return events;
+}
+
 /**
  * Liveness probe via signal 0 — POSIX sends no signal but performs the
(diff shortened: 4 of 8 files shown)


corinnekunze:

PR #170, opened 2026-07-12
Fix proxy telemetry integrity in standalone mode (v1.4.21)

### LINKED ISSUE #142: Standalone proxy mode sends zero telemetry to New Relic — no events, no metrics, no audit-trail recording
## Description

Standalone HTTP-proxy mode is a first-class, documented operating mode (`docs/ARCHITECTURE.md`, `docs/METRICS_TABLE.md`). Running it today sends nothing to New Relic and records nothing to the audit trail — the entire mode currently provides zero observability value for its stated purpose.

## Root cause

`ProxyManager`'s `onToolCall`/`onRequest` callbacks are wired to `logger.debug()` only (`src/index.ts:1802-1815`). `NrIngestManager` — the class that turns tool-call records into `AiMcpToolCall`/`AiProxyRequest` NR events and the 60-second `ai.mcp.*` gauge emission — is only ever constructed inside the `--stdio`/`--local` branch (`src/index.ts:1185,1613`, both gated by `if (options.stdio || options.local)` at line 628). The standalone proxy-mode branch is the `else` of that same `if` (line 1778) and never constructs it.

Confirmed exhaustively (every `ProxyManager`/`NrIngestManager` construction site in the codebase checked, three independent times during audit and review): there is no second initialization path, no shared registry, no DI container — `nrIngest` is simply never declared in scope for the proxy-mode branch.

Three concrete consequences, all from this one root cause:
1. `AiMcpToolCall` (`docs/METRICS_TABLE.md:80`, "Emitted for proxied tool calls") never fires.
2. The `ai.mcp.*` proxy gauges (`docs/METRICS_TABLE.md:337`, "only when proxy mode is active") never fire — the doc's parenthetical is backwards.
3. `AuditTrailManager.recordProxyCall()` (fully implemented, classifies destructive/sensitive/network patterns) is never called — a destructive command or sensitive-file argument routed through the standalone proxy is never classified, logged, or surfaced as a security alert.

A related, now-moot documentation claim: CLAUDE.md correctly states `AiProxyRequest` events include `team_id`/`project_id`/`org_id` — the code that would set these fields is correct, but since the event never fires in proxy mode, the claim is true-but-unreachable. No separate action needed; this resolves automatically once the fix below lands.

## Fix

Wire `NrIngestManager` (or an equivalent lightweight telemetry path) into the standalone proxy-mode branch in `src/index.ts`, so `onToolCall`/`onRequest` actually reach `ingestToolCall`/`ingestProxyRequest`. Until this lands, `docs/METRICS_TABLE.md`'s proxy-mode claims should at minimum be corrected to match `docs/ARCHITECTURE.md`'s own (accurate) diagram, which already shows `onToolCall`/`onRequest` going to a debug log, not to NR.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11), the highest-priority finding in that audit — confirmed via exhaustive verification, not a narrow edge case.

### LINKED ISSUE #145: OTLP receiver: startup failure is a silent no-op — proxy mode reports healthy with the receiver absent
## Description

If the local OTLP/HTTP receiver fails to start (port already in use, or a bad `forwardEndpoint` failing its constructor's SSRF check), proxy mode continues successfully with the receiver silently absent — no crash, no visible signal beyond a debug-level stderr log line.

## Root cause

`ProxyManager.start()` (`src/proxy/proxy-manager.ts:178-192`) wraps `new OtlpReceiver(...)` + `await receiver.start()` in a try/catch whose only action on failure is `logger.warn('OTLP receiver failed to start — disabled', { error: String(err) })`. Proxy startup returns normally either way. An operator who enables `otlpReceiverEnabled` and hits `EADDRINUSE` on port 4318 (plausible — it's the OTel Collector's own default port) gets no receiver and no indication beyond reading stderr JSON logs; the proxy reports itself healthy and running.

## Fix

Surface a clear, non-debug-level signal when the receiver is enabled but fails to start — e.g. a health-check field, or a hard-fail-at-startup option for operators who want to know immediately rather than discover it later.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11).

chrisdhaan:
## Summary

- Fixes #142: standalone HTTP-proxy mode sent zero telemetry to New Relic. `ProxyManager`'s `onToolCall`/`onRequest` callbacks were wired to `logger.debug()` only, and `NrIngestManager` was never constructed on that code path. Proxy mode now constructs and starts an `NrIngestManager` (gated on `mode !== 'local'`, mirroring the existing `--stdio`/`--local` behavior) and feeds it from the proxy callbacks via a new, independently-tested `buildProxyTelemetryCallbacks()` helper in `src/index.ts`.
- Fixes #145: if the local OTLP/HTTP receiver failed to start (e.g. port already in use), proxy mode continued reporting itself healthy with the receiver silently absent, with no signal beyond a log line. `ProxyManager` now tracks OTLP receiver status (`disabled`/`running`/`failed`) and surfaces it in the `GET /health` response whenever the receiver is enabled; the failure is now logged at `error` level instead of `warn`.
- Bumps version to 1.4.21 (`package.json` + `package-lock.json`) and adds a CHANGELOG entry.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3928 passed, 3 pre-existing skips, 0 failures
- [x] `npm run lint` — clean
- [x] New unit tests for `buildProxyTelemetryCallbacks()` (src/index.test.ts): forwards tool-call/request records to `NrIngestManager` when present, no-op-safe when absent (local mode)
- [x] New unit tests for OTLP receiver status (src/proxy/proxy-manager.test.ts): `disabled` when not enabled, `running` on successful start, `failed` on a genuine `EADDRINUSE` (real bound socket, not mocked)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.21] - 2026-07-11
+
+### Fixed
+
+- Standalone HTTP-proxy mode sent zero telemetry to New Relic — `ProxyManager`'s `onToolCall`/`onRequest` callbacks were wired to `logger.debug()` only, and `NrIngestManager` was never constructed on that code path, so `AiMcpToolCall`/`AiProxyRequest` events, the `ai.mcp.*` proxy gauges, and audit-trail security recording for proxied tool calls never fired. Proxy mode now constructs and starts an `NrIngestManager` (gated on `mode !== 'local'`, matching the existing `--stdio`/`--local` behavior) and feeds it from the proxy callbacks.
+- If the local OTLP/HTTP receiver failed to start (e.g. port already in use), proxy mode continued reporting itself healthy with the receiver silently absent, with no signal beyond a log line. `ProxyManager` now tracks OTLP receiver status (`disabled` / `running` / `failed`) and surfaces it in the `GET /health` response whenever the receiver is enabled, and the failure is now logged at `error` level.
+
 ## [1.4.20] - 2026-07-11
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.20",
+  "version": "1.4.21",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.20",
+      "version": "1.4.21",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.20",
+  "version": "1.4.21",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -5,6 +5,7 @@ import { resolve, join } from 'node:path';
 import {
   parseArgs,
   maskCredential,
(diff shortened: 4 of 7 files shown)


corinnekunze:

PR #171, opened 2026-07-12
Fix install verification gaps (v1.4.22)

### LINKED ISSUE #155: setup-wizard.ts daemon install claims availability without health verification
## Summary

`setup-wizard.ts`'s background dashboard daemon install (macOS) prints:

> "✓ Background dashboard daemon installed — dashboard will be available at `http://127.0.0.1:<port>` after login."

immediately after `installDashboardDaemon()` returns without throwing (`setup-wizard.ts:599-605`). But `installDashboardDaemon()` (`schedule.ts:271-286`) only verifies that `launchctl load` itself did not throw — it never polls the dashboard's HTTP health endpoint or otherwise confirms the daemon process actually started and is serving traffic. A `launchctl load` can succeed while the spawned process immediately crashes (e.g. missing binary permissions, a corrupt plist from a bad `nodeDir`/`binaryPath`), and the user sees an unconditional success message with no follow-up warning.

## Why this matters

This is the exact "claims success without verifying" bug shape documented in the historical issue #83 (a restart-offer could report success when the restart had actually failed). The **same codebase already has the correct fix** for this exact call: `offerRestarts()` → `verifyRestart()` (`cli.ts:331-345`, `355-378`) polls `GET /api/health` for up to 5s after calling the identical `installDashboardDaemon()`, and prints a distinct "could not be verified as healthy" message if the poll fails. The setup wizard's daemon-install step just doesn't reuse that pattern.

## Suggested fix

Reuse `verifyRestart()`/`waitForHealthyDashboard()` (or extract a shared helper) in `setup-wizard.ts`'s daemon-install success path, and downgrade the message (or add a follow-up warning) when the health poll fails — matching what `cli.ts`'s `update` command already does correctly for the identical underlying call.

## Related, lower-priority coverage gap

The `installDashboardDaemon()`/`installSchedule()` primitives themselves have zero test coverage of their own `launchctl load`-failure path (`schedule.ts:178-181`, `281-284`) — the exact error signal every caller (including the fix above) would rely on to detect install failure. Worth covering with a regression test alongside this fix.

No test in `setup-wizard.test.ts` covers the daemon-install step at all.

## Scope note

### LINKED ISSUE #156: checkStorageWritable diagnostic doesn't distinguish a writable file from a writable directory
## Summary

`diagnostics.ts`'s `checkStorageWritable` check (`diagnostics.ts:311-314`) reports "Storage writable: ok" using only `accessSync(storagePath, constants.W_OK)`. This call succeeds equally for a writable regular *file* as for a writable *directory* — `fs.accessSync` with `W_OK` does not throw for a plain file.

If `storagePath` were ever a file instead of a directory (e.g. a corrupted install, a manual `touch`, or a race during a prior partial/corrupted install), this diagnostic would report `ok` while `LocalStore`'s downstream `mkdirSync(storagePath, { recursive: true, mode: 0o700 })` calls (`local-store.ts:113,260,296,371`) would fail with `ENOTDIR`, silently blocking all telemetry persistence with no diagnostic signal pointing at the real cause.

## Why this matters

This is the same "success signal without verifying the underlying operation would actually succeed" bug shape as historical issue #83 — the third recurrence of this pattern found during this audit series (also found in `setup-wizard.ts`'s daemon-install step, see linked issue). Trigger condition is narrow/unlikely, but the check never rules it out.

## Suggested fix

Add a `statSync(storagePath).isDirectory()` check alongside the existing `accessSync(W_OK)` call, and report a distinct failure message (e.g. "Storage path exists but is not a directory") when it's a file.

Ship with a regression test — none currently exists for this branch.

## Scope note

chrisdhaan:
## Summary

- `setup-wizard.ts`'s background-dashboard-daemon install step reported success right after `launchctl load` returned without throwing, which only confirms the plist loaded, not that the daemon process actually came up healthy — a plist can load successfully while the spawned process immediately crashes. It now polls the dashboard's `/api/health` endpoint via the same `getDashboardAddress()`/`waitForHealthyDashboard()` helpers `cli.ts`'s `update` command already uses, and downgrades the success message to a warning (with `launchctl list` / log-file pointers) when the health check fails. Also adds regression tests for the `launchctl load`-failure path in both `installSchedule` and `installDashboardDaemon`.
- `checkStorageWritable()` used `accessSync(W_OK)` alone, which succeeds identically for a writable file and a writable directory — a corrupted install with a plain file sitting at the storage path would falsely report "ok" while `LocalStore`'s downstream `mkdirSync()` calls would fail with `ENOTDIR`. It now also calls `statSync().isDirectory()` and reports a distinct "exists but is not a directory" failure when the path exists but isn't a directory.
- Bumps version to 1.4.22 and adds a CHANGELOG entry.

Fixes #155. Fixes #156.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3936 passed, 3 pre-existing skips, 0 failures
- [x] `npm run lint` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.22] - 2026-07-11
+
+### Fixed
+
+- `setup-wizard.ts`'s background-dashboard-daemon install step reported success right after `launchctl load` returned without throwing, which only confirms the plist loaded, not that the daemon process actually came up healthy — a plist can load successfully while the spawned process immediately crashes. It now polls the dashboard's `/api/health` endpoint via the same `getDashboardAddress()`/`waitForHealthyDashboard()` helpers `cli.ts`'s `update` command already uses, and downgrades the success message to a warning (with `launchctl list` / log-file pointers) when the health check fails.
+- `checkStorageWritable()` used `accessSync(W_OK)` alone, which succeeds identically for a writable file and a writable directory — a corrupted install with a plain file sitting at the storage path would falsely report "ok" while `LocalStore`'s downstream `mkdirSync()` calls would fail with `ENOTDIR`. It now also calls `statSync().isDirectory()` and reports a distinct "exists but is not a directory" failure (with a `rm && mkdir` fix suggestion) when the path exists but isn't a directory.
+
 ## [1.4.21] - 2026-07-11
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.21",
+  "version": "1.4.22",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.21",
+      "version": "1.4.22",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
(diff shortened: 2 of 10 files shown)


corinnekunze:

PR #172, opened 2026-07-12
Fix replay-timeline bug + dashboard routing hygiene (v1.4.23)

### LINKED ISSUE #162: analyzeReplayTimeline()'s main persisted-session path has no defensive sort, unlike its own fallback branches
## Summary

`src/dashboard/routes/replay-analyzer.ts`'s `analyzeReplayTimeline()` (lines 23-42) performs no `.sort()` of its own and trusts its input array's ordering — every anti-pattern/segment detector inside it assumes chronological order.

Its primary production caller, `buildReplayResponse()`'s **persisted-session branch** (`src/dashboard/routes/api-handler.ts:535-553` — the code path used for every completed/historical session, i.e. the main replay use case, not a fallback), passes `session['timeline']` straight through with **no sort call**. Meanwhile the function's two *fallback* branches (the `TaskDetector`-derived branch and the `toolCallBuffer`-derived branch) both explicitly `.sort((a, b) => a.timestamp - b.timestamp)` before calling `analyzeReplayTimeline()`.

## Why this matters

Traced the write side: nothing upstream guarantees chronological order either. `session-store.ts`'s `buildSessionSummary()` builds the timeline by iterating tasks in array order and appending each task's tool calls in that task's array order — an append-only construction, not a sorted one. `TaskDetector`'s internal arrays are also append-only. In the common case (a single-threaded hook stream recorded in real time) this happens to be chronological — but nothing in the type system, runtime, or test suite enforces it. No test anywhere (`replay-analyzer.test.ts`, `api-handler.test.ts`, `session-store.test.ts`) constructs an out-of-order timeline to catch a violation.

If this precondition is ever violated, anti-pattern segment boundaries and iteration counts would be silently wrong, with no error raised — a UI-correctness bug, not a crash.

## Suggested fix

Add the same `.sort((a, b) => a.timestamp - b.timestamp)` the two fallback branches already use, either inside `analyzeReplayTimeline()` itself (defensive, closes the gap for any future caller) or at the persisted-session call site in `buildReplayResponse()`.

## Scope note

Found during the "group 8" subsystem audit (2026-07-11, bundled pass over `src/dashboard/`, `src/digest/`, `src/tracing/`, `src/transport/`, `src/security/`). See `docs/audits/2026-07-11-group8-audit.md` on branch `docs/group8-audit-2026-07-11` for full findings context.

### LINKED ISSUE #165: static-handler.ts's explicit traversal check misses backslash-only payloads (defense-in-depth currently covers it)
## Summary

`src/dashboard/routes/static-handler.ts`'s explicit "reject any path segment equal to `'..'`" pre-check (line 61: `filename.split('/').some((c) => c === '..')`) only splits the incoming path on `/`. A request path built entirely from backslash-delimited `..` segments (no forward slashes at all, e.g. `..\..\Windows\System32\config\SAM`) passes this specific check unchecked — confirmed directly: `'..\\..\\Windows\\...'.split('/')` yields a single-element array that never equals `'..'`.

## Why this isn't an active exploit today

The code's second containment layer (`path.relative(root, target)` + `startsWith('..')`/`isAbsolute()`, using the OS-native `node:path` module) independently catches the same payload:
- **On POSIX** (this project's actual runtime/test environment), a literal backslash is never a path separator, so the whole payload resolves as one literal (nonexistent, harmless) filename component that cannot escape `root` at all — there is nothing for the second layer to "catch," the risk is structurally absent.
- **On a genuine Windows deployment**, `node:path`'s native `win32` mode treats `\` as a separator, so the second-layer check independently rejects the same payload and still returns `403` — verified directly.

So there is real defense-in-depth and no live exploit path was found on either platform. But the explicit, CodeQL-visible sanitizer this code's own comments describe as load-bearing has a real gap for this input shape, and the dedicated `static-handler.windows-sep.test.ts` file only pins a raw Node.js `path.win32` API — it never calls `createStaticHandler()` — so nothing in the test suite proves the end-to-end handler is safe under real Windows path semantics (this repo also has no Windows CI runner).

## Suggested fix

Extend the explicit first-layer check to also split on `\`, closing the gap in the primary sanitizer rather than relying solely on the fallback layer.

## Scope note

Found during the "group 8" subsystem audit (2026-07-11, bundled pass over `src/dashboard/`, `src/digest/`, `src/tracing/`, `src/transport/`, `src/security/`). See `docs/audits/2026-07-11-group8-audit.md` on branch `docs/group8-audit-2026-07-11` for full findings context.

### LINKED ISSUE #166: Dashboard hygiene: registerRoute() dead code + stale setMaxListeners comment
## Summary

Two small, unrelated hygiene findings in the dashboard subsystem, bundled here since both are low-severity dead-code/stale-comment issues:

**1. `DashboardServer.registerRoute()` is dead code** (`src/dashboard/dashboard-server.ts:130-132`). It's a public extension point for registering custom routes, but nothing in `src/index.ts` (the only production consumer of `DashboardServer`) ever calls it, and no test exercises it either. A regression here (e.g. a typo in the route-key format) would go completely undetected.

**2. `live-event-bus.ts`'s `setMaxListeners(200)` comment is stale and, once "corrected," would still be wrong.** The comment claims "each SSE connection adds 4 listeners," giving headroom for ~50 connections — but `sse-handler.ts` actually registers 5 listeners per connection (`context-update` was added after the comment was written and never updated). The naive fix (200÷5=40 connections) is *also* wrong: `EventEmitter.setMaxListeners(n)` caps the listener count **per distinct event name**, not in aggregate, and each connection's 5 listeners are registered under 5 separate event names — so the real headroom is **200 concurrent connections**, not 40 or 50. Not a functional bug (200 is ample headroom under any of these numbers), but the comment misstates both the listener count and the resulting capacity math.

## Suggested fix

- Remove `registerRoute()` (no callers), or wire it up to something real and add test coverage.
- Fix the `setMaxListeners(200)` comment to say 5 listeners per connection, and correct the capacity claim to 200 concurrent connections (not 40 or 50) given `setMaxListeners`' per-event-name semantics.

## Scope note

Found during the "group 8" subsystem audit (2026-07-11, bundled pass over `src/dashboard/`, `src/digest/`, `src/tracing/`, `src/transport/`, `src/security/`). See `docs/audits/2026-07-11-group8-audit.md` on branch `docs/group8-audit-2026-07-11` for full findings context.

chrisdhaan:
## Summary

- `buildReplayResponse()`'s main persisted-session path (serves every completed/historical session's replay view) passed its timeline straight to `analyzeReplayTimeline()` with no sort, unlike its two fallback branches which both sort first. Nothing upstream guarantees a persisted timeline is chronological, so an out-of-order timeline could silently produce wrong anti-pattern segment boundaries with no error raised. The persisted-session path now sorts before analysis, matching the fallback branches.
- `static-handler.ts`'s explicit path-traversal pre-check only split incoming paths on `/`, so a request built entirely from backslash-delimited `..` segments passed that specific check unchecked. The check now also splits on `\`, closing the gap in the primary sanitizer. A second, independent containment layer already made the handler safe in practice; this fix closes the gap in the explicit, static-analysis-visible check without touching that second layer.
- Removed `DashboardServer.registerRoute()`, a public method with no callers and no test coverage. Corrected a stale comment on `LiveEventBus`'s `setMaxListeners(200)` call that understated both the per-connection listener count (5, not 4) and the resulting connection headroom (200 concurrent connections, not 40 or 50).
- Bumps version to 1.4.23 and adds a CHANGELOG entry.

Fixes #162. Fixes #165. Fixes #166.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3938 passed, 3 pre-existing skips, 0 failures
- [x] `npm run lint` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.23] - 2026-07-12
+
+### Fixed
+
+- `buildReplayResponse()`'s main persisted-session path passed its timeline straight to `analyzeReplayTimeline()` with no sort, unlike its two fallback branches which both sort before analysis. Nothing upstream guarantees a persisted timeline is chronological (session persistence is append-only), so an out-of-order timeline could silently produce wrong anti-pattern segment boundaries and iteration counts with no error raised. The persisted-session path now sorts before analysis, matching the fallback branches.
+- `static-handler.ts`'s explicit path-traversal pre-check only split incoming paths on `/`, so a request built entirely from backslash-delimited `..` segments (no forward slashes at all) passed this specific check unchecked. The check now also splits on `\`, closing the gap in the primary sanitizer.
+- Removed `DashboardServer.registerRoute()`, a public method with no callers and no test coverage. Corrected a stale comment on `LiveEventBus`'s `setMaxListeners(200)` call that understated both the per-connection listener count (5, not 4) and the resulting connection headroom (200 concurrent connections, not 40 or 50 — `setMaxListeners` caps per event name, not in aggregate).
+
 ## [1.4.22] - 2026-07-11
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.22",
+  "version": "1.4.23",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.22",
+      "version": "1.4.23",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.22",
+  "version": "1.4.23",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/dashboard/dashboard-server.ts
+++ b/src/dashboard/dashboard-server.ts
@@ -127,10 +127,6 @@ export class DashboardServer {
     });
   }
 
-  registerRoute(method: 'GET' | 'POST', path: string, handler: RouteHandler): void {
-    this.routes.set(`${method} ${path}`, handler);
-  }
-
   /** Phase 1 hook for tests + Phase 3 API route wiring. */
   getAlertEngine(): LocalAlertEngine | undefined {
     return this.opts.alertEngine;
--- a/src/dashboard/live-event-bus.ts
+++ b/src/dashboard/live-event-bus.ts
@@ -119,10 +119,11 @@ export class LiveEventBus {
 
   constructor(opts: LiveEventBusOptions = {}) {
     this.bufferSize = opts.replayBufferSize ?? DEFAULT_BUFFER_SIZE;
-    // Each SSE connection adds 4 listeners (tool-call, cost-update,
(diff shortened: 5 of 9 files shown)


corinnekunze:

PR #174, opened 2026-07-12
Fix proxy SSRF/DNS-rebinding gap + error-message leakage (v1.4.24)

### LINKED ISSUE #143: SSRF protection checks the literal hostname string, not the resolved IP — DNS rebinding is not actually prevented
## Description

`SECURITY.md` and an in-code comment both claim proxy-mode's SSRF protection prevents DNS rebinding. It doesn't — the implementation checks the literal hostname string, never the IP address the connection actually resolves to.

## Root cause

`validateSsrfUrl` (`src/security/ssrf.ts:109-146`) operates entirely on `url.hostname` — a JS string — via a regex/FQDN blocklist and numeric-encoding canonicalization. It never calls `dns.lookup()`/`resolve4`/`resolve6`, and never inspects the socket the eventual request connects to. Confirmed via grep: zero `dns` references anywhere in `src/proxy/` or `src/security/ssrf.ts`.

It's called twice per request — at upstream-construction time and again immediately before every forward (`src/proxy/upstream-http.ts:81-83`, comment: "Re-validate URL against SSRF rules immediately before fetch to prevent DNS rebinding") — but both calls re-check the *same unchanged string*. Node's own `http`/`https` client performs OS-level DNS resolution at connect time, entirely decoupled from either check.

**Concrete bypass**: a hostname an attacker controls (e.g. via a DNS record they own) can resolve to a public IP when an upstream is registered — passing both checks — then be repointed to `127.0.0.1` or a cloud metadata IP (`169.254.169.254`) before the request is actually sent. The pre-fetch "re-validation" cannot catch this: it tests the same string, which is unchanged.

This same vulnerability is reached from a second call site: `src/proxy/otlp-receiver.ts`'s `forward()` calls the identical `validateSsrfUrl` helper before every outbound forward to New Relic.

## Impact

A real, confirmed-in-principle security gap. Mitigating context: proxy mode is opt-in, and upstreams/forward endpoints are typically configured against known, trusted servers rather than arbitrary attacker-supplied URLs — so this is not a maximally severe, drop-everything issue. But the documented mitigation genuinely does not work as claimed.

## Fix

Resolve the hostname via `dns.lookup()`, validate the resolved IP(s) against the same RFC-1918/loopback/metadata rules already applied to the string, and pin the connection to that resolved address (via a custom `lookup` option, `family`/`hints`, or a post-connect socket-address check) rather than re-trusting the OS resolver at connect time. One fix in `src/security/ssrf.ts` closes the gap at both call sites (`upstream-http.ts` and `otlp-receiver.ts`) simultaneously. Correct the overclaiming comment at `upstream-http.ts:80` and the `SECURITY.md` wording regardless of fix timing.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11). Given the security nature, consider routing this through whatever this project's normal vulnerability-handling process is before it becomes a public changelog line item.

### LINKED ISSUE #144: Proxy upstreams leak raw error messages (host:port, upstream name, child-process errors) to HTTP clients
## Description

`SECURITY.md` has an explicit checklist item: error responses to HTTP clients must contain only a generic code, with detail going to the logger. Two upstream transport implementations violate this.

## Findings

1. **`upstream-http.ts`** — connection errors (`ECONNREFUSED` etc.) and timeout errors both funnel through the same handler (`upstream-http.ts:214-229`), which returns `{ error: 'upstream_unavailable', message: String(err) }` to the HTTP client. For a connection error this includes the target host:port (e.g. `connect ECONNREFUSED 127.0.0.1:1`); for a timeout it includes the configured upstream *name*. `SECURITY.md` explicitly gives `No route for ${url}` as a "Bad" example — `message: String(err)` here is the same category of leak.
2. **`upstream-stdio.ts`** — `dispatchToClient()`'s catch block (`upstream-stdio.ts:266-282`) sends the raw `err.message` from the child process straight to the HTTP client via `writeJsonRpcError(res, rpc.id, -32603, message)` (line 275). This could include file paths, stack-trace-adjacent detail, or arguments echoed back by a misbehaving tool. Confirmed as intended-but-undocumented-as-a-violation behavior: the existing test suite explicitly asserts the raw message reaches the client (`upstream-stdio.test.ts:266-282`, `parsed.error.message === 'Process crashed'`).

## Impact

Severity is low in the current single-operator deployment model — the HTTP client is the proxy operator's own Claude Code, not an arbitrary third party. It becomes a real information-disclosure issue if the proxy is ever exposed to less-trusted callers.

## Fix

Both sites need the same shape of fix: log the detail server-side (already happening), return a generic error code/message to the HTTP client. `src/proxy/otlp-receiver.ts` already gets this right — its 500 response on a forward failure carries no body at all — copy that pattern.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11). Worth fixing alongside #143 (SSRF), since both touch overlapping error-handling code paths in `upstream-http.ts`.

chrisdhaan:
## Summary

- SSRF protection validated the literal hostname string of a configured upstream/forward URL, but re-checked the exact same unchanged string a second time immediately before every network call, providing no real protection against DNS rebinding (a hostname resolving to a safe address when first validated, then to `127.0.0.1` or a cloud metadata address before the connection is actually made). Both `HttpUpstream` (proxy forwarding) and `OtlpReceiver` (OTLP forward endpoint) now resolve the hostname exactly once at connection time via a custom DNS lookup, validate every resolved address against the same blocklist rules, and pin the actual connection to that validated address.
- Two proxy upstream transports returned raw error detail to HTTP clients on failure: `HttpUpstream` included the literal connection-error text (which can contain the upstream's host:port) in its JSON error body, and `StdioUpstream` forwarded a failed child process's raw error message (which can include file paths or tool-echoed arguments) straight through as the JSON-RPC error message. Both now return only a generic error code/message to the client; the full detail is still logged server-side.
- Bumps version to 1.4.24 and adds a CHANGELOG entry.

Fixes #143. Fixes #144.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 3946 passed, 3 pre-existing skips, 0 failures
- [x] `npm run lint` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.24] - 2026-07-12
+
+### Fixed
+
+- Standalone proxy mode's SSRF protection validated the literal hostname string of a configured upstream/forward URL, but re-checked the exact same unchanged string a second time immediately before every network call — providing no real protection against DNS rebinding (a hostname resolving to a safe address when first validated, then to `127.0.0.1` or a cloud metadata address before the connection is actually made). Both `HttpUpstream` (proxy forwarding) and `OtlpReceiver` (OTLP forward endpoint) now resolve the hostname exactly once at connection time via a custom DNS lookup, validate every resolved address against the same blocklist rules, and pin the actual connection to that validated address.
+- Two proxy upstream transports returned raw error detail to HTTP clients on failure: `HttpUpstream` included the literal connection-error text (which can contain the upstream's host:port) in its JSON error body, and `StdioUpstream` forwarded a failed child process's raw error message (which can include file paths or tool-echoed arguments) straight through as the JSON-RPC error message. Both now return only a generic error code/message to the client; the full detail is still logged server-side.
+
 ## [1.4.23] - 2026-07-12
 
 ### Fixed
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -147,7 +147,9 @@ if (!ALLOWED_SCHEMES.has(this.url.protocol)) throw new Error('…scheme not allo
 if (BLOCKED_HOST_RE.test(this.url.hostname)) throw new Error('…private addresses not allowed');
 ```
 
-This is used by both `HttpUpstream` (MCP proxy forwarding) and `OtlpReceiver` (`otlpForwardEndpoint` config). Any new network client that takes a URL from config or user input should call `validateSsrfUrl()` at construction time.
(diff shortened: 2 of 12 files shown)


corinnekunze:

PR #175, opened 2026-07-12
Fix proxy OTLP receiver hardening (v1.4.25)

### LINKED ISSUE #146: OTLP receiver enrichment is JSON-only; docs overstate coverage for the common protobuf case
## Description

`docs/ADVANCED.md:45` claims: "Any OTel-instrumented app pointing at `http://localhost:4318` will have its telemetry enriched with the current coding session context... and forwarded to NR." This is only true for JSON-encoded OTLP bodies.

## Root cause

`enrichPayload()` (`src/proxy/otlp-receiver.ts:308-320`) wraps `JSON.parse` in a try/catch; on failure (i.e., a protobuf-encoded body, which is the default wire format for most production OTel SDKs — Node, Python, Java) the body is forwarded completely unmodified, with a comment "Not JSON (likely protobuf) — forward as-is." `docs/ADVANCED.md:74` does separately note "protobuf payloads are forwarded as-is," but that caveat is 29 lines below the headline claim and never linked back to it — so the sweeping claim at line 45 reads as unconditional when it's actually true only for a subset of real-world exporter configurations.

## Fix

Either add the JSON-only caveat directly to the headline claim in `docs/ADVANCED.md`, or (larger effort) add protobuf support to `enrichPayload()` so the feature works for the common case it's currently silently skipping.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11).

### LINKED ISSUE #147: OTLP receiver: auth timing side-channel, unbounded rate-limiter memory growth, and an undocumented config field that worsens both
## Description

Three related, individually low-severity gaps in the local OTLP/HTTP receiver, bundled here since fixing the third (documentation) affects how seriously to treat the first two.

## Findings

1. **Auth timing side-channel.** The Bearer-token check (`src/proxy/otlp-receiver.ts:251-268`) correctly uses `timingSafeEqual` for content comparison, but it's gated behind a plain `authHeader.length === expectedAuth.length` check that short-circuits *before* `timingSafeEqual` runs — leaking the correct key's length (not its content) via response-time difference. Largely theoretical under the documented `127.0.0.1`-only default bind address (the realistic threat model is another local process, and sub-microsecond timing over HTTP isn't practically measurable) — but real defense-in-depth gap.
2. **Rate-limiter memory growth.** `checkRateLimit()` (`otlp-receiver.ts:270-296`) prunes each IP's timestamp array down to empty but never removes the empty array from the outer `Map` (`otlp-receiver.ts:40`). Every distinct source IP that ever makes one request leaves a small (~100-150 byte) permanent entry for the process lifetime. Slow and low-severity under the loopback-only default; would matter far more under sustained traffic from many distinct IPs.
3. **`otlpReceiverBindAddress` is wired but completely undocumented.** Fully wired end-to-end (env `NR_AI_OTLP_RECEIVER_BIND_ADDRESS`, `src/config.ts:752-758`, default `127.0.0.1`) but appears in neither `docs/ADVANCED.md` nor `CLAUDE.md`'s config table. This matters beyond a simple omission: an operator who discovers and changes it to a non-loopback address (e.g. `0.0.0.0`) would materially worsen the severity of findings 1 and 2 above, with no documentation warning them of that trade-off.

## Fix

1. Compare lengths via `timingSafeEqual` too (e.g. pad to a fixed length before comparing, or use a length-independent constant-time comparison), or accept the minor risk explicitly in a comment.
2. Delete the `Map` entry when an IP's timestamp array empties out.
3. Document `otlpReceiverBindAddress` in `ADVANCED.md`/`CLAUDE.md`'s config table, explicitly noting the severity trade-off of changing it from `127.0.0.1`.

## Scope note

Found during a `src/proxy/` subsystem audit (2026-07-11). None of these are covered by the existing 41-test suite for this file — add regression tests alongside whichever fixes land.

chrisdhaan:
## Summary

- `docs/ADVANCED.md`'s inbound OTLP receiver claim overstated coverage — "any OTel-instrumented app" is enriched with session context, but this only applies to JSON-encoded OTLP payloads. Most production OTel SDKs (Node, Python, Java) default to protobuf, which has always been forwarded unmodified. The headline claim now states the JSON-only scope directly.
- The OTLP receiver's Bearer-token auth check short-circuited on a header-length comparison before running its timing-safe content comparison, leaking the correct API key's length via response-time differences. Both sides are now hashed to a fixed-length digest before comparison, removing the length-dependent branch entirely.
- The OTLP receiver's per-IP rate limiter pruned expired request timestamps down to an empty array but never removed that array from its internal map, so every distinct source IP that ever made one request left a small permanent entry for the life of the process. Empty entries are now deleted instead of retained.
- Documented the previously-unlisted `otlpReceiverBindAddress` config field, noting that widening it beyond the `127.0.0.1` default increases exposure to the two fixes above.
- Version bump to 1.4.25.

## Test plan

- [x] `npm run build && npm test` — all tests passing, including 2 new regression tests for the auth-timing and rate-limiter fixes
- [x] `npm run lint` clean

Fixes #146. Fixes #147.


corinnekunze:

PR #176, opened 2026-07-12
Install dead-code hygiene (v1.4.26)

### LINKED ISSUE #158: json-utils.ts's readJsonFile() is a dead function whose only historical use caused a real credential-wipe bug
## Summary

`json-utils.ts:17-23` exports a lenient JSON reader, `readJsonFile()`, that silently returns `{}` on any read error (including `EACCES`/permission errors, not just missing-file). It is exported but has **zero production callers anywhere in `src/`** (confirmed via `grep -rn "readJsonFile\b" src/`) — the only hits are its own definition and a comment in `cli.test.ts:1155-1157`.

That comment documents that a prior version of `clearSavedPlatform` used exactly this function, and that usage caused a real bug: silently returning `{}` on an unreadable `config.json` caused the subsequent `writeJsonFile` call to overwrite the file with `{}`, wiping the user's `licenseKey` and `accountId`. The call site was fixed by switching to the stricter `readJsonFileStrict()` (`json-utils.ts:29-43`, which throws on non-ENOENT errors instead of swallowing them).

## Why this matters

The original lenient function that caused the bug was never removed, deprecated, or annotated with a warning — it remains exported under an innocuous name (no doc comment warning against its use, unlike `readJsonFileStrict`'s comment explaining why it's the correct choice) and available for a future caller, in this file or a new one, to reach for and silently reintroduce the exact same class of data loss.

Nothing about `readJsonFile()`'s own behavior is factually wrong — it matches its documented contract exactly. This is a hygiene/latent-risk finding, not a live functional bug.

## Suggested fix

Either remove `readJsonFile()` entirely (it has no current callers), or add a doc comment explicitly warning against its use and pointing to `readJsonFileStrict()` as the safe alternative.

## Scope note

### LINKED ISSUE #159: src/install/index.ts barrel module has zero reachable consumers
## Summary

`src/install/index.ts` re-exports `generateHookEntries`, `generateMcpServerEntry`, `generateNrConfig`, `mergeSettings`, `removeSettings`, `detectSettingsPath`, `HookEntry`, `HookEntries`, `McpServerConfig`, `NrObserveConfig` from `install-helper.js`, and `runInstallCli`, `createInstallProgram` from `cli.js`. Every re-export is individually correct (confirmed each resolves to a real, correctly-named source symbol) — but a repo-wide grep confirms **no file in `src/` imports from `./install/index.js`**. Both real call sites for what it re-exports (`src/index.ts:362`, `src/install/setup-wizard.ts:13`) import `runInstallCli` directly from `./install/cli.js`, bypassing the barrel entirely.

`package.json`'s `exports` field also doesn't expose `dist/install/index.js` — only the top-level `dist/index.js` is reachable by external consumers. This file compiles into `dist/install/index.js` as a side effect of `tsc` but has zero reachable consumers, internal or external.

## Why this matters

Nothing this file claims is factually wrong — it's a dead-code/orphaned-module finding, not a functional defect. Low severity, but worth cleaning up as accumulated dead surface area (a second, related finding from the same audit is `json-utils.ts`'s unused `readJsonFile()`, see linked issue).

## Suggested fix

Delete the file, or if there's a reason to keep it as a documented public re-export surface for future consumers, say so in a comment.

## Scope note

chrisdhaan:
## Summary

Removes two confirmed-dead code artifacts in `src/install/`, and bumps the version to 1.4.26.

- Removes `readJsonFile()`, a lenient JSON reader in `src/install/json-utils.ts` with zero production callers. Its only historical caller was replaced by the stricter `readJsonFileStrict()` after a real credential-wipe bug (a permission error silently swallowed into `{}`, then written back over `config.json`, erasing `licenseKey`/`accountId`). Fixes #158
- Removes `src/install/index.ts`, a barrel module with zero reachable consumers — every real call site already imports directly from `./cli.js` or `./install-helper.js`, and `package.json`'s `exports` field never exposed the barrel externally either. Fixes #159
- Bumps `package.json`/`package-lock.json` to 1.4.26 and adds a CHANGELOG entry.

No behavior change — pure dead-code removal.

## Test plan

- [x] `npm run build` — clean
- [x] `npm run lint` — clean


corinnekunze:

PR #177, opened 2026-07-12
Fix #151, #152: Storage cleanup — dead code + redaction consistency

### LINKED ISSUE #151: LocalStore.saveSession() and loadRecentSessions() are dead code in production
## Description

`LocalStore` carries a parallel session-persistence implementation that nothing in the running application actually uses.

## Root cause

`saveSession()` (`src/storage/local-store.ts:886-898`) and `loadRecentSessions()` (`:900-930`) have zero non-test call sites — confirmed via grep across all of `src/`. Every real session-persistence path in `src/index.ts` goes through the separate `SessionStore` class (`src/storage/session-store.ts`) instead, which has its own independent `writeFileSync`/`mkdirSync`/path-escape-guard implementation and a different filename scheme (`${date}_${sessionId}.json` vs. `LocalStore`'s `${sessionId}.json`).

Both methods are exercised only by `local-store.test.ts`.

## Fix

Decide whether to delete these two methods (and their tests), or document why `LocalStore` still carries a parallel, unused session-persistence implementation alongside `SessionStore`'s.

## Scope note

### LINKED ISSUE #152: session-store.ts: redaction inconsistency for file paths, plus a narrow cross-process write collision
## Description

Two related session-store.ts hardening gaps found during audit, bundled here given their size.

## Findings

1. **`filesRead`/`filesModified` arrays are persisted unredacted while the parallel `timeline[].filePath` field for the same underlying data IS redacted.** `buildSessionSummary()` (`session-store.ts:359-360`) populates `filesRead: [...allFilesRead].sort()` and `filesModified: [...allFilesModified].sort()` directly with no call to `redactSensitive()`, while `timeline[].filePath` (line 336) does call it — both are sourced from the same underlying `task.filesRead`/`task.filesModified` data. **Also unredacted, same open question**: `sessionName`/`repoName` (derived from `basename(cwd)`) — free-text-adjacent fields with no design statement either way on whether they should be redacted.
   - Practical severity is low: `redactSensitive()`'s patterns (`src/config.ts:89-106`) match secret *values* (API-key prefixes, JWTs, connection-string userinfo) — a plain filesystem path essentially never matches unless it happens to literally contain a secret-shaped substring. But it's a real, verifiable inconsistency between two representations of the same data within one persisted document.
2. **A narrow cross-process same-sessionId write collision.** If a resumed/forked session opens two MCP processes against the same real session ID, both processes' single `writeFileSync` calls target the identical filename and the later shutdown silently wins — no corruption (each write is a full-file overwrite, not a partial merge), but the earlier session's summary is lost with no error. Reachability in a real resume/fork scenario is not yet confirmed — this is a theoretical gap worth investigating, not a settled bug.

## Fix

1. Decide whether `filesRead`/`filesModified`/`sessionName`/`repoName` should go through `redactSensitive()` for consistency with `timeline[].filePath`, and apply uniformly.
2. Investigate whether the resume/fork same-sessionId scenario is actually reachable in practice; if so, consider a lock/warning mechanism.

## Scope note

chrisdhaan:
## Summary

- Removes `LocalStore.saveSession()` / `loadRecentSessions()`, a parallel session-persistence implementation with zero non-test callers — real persistence goes through `SessionStore` instead. Fixes #151.
- `buildSessionSummary()` redacted `timeline[].filePath`/`.command` via `redactSensitive()` but persisted `filesRead`/`filesModified`/`sessionName`/`repoName` unredacted, despite being sourced from the same underlying task data. All four now route through the same redaction call. Fixes #152.
- `SessionStore.saveSession()` now logs a warning before overwriting an existing session file for the same `sessionId`+date — the scenario a resumed/forked session running two MCP processes against one real session ID would produce. Overwrite behavior (last-write-wins) is unchanged; this is visibility-only, no locking mechanism. Fixes #152.
- Version bump to 1.4.27.

## Test plan

- [x] `npm run build` — succeeds
- [x] `npm test` — 3947/3950 passing (3 pre-existing skips), 0 failures
- [x] New tests added for redaction consistency and the overwrite-warning log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.27] - 2026-07-12
+
+### Fixed
+
+- `LocalStore.saveSession()` and `loadRecentSessions()` were a parallel, unused session-persistence implementation with zero non-test call sites — every real session-persistence path goes through the separate `SessionStore` class instead. Removed both methods and their tests.
+- `buildSessionSummary()` redacted `timeline[].filePath` via `redactSensitive()` but persisted `filesRead`/`filesModified`/`sessionName`/`repoName` unredacted, even though they're sourced from the same underlying task data. All four now go through the same redaction call for consistency.
+- `SessionStore.saveSession()` now logs a warning when it's about to overwrite an existing session file for the same `sessionId`+date — the scenario a resumed/forked session running two MCP processes against one real session ID would produce. The overwrite behavior (last-write-wins) is unchanged; this only makes a previously-silent collision visible.
+
 ## [1.4.26] - 2026-07-12
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.26",
+  "version": "1.4.27",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.26",
+      "version": "1.4.27",
       "license": "Apache-2.0",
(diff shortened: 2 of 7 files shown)


corinnekunze:

PR #178, opened 2026-07-12
Fix #150, #163: Docs accuracy — LocalStore attribution + digest scheduling

### LINKED ISSUE #150: CLAUDE.md's "LocalStore handles atomic buffer operations" claim describes the wrong class
## Description

CLAUDE.md states: "`LocalStore` handles atomic buffer operations (append, drain with rename-then-read pattern)." The real production hook-event writer is not `LocalStore` at all.

## Root cause

`LocalStore.appendToBuffer()` (`src/storage/local-store.ts:124-131`) has zero production call sites — confirmed via grep across all of `src/`, only referenced from its own test file. The real writer is `src/hooks/collector-script.ts` (lines ~309-330 and ~941-961), which does its own independent raw `openSync(path, O_WRONLY|O_CREAT|O_APPEND, 0o600)` + `writeFileSync` + `closeSync`, never calling into `LocalStore`.

The two implementations are atomically equivalent (both use `O_APPEND` semantics), so there's no live *correctness* gap between them — but the documentation attributes real, running-system behavior to a class/method the system doesn't actually use for that purpose. The drain side (`drainBuffer()`/`drainPath()`'s main path) genuinely is correct and genuinely is in `LocalStore` — only the append-side attribution is wrong.

## Fix

Either update CLAUDE.md to accurately describe `collector-script.ts` as the real buffer-append implementation, or have `collector-script.ts` actually delegate to `LocalStore.appendToBuffer()` (reducing duplicate atomicity-sensitive logic to one implementation, which would also make the doc claim true again).

## Scope note

### LINKED ISSUE #163: Docs claim weekly digest delivery is automatic; no scheduler exists, delivery is manual-only
## Summary

`docs/COMMANDS_TABLE.md`'s example response for `nr_observe_subscribe_digest` states "Digest will be sent on the configured schedule," and `docs/PRODUCT_BRIEF.md` describes weekly digest delivery as "automatic." Neither is true today.

`digestSchedule` (a cron-expression config field) is stored and read only by config-loading plumbing and the dashboard's settings UI/API — nothing in `src/` ever consumes it to actually trigger a send. Confirmed via repo-wide grep: no cron job, no `setInterval`, nothing that reads `digestSchedule` and calls `sendSlackDigest()`. The only real send path is the manual `nr_observe_send_digest` MCP tool, called on-demand.

## Decision

Discussed during the "group 8" subsystem audit (2026-07-11) — **decided to fix the docs rather than build the scheduler**, since building a reliable in-process scheduler for a long-running MCP server (crash/restart semantics, missed-window handling, etc.) is a real feature project, not a quick fix, and the manual `nr_observe_send_digest` path already covers the core need.

## Suggested fix

Update `docs/COMMANDS_TABLE.md`'s `nr_observe_subscribe_digest`/`nr_observe_unsubscribe_digest` entries and `docs/PRODUCT_BRIEF.md` to state that digest delivery is manual-only today via `nr_observe_send_digest`, and that `digestSchedule` is currently stored for future use but not yet wired to an automatic trigger.

## Scope note

Found during the "group 8" subsystem audit (2026-07-11, bundled pass over `src/dashboard/`, `src/digest/`, `src/tracing/`, `src/transport/`, `src/security/`). See `docs/audits/2026-07-11-group8-audit.md` on branch `docs/group8-audit-2026-07-11` for full findings context.

chrisdhaan:
## Summary

Two doc-accuracy fixes — no source behavior change except one string literal.

- CLAUDE.md attributed hook-event buffer appends to `LocalStore`; the real writer is `collector-script.ts`'s own raw file-append logic (`openSync`/`writeFileSync`/`closeSync` with `O_APPEND`). `LocalStore` genuinely owns the drain side (rename-then-read) — only the append-side attribution was wrong. Fixes #150.
- `docs/COMMANDS_TABLE.md` and the `nr_observe_subscribe_digest` tool's own runtime response message both claimed weekly Slack digest delivery happens automatically on a configured schedule. No scheduler exists — delivery is manual-only via `nr_observe_send_digest`. Both now say so; `digestSchedule` is documented as stored for future use only. Fixes #163.
- Version bump to 1.4.28 + CHANGELOG entry.

## Test plan

- [x] `npm install && npm run build` passes
- [x] `npm test` — 129/130 suites, 3947/3950 tests passing
- [x] `npm run lint` — 1 error + 5 warnings, confirmed pre-existing and unrelated to this change
- [x] Every corrected doc claim independently verified against current source

🤖 Generated with subagent-driven-development (Claude Sonnet 5)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.28] - 2026-07-12
+
+### Fixed
+
+- CLAUDE.md attributed hook-event buffer appends to `LocalStore`; the real writer is `collector-script.ts`'s own raw file-append logic. `LocalStore` genuinely owns the drain side (rename-then-read) — only the append-side attribution was wrong. Docs corrected.
+- Multiple places claimed weekly Slack digest delivery happens automatically on a configured schedule: `docs/COMMANDS_TABLE.md`'s example response, the product documentation's feature description, and the `nr_observe_subscribe_digest` tool's own runtime response message. No scheduler exists — delivery is manual-only via `nr_observe_send_digest`. All three now say so; `digestSchedule` is documented as stored for future use only.
+
 ## [1.4.27] - 2026-07-12
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -139,7 +139,7 @@ Claude Code
   │
   ├─ PreToolUse / PostToolUse hooks
   │    └─> preflight (collector-script.ts)
-  │         └─> writes to buffer.jsonl (LocalStore)
+  │         └─> writes to buffer.jsonl directly (raw fd append)
   │
   └─ MCP stdio connection
        └─> NrMcpServer (server.ts)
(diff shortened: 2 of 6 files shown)


corinnekunze:

PR #181, opened 2026-07-13
Fix web-dashboard high-visibility bugs (v1.4.30)

### LINKED ISSUE #128: Alerts.tsx: threshold subtitle wrongly claims these values feed the local alert engine
## Description

The "Alert Thresholds" card's subtitle reads: "Thresholds for the local alert engine. Require a server restart to take effect." Both the system named and the trigger described are wrong.

## Root cause

`personalAlertThresholds` (`dailyCostUsd`/`sessionCostUsd`/`efficiencyScoreMin`/`antiPatternCountMax`/`stuckLoopCountMax`) is never read by `LocalAlertEngine` — confirmed via grep across `src/index.ts`, `src/dashboard/dashboard-server.ts`, and `src/alerts/local-alert-engine.ts`. `LocalAlertEngine`'s rules load exclusively from a separate `rules.json` file (`config.alerts.rulesPath`) with its own schema (`src/alerts/local-alert-rule.ts`), never edited by this UI.

The only real consumer of these five values is `src/deploy/deploy-alerts.ts`'s `loadPersonalThresholds()`, which templates NerdGraph **cloud** alert conditions — used solely by the separate, manual `preflight deploy-alerts` CLI subcommand (`npm run deploy:alerts`). Restarting the dashboard/MCP server does not invoke that command and does not make these values affect anything locally.

`Alerts.tsx:239` (subtitle) and `:265-269` (the "Saved. Restart server..." confirmation message shown after a save) both repeat this false premise.

## Impact

A user editing these fields and restarting the server observes zero behavior change in the actual live alert system. The values only matter if they later, separately, run `deploy-alerts` against a New Relic account. This is the most actively misleading finding in a recent dashboard audit — it doesn't just describe a wrong number, it tells the user an action will do something it structurally cannot do.

## Fix

Correct both strings to accurately describe the real consumer (the `deploy-alerts` CLI, cloud alert conditions) or, if the values are meant to eventually feed `LocalAlertEngine`, wire that up for real. Either way, `Alerts.tsx` currently has zero test coverage (`Alerts.test.tsx` does not exist) — any fix here should add tests alongside it, or this exact defect can silently regress.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11), independently re-verified against `local-alert-engine.ts`/`deploy-alerts.ts`/`index.ts` three separate times before being reported.

### LINKED ISSUE #129: Sessions.tsx: "Outcome" card always shows a hardcoded value, never a real classification
## Description

The session detail page's "Outcome" card renders `data.outcome` as if it reflects a real classification of how the session went. In practice, every real code path hardcodes it to one of exactly two values.

## Root cause

`buildSessionSummary()` (`src/storage/session-store.ts:385`) writes `outcome: 'completed'` unconditionally for every persisted session, with no branching logic. The live-own-session branch (`src/dashboard/routes/api-handler.ts:1664`) and the live-registry-synthesized branch (`:1711`) both hardcode `outcome: 'in progress'`. No other code path produces any other value.

`Sessions.test.tsx`'s fixtures use values like `'feature'`/`'bug_fix'`/`'exploration'` (lines 54, 61, 197, 211), implying a real outcome-classification field exists and is wired — it does not; those are test-only values no real code path ever produces.

## Impact

The card labeled "Outcome" carries zero information beyond "is this session still running," on every session detail page a user visits. A user reasonably reads "completed"/"in progress" as a quality/outcome signal it was never designed to be.

## Fix

Either implement real outcome classification (would need new logic — no existing heuristic in the codebase for this), or remove/relabel the card to accurately reflect that it's a running-state indicator, not an outcome classification.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11).

### LINKED ISSUE #130: Sessions.tsx: "Session Quality" and "Tool Selection" cards never render — backing fields never attached to the response
## Description

The session detail page has two fully-implemented cards — "Session Quality" (Diff Apply/Test Pass/Backtracks/Self-corrections) and "Tool Selection" (score + breakdown) — that never render in the running application, regardless of session state.

## Root cause

Both cards are gated on `data.qualityProxy` / `data.toolSelectionScore` (`Sessions.tsx:412-461`). `QualityProxyTracker` and `ToolSelectionScorer` are real, live-fed trackers — actively used by the Today view and by their own standalone routes (`GET /api/quality-proxy`, `GET /api/tool-selection-score`, `api-handler.ts:1127-1141`). But nothing ever merges their output onto any of the three `/api/sessions/:id` response branches (`api-handler.ts:1634-1721` — persisted via `sessionStore.loadSession()`, own-live, and registry-synthesized; all three confirmed to omit both fields by reading each object literal in full). `data.qualityProxy`/`data.toolSelectionScore` are always `undefined` on this component.

The combined guard wrapping both cards (`Sessions.tsx:412`) means the entire `grid-cols-2` section — container included — never renders at all.

## Impact

Two large, fully-built, visually plausible sections of the most detailed session view in the app are permanently invisible. Likely the most user-visible "missing feature that looks like it should be there" finding in a recent dashboard audit.

## Fix

Merge `qualityProxyTracker.getMetrics()` and `toolSelectionScorer.scoreSession(...)` output onto the relevant `/api/sessions/:id` response branches in `api-handler.ts`, mirroring how the Today view already gets this data. No test in `Sessions.test.tsx` currently exercises either field, which is consistent with them never being wired — add coverage alongside the fix.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11). This is one of three independently-confirmed instances in that audit of the same underlying pattern (a real backend field silently dropped before reaching a specific view's response) — see also #133 (Audit.tsx's dropped severity).

### LINKED ISSUE #131: Today.tsx: anti-pattern banner renders literal '?' instead of a real count on the fallback path
## Description

Today.tsx's anti-pattern alert banner, when reached via its non-SSE API-fallback path, renders the literal character `?` in place of a real count for most anti-pattern types.

## Root cause

`Today.tsx:392-402`'s fallback chain reads `apiAntiPatterns[0].count ?? .iterations ?? .readCount ?? '?'`, never checking `repeatCount` or `editCount`. `SessionAntiPattern` declares both fields (`src/storage/types.ts:88-89`) and `AntiPatternDetector` (`src/metrics/anti-patterns.ts`) actually populates `repeatCount` for `stuck_loop` and `editCount` for `blind_editing` (and likely `agentCount` for `over_delegation`) — none of which match `count`/`iterations`/`readCount`. Real data exists on the object; the UI renders `'?'` instead, for 2-3 of 5 anti-pattern types.

The live-store (SSE-fed) branch is unaffected — this only hits the fallback path.

## Impact

A literal question mark rendered in place of a real number, on the most-trafficked view in the app, whenever this view is reached via the non-SSE fallback (e.g. before the SSE connection establishes).

## Fix

Extend the fallback chain to check `repeatCount`/`editCount` (and any other anti-pattern-type-specific count field `AntiPatternDetector` actually populates) before falling back to `'?'`.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11).

### LINKED ISSUE #132: Alerts.tsx: Unsubscribe leaves stale unsaved text in the webhook input, contradicting the Pill next to it
## Description

If a user edits the Slack webhook URL input (without saving) and then clicks "Unsubscribe," the input keeps showing the stale, never-saved URL while the adjacent "Configured"/"Not configured" Pill correctly flips to "Not configured" — a visibly inconsistent UI state.

## Root cause

Clicking "Unsubscribe" (`Alerts.tsx:334`) calls `saveMutation.mutate({ digestWebhookUrl: null })` directly, bypassing local `webhookUrl` edit-buffer state entirely. `saveMutation.onSuccess` (`:107`) only invalidates the settings query — it never resets `webhookUrl` back to `undefined`. Since the input's displayed value prioritizes local state whenever it's `!== undefined` (`:294-296`), the text box continues showing the unsaved value until the user manually clears or overwrites it again.

## Fix

Reset `webhookUrl` to `undefined` in `saveMutation.onSuccess` (or specifically in the Unsubscribe handler) so the input falls back to reflecting server truth immediately.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11). No `Alerts.test.tsx` exists — add coverage for this alongside the fix, and alongside #128 (the same view's threshold-subtitle fix), so neither regresses silently.

chrisdhaan:
## Summary

Five high-visibility dashboard bug fixes in `src/web/`, plus a version bump to 1.4.30.

- Fixes #128. Alerts.tsx's "Alert Thresholds" card wrongly claimed to feed "the local alert engine" and require a server restart. Corrected to describe the real consumer (the separate `deploy-alerts` CLI, which templates cloud alert conditions and reads config fresh on every run — no restart involved).
- Fixes #129. Sessions.tsx's "Outcome" card always showed one of exactly two hardcoded values (`completed`/`in progress`) with no real classification logic anywhere in the codebase. Relabeled as "Status" to accurately reflect it as a running-state indicator rather than a quality/outcome classification.
- Fixes #130. Sessions.tsx's "Session Quality" and "Tool Selection" cards were fully built but never rendered because the backend never attached the tracker data to any of the three `/api/sessions/:id` response branches. Now attached to the branches where the data is actually and correctly available: `qualityProxy` on the persisted branch (derived from the session's own stored timeline) and the own-live branch (via the live tracker), both gated on having a real signal; `toolSelectionScore` on the own-live and registry-synthesized branches, both correctly scoped to that session's own records (not the shared cross-session buffer) to avoid leaking another concurrent session's data.
- Fixes #131. Today.tsx's anti-pattern banner rendered a literal `?` instead of a real count for `stuck_loop`, `blind_editing`, and `over_delegation` patterns on its non-SSE API-fallback path; the fallback chain only checked 3 of 6 possible count fields. Extended to check all of them.
- Fixes #132. Alerts.tsx's Unsubscribe button left a stale, unsaved webhook-URL edit visible in the input even though the adjacent status Pill correctly flipped to "Not configured." The input now resets to server truth immediately on that path.

## Test plan

- [x] Full suite: `npm run build && npm test && npm run lint` — clean (3952/3955 passing here)
- [x] New/updated tests: `Today.test.tsx`, `Alerts.test.tsx` (new file), `Sessions.test.tsx`, `api-handler.test.ts`
- [x] Manual browser verification against a running local dashboard for #128, #129, #130, #132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.30] - 2026-07-12
+
+### Fixed
+
+- The "Alert Thresholds" card claimed these five values feed "the local alert engine" and require a server restart. Neither is true — the real (and only) consumer is the separate `deploy-alerts` CLI, which templates cloud alert conditions and reads config fresh on every invocation. Subtitle and save-confirmation copy corrected.
+- Clicking "Unsubscribe" on the Slack digest webhook left an unsaved, stale edit visible in the input box even though the adjacent status Pill correctly flipped to "Not configured." The input now resets to server truth immediately.
+- The session detail page's "Outcome" card always showed one of exactly two hardcoded values (`completed`/`in progress`) — no real outcome classification exists. Relabeled as "Status" to accurately reflect it as a running-state indicator, not a quality/outcome classification.
+- The session detail page's "Session Quality" and "Tool Selection" cards were fully built but never rendered: the backing tracker data was never attached to any of the three `/api/sessions/:id` response branches. Now attached to the branches where the data is actually available (the live session and, for quality, persisted sessions via their own stored timeline).
+- The Today view's anti-pattern banner rendered a literal `?` instead of a real count for `stuck_loop`, `blind_editing`, and `over_delegation` patterns when reached via its non-SSE API-fallback path — the fallback chain only checked 3 of the 6 possible count fields. Extended to check all of them.
+
 ## [1.4.29] - 2026-07-12
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.29",
+  "version": "1.4.30",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.29",
+      "version": "1.4.30",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.29",
+  "version": "1.4.30",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/dashboard/routes/api-handler.test.ts
+++ b/src/dashboard/routes/api-handler.test.ts
@@ -310,6 +310,183 @@ describe('api-handler GET /api/sessions/:id', () => {
     await handler(req, res);
     expect(status()).toBe(404);
   });
+
+  it('attaches qualityProxy (derived from timeline) to a persisted session with real signals', async () => {
+    const fakeSession = {
+      sessionId: 'sess-quality-1',
+      testRunCount: 4,
+      testPassCount: 3,
+      timeline: [
+        { timestamp: 1, toolName: 'Edit', durationMs: 10, success: true, filePath: 'a.ts' },
+        { timestamp: 2, toolName: 'Edit', durationMs: 10, success: false, filePath: 'b.ts' },
+      ],
+    };
+    const handler = createApiHandler({
+      sessionStore: {
+        loadTodaySessions: () => [],
+        listSessions: () => [],
+        loadSession: (id: string) => (id === 'sess-quality-1' ? fakeSession : null),
+      } as unknown as Parameters<typeof createApiHandler>[0]['sessionStore'],
+    });
+    const req = { method: 'GET', url: '/api/sessions/sess-quality-1' } as IncomingMessage;
+    const { res, status, body } = fakeRes();
+    await handler(req, res);
+    expect(status()).toBe(200);
+    const parsed = JSON.parse(body()) as { qualityProxy?: { diffApplyRate: number | null } };
+    expect(parsed.qualityProxy).toBeDefined();
+    expect(parsed.qualityProxy?.diffApplyRate).toBeCloseTo(0.5);
+  });
+
+  it('does not attach qualityProxy to a persisted session with zero signals (regression guard)', async () => {
+    const fakeSession = {
+      sessionId: 'sess-abc-999',
+      startTime: Date.now() - 5000,
+      toolCallCount: 10,
+    };
+    const handler = createApiHandler({
+      sessionStore: {
+        loadTodaySessions: () => [],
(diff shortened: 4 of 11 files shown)


corinnekunze:

PR #184, opened 2026-07-13
Fix web-dashboard cosmetic cleanup (v1.4.32)

### LINKED ISSUE #136: Two fully-built, fully-tested dashboard components have zero real consumers
## Description

Two shared UI components in `src/web/components/` are complete, professionally implemented, and covered by their own test suites — but are never imported by any of the app's 7 views.

## Findings

1. **`StatusIndicator`** (`good`/`warn`/`bad`/`neutral` widget, full test suite in `StatusIndicator.test.tsx`) — repo-wide grep for `StatusIndicator` outside its own definition/test file returns nothing.
2. **`Sparkline`** (min/max-scaled SVG trend line with hover tooltip and an accessibility label, `Sparkline.test.tsx`) — same result, zero import sites.
3. (Minor, same neighborhood) **`HourlyCostBlocks.describeChart()`'s zero-activity branch is unreachable dead code** — the component's own early-return (`maxCost === 0` → `null`) already exits before `describeChart` is ever called with that same zero value, so the branch inside it can never execute. No user-visible impact.

## Impact

Neither #1 nor #2 produces a wrong on-screen claim (nothing references them, so nothing can contradict them), but they represent real, uncounted maintenance surface — test suites that keep passing and drawing CI attention while never protecting anything a user sees. They look like they were built for a feature that either shipped differently than planned, or hasn't shipped yet.

## Fix

Product decision: wire `StatusIndicator`/`Sparkline` into a view where they'd add value, or remove both components along with their tests. For #3, either delete the unreachable branch or restructure so it's reachable if that's ever desired.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11). This is a new failure pattern not seen in the sibling MCP-tools audit (which only found dead *write* paths, never a dead *component*).

### LINKED ISSUE #137: ActivityHeatmap grid: UTC-string/local-accessor mismatch shifts weekday rows by a day
## Description

The History view's "Activity · Last 12 Weeks" grid places every cell's weekday row (and can shift month-label boundaries) one day earlier than intended, for any user in a timezone west of UTC.

## Root cause

`GridHeatmap` (`src/web/components/ActivityHeatmap.tsx:164-165,189`) constructs `new Date(days[0].date)` from a bare `YYYY-MM-DD` string sent by the backend (`src/dashboard/routes/api-handler.ts:1282,1293`, built via `dayStart.toISOString().slice(0, 10)`). Per the JS spec, a bare date-only string parses as **UTC midnight** — but `.getDay()` and `.toLocaleString(..., {month: 'short'})` both read **local** time. In any timezone west of UTC, local time at UTC midnight is still the previous calendar day, so the weekday computed is one day early.

Confirmed with a concrete example: `new Date('2026-07-08').getUTCDay()` → 3 (Wednesday, correct); `new Date('2026-07-08').getDay()` → 2 (Tuesday, wrong) in `America/New_York`.

Notably, the sibling `src/web/lib/bucket.ts`'s `buildDailyGrid()` function, audited in the same pass, correctly uses local-time accessors throughout specifically to avoid this exact bug — the defect is isolated to `ActivityHeatmap`'s grid variant, which didn't follow the codebase's own established pattern.

## Fix

Use `.getUTCDay()` and UTC-based month formatting in `GridHeatmap`, or normalize the date string to local midnight before constructing the `Date`.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11), independently re-verified with a worked example during adversarial re-review.

chrisdhaan:
## Summary

Web-dashboard cosmetic cleanup, plus a version bump to 1.4.32.

- Fixes #136. Two fully-built, fully-tested dashboard components (`StatusIndicator`, `Sparkline`) had zero real consumers anywhere in the app, confirmed via direct grep both before and after removal. Deleted both along with their test suites. Also removed an unreachable dead-code branch in `HourlyCostBlocks`' internal `describeChart()` helper — provably unreachable given its only caller's own guard logic.
- Fixes #137. The History view's activity heatmap grid parsed UTC-anchored `YYYY-MM-DD` day-key strings from the backend with `new Date(dateString)` (UTC midnight per spec) but read them back with local-time accessors, shifting every weekday row (and near month boundaries, the month label) one day early for any user west of UTC. Fixed by switching to UTC accessors to match how the string was parsed, with the rationale documented at the fix site itself (not just in the test).
- Also adds `.codegraph/.gitignore` so a local CodeGraph index directory stops showing as untracked in `git status`.

## Test plan

- [x] Full suite: `npm run build && npm test && npm run lint` — clean (3954/3957 passing here; lint fully clean, no issues at all)
- [x] New tests: `HourlyCostBlocks.test.tsx` and `ActivityHeatmap.test.tsx` (both files had zero prior coverage). The timezone test explicitly forces `process.env.TZ` since the bug is invisible in a UTC-default CI environment.
- [x] Manual browser verification against a running local dashboard, including reading the actual rendered SVG cell position for a known date to independently confirm the timezone fix beyond a visual check


corinnekunze:

PR #185, opened 2026-07-13
v1.4.33: Web dashboard API/behavior fixes + diagnostics platform-param feature

### LINKED ISSUE #139: Audit.tsx: "Export JSONL" should respect the active filter and row cap
## Description

The "Export JSONL" button on the security audit-trail viewer always exports the full unfiltered dataset (`data ?? []`, `Audit.tsx:54,67`), ignoring both the active classification filter (`visible`) and the 200-row display cap (`visibleSlice`) — even though it's placed directly next to the filter chips.

## Decision (2026-07-11)

Discussed as part of a `src/web/` dashboard audit follow-up: **export should respect the active filter**, matching what a user standing next to filter chips would reasonably expect ("export what I'm looking at").

## Fix

Change the export call to use `visible` (the filtered array) instead of the raw unfiltered `rows`. Decide separately whether the 200-row display cap should also bound the export, or whether "export everything matching the filter" (unbounded) is the intended behavior — the audit only surfaced the filter-scope question, not the cap question.

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11).

### LINKED ISSUE #140: Sessions.tsx: decide whether filesRead/antiPatterns fields should be built into the UI or removed
## Description

`SessionDetail` declares `filesRead`/`antiPatterns` fields (`Sessions.tsx:75,77`), and both are genuinely populated server-side across all three `/api/sessions/:id` response branches — but neither is ever read in this file's JSX or passed to any child component. Not a false claim (nothing on-screen claims to show either), just latent dead code, most likely left over from an earlier design.

- `filesRead`: populated at `session-store.ts:359`, never rendered anywhere.
- `antiPatterns`: populated in all three response branches (`session-store.ts:376`, `api-handler.ts:1650-1661`, and an honest empty `[]` for the registry-synthesized branch), never read — the anti-pattern banner actually shown in the Replay section comes from the separate `/api/sessions/:id/replay` endpoint instead.

## Decision needed

Discussed as part of a `src/web/` dashboard audit follow-up (2026-07-11): decision deferred. **For now, keep the dead code as-is** — no change until this issue is picked up and a call is made between:
1. Build UI to show them (files read list, inline anti-pattern data on the session detail page), or
2. Delete the dead fields (cheaper, but discards data the backend already computes for free).

## Scope note

Found during a `src/web/` dashboard audit (2026-07-11).

### LINKED ISSUE #157: Dashboard /api/diagnostics never forwards a platform param, so non-Claude-Code users get a false hooks failure
## Summary

The dashboard's `GET /api/diagnostics` endpoint (`src/dashboard/routes/api-handler.ts:1342-1352`) calls `runDiagnostics()` without ever forwarding a `platform` query param. The CLI equivalent, `preflight doctor --platform <name>`, correctly threads a `--platform` option through and gets the right per-platform behavior from `checkHooksWired` (a `warn` with manual-verification guidance for a recognized non-Claude-Code platform, rather than a hard `fail`).

Dashboard users on Kiro, Cursor, Windsurf, or any of the other 8 registered platforms therefore always see the Claude-Code-specific hooks check, which will `fail` with "Settings file not found" (since they don't use `~/.claude/settings.json`) instead of the accurate `warn` the CLI already gives them for the same check.

## Decision

Discussed during the `src/install/` subsystem audit (2026-07-11) — **decided to build this** rather than accept the gap, since the CLI already supports the correct behavior and dashboard users on non-Claude-Code platforms are currently getting a confusing false-failure.

## Suggested fix

Add a platform selector (or auto-detect from session data, if available) to the dashboard's diagnostics view, and thread the selected value through as the `platform` query param on the `/api/diagnostics` request — mirroring the CLI's existing `--platform` option all the way down to the same `runDiagnostics()` call.

## Scope note

chrisdhaan:
## Summary

Three independent fixes:

- Audit view's "Export JSONL" button exported every entry regardless of the active classification filter or the 200-row display cap. Now exports exactly the filtered+capped set shown on screen.
- Sessions view never rendered the `filesRead`/`antiPatterns` fields the backend already computes for each session. Now shows a "Files Read" list (mirroring the existing "Files Modified" pattern) and "Anti-Patterns" pills (reusing the existing `SEGMENT_LABELS` taxonomy/`Pill` styling already used elsewhere in the same file).
- Dashboard's diagnostics panel always assumed Claude Code, so non-Claude-Code users (Cursor, Windsurf, etc.) got a false "Hooks wired: fail". `/api/diagnostics` now forwards the already-detected platform into `runDiagnostics()`, which already knew how to handle it correctly — it just never received the value. Backward compatible: Claude Code's `platformName` is literally `'claude-code'`, the exact string the existing check special-cases, so real Claude Code users see zero behavior change.

Fixes #139. Fixes #140. Fixes #157.

Version bumped to 1.4.33.

## Test plan

- [x] `npm run build && npm test && npm run lint` — build clean, tests 3956/3959 passing
- [x] Manually verified live against an isolated instance: Anti-Patterns pills + Files Read list render correctly, and `/api/diagnostics` correctly reports "not applicable to \"generic-mcp\"" instead of a false hooks-wired failure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.33] - 2026-07-13
+
+### Fixed
+
+- Audit view's "Export JSONL" button exported every audit entry regardless of the active classification filter or the 200-row display cap. Now exports exactly what's shown on screen.
+- Sessions view never rendered the `filesRead` or `antiPatterns` data the backend already computes for each session. Now shows a "Files Read" list and "Anti-Patterns" pills (reusing the same taxonomy labels used elsewhere in the session replay view) when present.
+- Dashboard's System Health diagnostics panel always ran the Claude-Code-specific hooks-wired check, regardless of which coding platform was actually detected — non-Claude-Code users (Cursor, Windsurf, etc.) saw a false "Hooks wired: fail" result. Now forwards the detected platform, which the existing check already knew how to handle correctly.
+
 ## [1.4.32] - 2026-07-13
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.32",
+  "version": "1.4.33",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.32",
+      "version": "1.4.33",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.32",
+  "version": "1.4.33",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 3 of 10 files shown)


corinnekunze:

Issue #186, opened 2026-07-13
useLiveEvents.ts: /api/anti-patterns REST hydration pushes target/count as undefined into the live store


chrisdhaan:
## Description

`src/web/hooks/useLiveEvents.ts`'s `hydrateFromApi()` fetches `/api/anti-patterns` directly (bypassing `src/web/api/client.ts`) and pushes each result into the live store via `store.pushAntiPattern(ap as unknown as { type: string; target: string; count: number })`. The live store's `pushAntiPattern` expects an `AntiPatternEvent` (`src/web/store/liveStore.ts`) with required `target: string` and `count: number` fields.

The real server-side `AntiPattern` type (returned by `/api/anti-patterns`, see `src/metrics/anti-patterns.ts`) has no `target` or `count` fields at all — its fields are `type`, `file?`, `command?`, `bashCategory?`, `iterations?`, `readCount?`, `repeatCount?`, `editCount?`, `agentCount?`, `suggestion`.

## Impact

Every anti-pattern object hydrated via this REST path — i.e. on initial page load, before the first SSE anti-pattern event arrives — gets `target: undefined` and `count: undefined` pushed into the live store, despite the store's type declaring both as required. Any UI reading `target`/`count` off a REST-hydrated entry would render `undefined` until the first real SSE anti-pattern event replaces it.

## Why this was invisible until now

Before v1.4.34 (part of the #141 typed-response-contracts series), the fetch's result was cast from `Promise<unknown>`, which permits an unchecked single-assertion cast to any shape. Typing it properly as `Promise<AntiPattern[]>` (the real, accurate shape) made the cast to `{ type; target; count }` fail TypeScript's structural-overlap check, forcing a `as unknown as` double-cast to keep compiling — which is the correct, behavior-preserving fix for that release (a type-only release, no runtime changes), but it makes this pre-existing mismatch visible for the first time.

## Suggested fix

Either:
- Change `AntiPatternEvent`/`pushAntiPattern`'s expected shape to match the real `AntiPattern` type (if the REST hydration path is meant to feed the same store as SSE), or
- Map the real `AntiPattern` fields to whatever `target`/`count` are supposed to represent (e.g. `target` from `file`/`command`, `count` from `iterations`/`readCount`/etc., mirroring the fallback chain already used in `Today.tsx` for the same fields), or
- Determine whether this hydration path is dead/unreachable in practice and remove it if so.

[...]

Issue #141, opened 2026-07-11
src/web/api/client.ts: consider typed response contracts instead of Promise<unknown>


chrisdhaan:
## Description

Nearly every export in `src/web/api/client.ts` (~25 of them) is typed `Promise<unknown>` — there is no compile-time contract for what shape a route actually returns, only `fetchHealth()` has a real type.

## Why this matters

This is the likely root-cause enabler behind a bug pattern that hit **3 different views** in a recent `src/web/` dashboard audit: a backend field is genuinely computed and exposed elsewhere, but a specific view's response object never attaches it, and nothing catches the mismatch at compile time because every consumer reads from an `unknown`-typed response. Confirmed instances: Sessions.tsx's Quality/Tool-Selection cards (#130), Audit.tsx's dropped severity (#133), and the dead `filesRead`/`antiPatterns` fields (#140).

## Decision (2026-07-11)

Discussed as part of that audit's follow-up: real value, but adding ~25 typed response contracts is its own scoped effort — **track as a future improvement, not bundled into the cleanup fixes from that audit.**

## Scope

Define response types for each route in `src/dashboard/routes/api-handler.ts` and thread them through the corresponding `client.ts` export, so a future field-mismatch is a compile error instead of a silent `undefined`.

PR #187, opened 2026-07-13
v1.4.34: client.ts typed response contracts — Today.tsx cluster

chrisdhaan:
## Summary

Progress on #141: gives `src/web/api/client.ts`'s dashboard API functions real TypeScript response interfaces instead of `Promise<unknown>`, removing the matching `as Promise<T>` casts at call sites. This release covers the 15 functions consumed by `src/web/views/Today.tsx`, plus reconciling two duplicate/undertyped response shapes in `src/web/hooks/useLiveEvents.ts`. Part of an ongoing effort to type the rest of the dashboard's API layer — #141 stays open until the remaining views are covered.

- All 15 interfaces were derived by reading the real route handlers in `src/dashboard/routes/api-handler.ts` and the underlying tracker classes (`src/metrics/*.ts`), not guessed.
- Every interface is declared locally in `client.ts` (matching the existing `HealthResponse` precedent), since `src/web` builds under a separate `tsconfig.web.json` that excludes server-side source.
- View-local interfaces that were exact duplicates of the new canonical types were deleted in favor of importing the canonical type; narrower view-local projections were left in place.
- `fetchActivityHeatmap` became an overloaded function (`'today'`/`'history'` literal args each resolve to their own concrete response type) since it returns one of two shapes depending on its argument.

**Also discovered, filed as #186, not fixed here (behavior change, out of scope for a type-only release):** `useLiveEvents.ts`'s `/api/anti-patterns` REST hydration path pushes `target: undefined`/`count: undefined` into the live store, since the real server `AntiPattern` type has no such fields. Invisible before this release because the code cast from `unknown`; now visible as a forced `as unknown as` double-cast.

**Tooling note for future `src/web/` work:** `npm run build` (Vite/esbuild, no type-checking) and `npm run lint` (no `parserOptions.project`) don't catch TypeScript errors in `src/web/**`. Use `npx tsc --noEmit -p tsconfig.web.json` to actually type-check that directory.

## Test plan

- [x] `npm run build && npm test && npm run lint` clean
- [x] `npx tsc --noEmit -p tsconfig.web.json` — zero new errors vs. base
- [x] Zero remaining `as Promise<` casts in `Today.tsx`; all 15 target functions no longer `Promise<unknown>`
- [x] `scripts/verify-public-diff.sh` passed clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.34] - 2026-07-13
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


corinnekunze:

PR #190, opened 2026-07-14
v1.4.36: client.ts typed response contracts — Settings/Alerts/ContextBar/Audit cluster

### LINKED ISSUE #141: src/web/api/client.ts: consider typed response contracts instead of Promise<unknown>

chrisdhaan:
## Summary

Part 3 of 3 for #141. Types the last 7 `Promise<unknown>` exports in `src/web/api/client.ts`, consumed by the Settings, Alerts, ContextBar, and Audit views:

- `fetchAuditLog` → `AuditEntry` (also fixes a real type-accuracy bug: `sessionId` is `string | null` at runtime, not `string | undefined` as previously declared — render-invisible, since the one render site already uses `??`)
- `fetchContext` → `ContextResponse` (includes `history` for full accuracy, even though unused today)
- `fetchSettings` → new shared `SettingsResponse` superset type (consumed by both Settings.tsx and Alerts.tsx, each keeping its own narrower local `SettingsData` via structural assignability — no cast needed)
- `fetchDiagnostics` → `DiagnosticCheck[]` (moved from a byte-identical local interface)
- `patchSettings` → `PatchSettingsResponse`
- `postDigestSend` → `DigestSendResponse`

Also deletes the confirmed-dead `fetchSessionToday` export (zero consumers anywhere in `src/web`, including tests) and its now-unused `qk.sessionToday` query key — the one other `Promise<unknown>` left in the file besides these 7.

**`grep -n "Promise<unknown>" src/web/api/client.ts` now returns nothing** — every exported function in `client.ts` has a real response interface. Fixes #141.

Every new interface is redeclared locally in `client.ts` (never imported from server source — `tsconfig.web.json` excludes it entirely), mirroring the real backend shape cited in each interface's own comment.

No behavior changes anywhere in this PR.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.web.json` — zero new errors vs. baseline
- [x] `npm run build` — clean
- [x] `npm test` — 129/130 suites, 3956/3959 tests passing
- [x] `npm run lint` — only the pre-existing gap in `scripts/*.ts` (1 error + 5 warnings), nothing new
- [x] Every new interface independently re-verified against its real backend source by a final whole-branch review, not just per-task reports

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.36] - 2026-07-13
+
+- Typed the last 7 `client.ts` response contracts, consumed by the Settings, Alerts, ContextBar, and Audit views (`fetchAuditLog`, `fetchContext`, `fetchSettings`, `fetchDiagnostics`, `patchSettings`, `postDigestSend`), removing the `as Promise<T>` casts these views previously needed. Also removed the unused `fetchSessionToday` export, which had zero consumers. This closes the effort to type the rest of the dashboard's API layer (#141) — every exported function in `client.ts` now has a real response interface instead of `Promise<unknown>`.
+
 ## [1.4.35] - 2026-07-13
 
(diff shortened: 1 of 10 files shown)


corinnekunze:

PR #192, opened 2026-07-14
Fix: undici missing as a direct dependency, breaking fresh installs

### LINKED ISSUE #189: v1.4.34: registry manifest missing runtime deps (undici, hono, js-yaml) — fresh install crashes at startup
**Summary:** A fresh `npm install -g @newrelic/preflight@1.4.34` produces a broken installation: the CLI crashes at startup with `ERR_MODULE_NOT_FOUND` for `undici`.

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'undici' imported from /usr/lib/node_modules/@newrelic/preflight/dist/proxy/otlp-receiver.js
```

**Root cause:** The npm registry manifest (packument) for 1.4.34 does not match the `package.json` inside the published tarball. Three runtime dependencies are missing from the registry metadata, so npm never installs them:

- `undici@^7.28.0`
- `hono@^4.12.25`
- `js-yaml@^4.2.0`

Compare:

```
$ npm view @newrelic/preflight@1.4.34 dependencies
# → lists @modelcontextprotocol/sdk, @opentelemetry/*, commander, dotenv, zod
# → undici / hono / js-yaml are absent

$ grep -B2 -A1 undici /usr/lib/node_modules/@newrelic/preflight/package.json
    "hono": "^4.12.25",
    "js-yaml": "^4.2.0",
    "undici": "^7.28.0"
  },
  "devDependencies": {
# → present in the tarball's "dependencies"
```

npm resolves the dependency tree from the packument, not the tarball, so every fresh install of 1.4.34 is missing these packages and crashes on startup.

**Repro:**

```
npm install -g @newrelic/preflight@1.4.34
preflight --version   # → ERR_MODULE_NOT_FOUND: undici
```

**Workaround (until a fixed version is published):**

```
cd $(npm root -g)/@newrelic/preflight
npm install --ignore-scripts --no-save 'undici@^7.28.0' 'hono@^4.12.25' 'js-yaml@^4.2.0'
```

**Suggested fix:** Publish a new patch release so the registry manifest matches the tarball's `package.json`. (This kind of mismatch can happen when the packed manifest is mutated during publish — worth verifying the publish pipeline, since re-publishing the same content could reproduce it.)

chrisdhaan:
## Summary
- `src/proxy/otlp-receiver.ts` imports `undici` directly, but it was only declared as a version `overrides` pin (added for security hardening), never as an actual dependency.
- A clean `npm install`/`npm install -g` skips devDependencies, so `undici` was never installed and the proxy crashed with `ERR_MODULE_NOT_FOUND: undici` on startup. It only "worked" in CI because `jsdom` (a devDependency) pulls `undici` in transitively.
- Adds `undici` to `dependencies`, bumps to 1.4.38, updates CHANGELOG.

Fixes #189

## Test plan
- [x] `npm run build`
- [x] `npm test` (passing)
- [x] `npm run lint` (pre-existing unrelated issues only)
- [x] Confirmed against the live `registry.npmjs.org` packument and the actual published 1.4.34/1.4.37 tarballs that `undici` is absent from `dependencies` in both

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.38] - 2026-07-14
+
+### Fixed
+
+- `undici` is now a direct runtime dependency. `src/proxy/otlp-receiver.ts` imports it directly, but it was declared only as a version `overrides` pin (added for security hardening), never as an actual dependency — so a clean `npm install`/`npm install -g` (no devDependencies) never installed it, and the proxy crashed with `ERR_MODULE_NOT_FOUND: undici` on startup. It previously worked in this repo's own dev/CI environment only because `jsdom` (a devDependency) pulls in `undici` transitively.
+
 ## [1.4.37] - 2026-07-13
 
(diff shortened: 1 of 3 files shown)


corinnekunze:

PR #194, opened 2026-07-14
v1.4.39: Proxy defensive-coding fixes

### LINKED ISSUE #148: Proxy subsystem: minor defensive-coding gaps found during audit
## Description

Three small, low-priority gaps found during a `src/proxy/` subsystem audit (2026-07-11). Each has a clear, no-downside resolution — bundled here rather than filed as separate issues given their size.

## Items

1. **Dangerous env-var denylist doesn't cover non-Node interpreters.** `DANGEROUS_ENV_KEYS` (`src/proxy/upstream-stdio.ts:28-34`) matches `SECURITY.md` exactly (`LD_PRELOAD`, `LD_LIBRARY_PATH`, `DYLD_INSERT_LIBRARIES`, `DYLD_LIBRARY_PATH`, `NODE_OPTIONS`) but doesn't cover equivalent injection vectors for other interpreters (`PYTHONPATH`/`PYTHONSTARTUP`, `RUBYOPT`, `PERL5LIB`, `LD_AUDIT`, `GCONV_PATH`, `BASH_ENV`/`ENV`, `ELECTRON_RUN_AS_NODE`) that an operator-configured stdio upstream could be pointed at (since `UpstreamConfig.command` accepts any absolute-path executable, not just Node). Not attacker-reachable under the current threat model (this is operator-controlled config, not a per-request path) — decision made 2026-07-11: expand the list anyway, since stripping additional env vars for a child process that's supposed to be a clean MCP server has no functional downside.
2. **Dispatch timeout abandons but doesn't cancel the underlying child-process call.** `dispatchToClient()` (`upstream-stdio.ts:285-333`) correctly returns an error to the HTTP client within the timeout window, but the original request to the child process keeps running in the background with no `AbortController` anywhere in the file — no code left awaiting it, and the eventual response is silently dropped. Under repeated timeouts, a wedged child accumulates abandoned in-flight requests with no backpressure or cancellation. Decision made 2026-07-11: needs a real fix (likely requires SDK-level abort-signal support, which the current call path doesn't accept) — not a product decision, just real engineering work.
3. **OTLP receiver's `injectResourceAttributes` never de-duplicates.** If a source payload already carries one of the enrichment keys (`ai.session.id`, `ai.developer`, `ai.project_id`, `ai.team_id`), the injection loop (`src/proxy/otlp-receiver.ts:333-341`) unconditionally pushes a new entry, producing two entries for the same key with different values — undefined/last-wins behavior depending on the downstream OTLP consumer. Decision made 2026-07-11: add a check-before-push guard that skips the injection if the key already exists, so Preflight's enrichment never clobbers a value the instrumented app deliberately set itself.

## Fix

Each item is a small, self-contained change — pick up opportunistically alongside other work in the same files.

chrisdhaan:
## Summary

Three independent defensive-coding fixes in `src/proxy/`:

- **Env-var denylist**: `DANGEROUS_ENV_KEYS` in `src/proxy/upstream-stdio.ts` now covers non-Node interpreter injection vectors (`PYTHONPATH`, `PYTHONSTARTUP`, `RUBYOPT`, `PERL5LIB`, `LD_AUDIT`, `GCONV_PATH`, `BASH_ENV`, `ENV`, `ELECTRON_RUN_AS_NODE`) in addition to the existing Node/dynamic-linker keys, since `UpstreamConfig.command` accepts any absolute-path executable.
- **Dispatch timeout cancellation**: `dispatchToClient()` now creates an `AbortController` and passes `{signal}` to every MCP SDK call it dispatches, aborting it in the existing timeout callback — a timed-out request now actually stops instead of continuing to run against the child process with nothing awaiting it.
- **OTLP enrichment de-duplication**: `injectResourceAttributes()` in `src/proxy/otlp-receiver.ts` now skips injecting an enrichment key the payload already carries, so Preflight's enrichment never clobbers a value the instrumented app deliberately set itself.

Fixes #148

## Test plan

- [x] `npm run build && npm test && npm run lint` — 4005/4008 tests passing
- [x] New regression test for the dispatch-timeout abort (asserts the underlying SDK call's `options.signal.aborted`)
- [x] New regression test for the OTLP dedup guard (asserts exactly one surviving entry with the original value)
- [x] Existing parametrized test auto-covers all 14 denylist keys

Co-authored-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.39] - 2026-07-14
+
+### Fixed
+
+- Proxy stdio upstream: expanded the dangerous env-var denylist to cover non-Node interpreters (`PYTHONPATH`, `RUBYOPT`, `PERL5LIB`, and others) that an operator-configured stdio upstream command could be pointed at.
+- Proxy stdio upstream: a dispatch timeout now aborts the underlying MCP client call instead of only racing past it, so a wedged child process no longer accumulates abandoned in-flight requests.
+- Proxy OTLP receiver: resource-attribute enrichment no longer duplicates a key the instrumented application already set on the payload.
+
 ## [1.4.38] - 2026-07-14
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.38",
+  "version": "1.4.39",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.38",
+      "version": "1.4.39",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.38",
+  "version": "1.4.39",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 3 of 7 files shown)


corinnekunze:

PR #195, opened 2026-07-14
v1.4.40: Web dashboard UX/defensive-coding fixes

### LINKED ISSUE #138: Web dashboard: minor UX/defensive-coding gaps
## Description

A collection of small, low-priority gaps found during a `src/web/` dashboard audit (2026-07-11). None involve a false on-screen claim or a decision to make — each is an obviously-worth-doing, no-ambiguity fix, bundled here rather than filed as separate issues given their size.

## Items

1. **Today.tsx empty-state race** — `noActivityToday`'s pending-gate (`Today.tsx:274-281`) omits `liveSessions`/`concurrency`/`todayHeatmap`, so a session can register live while the empty state still shows for roughly one poll cycle, hiding `LiveSessionPane`. Self-correcting, not permanent.
2. **Missing `isError` handling** — `History.tsx`'s `weekly`/`sessions`/`costPerOutcome`/`concurrencyHistory` queries and `Alerts.tsx`'s budget query all silently degrade to an empty/default render on a 503, indistinguishable from "no data yet."
3. **GitEfficiency.tsx hardcoded `origin/main`** — `src/index.ts:864,869-870` hardcode `'main'` and always diff against `origin/main` instead of reading the real `repoContext.defaultBranch`/`.remoteName` fields, which exist but are never used. Not yet user-facing (nothing renders these fields today) but a latent correctness risk for any repo whose default branch isn't `main`.
4. **`liveStore` dedupe-key format mismatch** — a code comment (`liveStore.ts:136`) claims hydrate-path and SSE-path tool-call pushes "don't push the same event twice," but the two paths construct `id` differently (`${timestamp}-${toolName}` vs. a UUID) with no guaranteed collision. Unproven by any test, no observed failure in practice.
5. **`Kpi`'s `tone="accent"` renders identically to `tone="good"`** (`Kpi.tsx:6,10`) — likely a vestigial/redundant enum value; only one call site uses it.
6. **`formatDuration` has no `NaN`/negative guard** (`src/web/lib/format.ts`) — unlike `formatNumber` in the same file, which explicitly em-dashes non-finite input. No in-scope caller passes such values today.
7. **Alerts.tsx Slack Digest card has no independent loading guard** — while settings are loading, this card renders only its header with no spinner (the Alert Thresholds card above it does show one, but that guard doesn't cover this card).

8. **Audit.tsx classification Pill shows the raw key, not a friendly label** (`Audit.tsx:126-130`) — renders the literal string e.g. `sensitive_file` instead of mapping it through the existing `FILTERS` list to show "Sensitive files". Cosmetic only, value shown is accurate.

## Fix

Each item is a small, self-contained change — pick up opportunistically alongside other work in the same files.

chrisdhaan:
## Summary

Fixes #138 — 8 small UX/defensive-coding gaps in the web dashboard:

- Today.tsx's empty-state gate omitted `concurrency`/`todayHeatmap`/`liveSessions` pending state, letting the empty state show briefly even as a live session registered.
- History.tsx and Alerts.tsx now surface query load errors instead of silently rendering as "no data yet."
- Git Efficiency's branch-divergence and repo-context now derive the real default branch instead of assuming `main`.
- `formatDuration` now guards against non-finite/negative input, matching `formatNumber`'s existing guard.
- Alerts.tsx's Slack Digest card now shows a loading indicator while settings load, matching the cards above it.
- Audit.tsx's classification Pill now shows the friendly label instead of the raw internal key.
- Removed `Kpi`'s redundant `tone="accent"`, which rendered identically to `tone="good"`.
- Corrected a stale comment in `liveStore.ts` overstating cross-path dedup guarantees for live tool-call events.

## Test plan

- [x] `npm install && npm run build` — clean
- [x] `npm test` (Jest) — 4010 passed, 0 failed
- [x] `npm run test:web` (Vitest) — 278 passed, 1 pre-existing failure (`AlertBanner.test.tsx`, confirmed unrelated to this change and present on `main` before this PR)

🤖 Generated with subagent-driven-development

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.40] - 2026-07-14
+
+### Fixed
+
+- Today.tsx's empty-state gate omitted concurrency/activity-heatmap/live-session pending state, letting the empty state show briefly even as a live session registered.
+- History.tsx and Alerts.tsx now surface query load errors instead of silently rendering as "no data yet."
+- Git Efficiency's branch-divergence and repo-context now derive the real default branch instead of assuming `main`.
+- `formatDuration` now guards against non-finite/negative input, matching `formatNumber`'s existing guard.
+- Alerts.tsx's Slack Digest card now shows a loading indicator while settings load, matching the cards above it.
+- Audit.tsx's classification Pill now shows the friendly label instead of the raw internal key.
+- Removed `Kpi`'s redundant `tone="accent"`, which rendered identically to `tone="good"`.
+- Corrected a stale comment in `liveStore.ts` overstating cross-path dedup guarantees for live tool-call events.
+
 ## [1.4.39] - 2026-07-14
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.39",
+  "version": "1.4.40",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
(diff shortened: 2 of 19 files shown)


corinnekunze:

PR #207, opened 2026-07-15
v1.4.43: Fix Windows path handling (WSL transcript + native PATH delimiter)

### LINKED ISSUE #193: wsl-windows-cc: model/token/cost collection silently fails — WSL collector can't read the Windows transcript_path
**Summary:** In the officially supported `wsl-windows-cc` setup (Windows Claude Code desktop + preflight installed inside WSL, hooks generated by `preflight install --windows-cc`), tool events arrive fine, but **token/model/cost collection never produces a single event**. The dashboard permanently shows `unknown` with no Eff. / Success / Avg $ under MODEL PERFORMANCE, and Cost per Outcome never populates. No error is surfaced anywhere, so the integration *looks* healthy.

**Root cause:** `collectTranscriptTokens()` in `dist/hooks/collector-script.js` prefers the hook payload's `transcript_path` verbatim. Windows Claude Code sends a Windows-format path (`C:\Users\<user>\.claude\projects\...\<session>.jsonl`), but the collector runs inside WSL, where Node's `statSync()` on that path throws ENOENT. The catch block silently ignores it ("transcript may not exist yet"), so every invocation bails before reading any usage data. The same file is perfectly readable at `/mnt/c/Users/<user>/...`.

The `getTranscriptPath(cwd, sessionId)` fallback is doubly broken for this platform:

- `cwd.replace(/\//g, '-')` is a no-op on a backslash-separated Windows `cwd`, so the derived project dir is garbage
- `homedir()` resolves to the **Linux** home (`/home/<user>/.claude`), not the Windows `%USERPROFILE%\.claude` where the transcripts actually live

Verified on 1.4.38 inside WSL:

```
$ node -e "fs.statSync('C:\Users\<user>\.claude\projects\<proj>\<session>.jsonl')"
# → ENOENT
$ ls /mnt/c/Users/<user>/.claude/projects/<proj>/<session>.jsonl
# → exists
```

**Impact:** every `wsl-windows-cc` user gets zero model/token/cost data (model shows as `unknown`, costs stay null). Tool pre/post events work normally, which masks the failure — the generated hook command (`wsl.exe -e "<path>/preflight-collector" pre-tool`) and the buffer pipeline are fine; only the transcript read is broken.

**Repro:**

1. Windows Claude Code (desktop), preflight installed in WSL via `preflight install --windows-cc`
2. Run any Claude Code session with tool calls
3. `buffer-<session>.jsonl` contains `mode:"pre"` / `mode:"post"` records but never a `mode:"token"` record; dashboard MODEL PERFORMANCE shows `unknown`

**Suggested fix:** in `collectTranscriptTokens()`, when running on Linux and the path matches `/^[A-Za-z]:[\/]/`, translate it to the WSL mount before `statSync` (`C:\foo\bar` → `/mnt/c/foo/bar`, i.e. what `wslpath` does). Applying the same translation to the `cwd`-based fallback (and pointing it at the Windows `.claude` home) would cover payloads without `transcript_path`. Doing this in the collector rather than in `generateHookEntries()` also covers hand-rolled hook commands. A once-only stderr warning on transcript-read failure (the file already has that pattern for transcript-pos write failures) would have made this diagnosable in seconds instead of requiring a source dive.

---

P.S. Since diagnosing this I've migrated to a Windows-native install (`platformTarget: native`), which sidesteps this entire class of cross-boundary path problems — so this should be my last report from a `wsl-windows-cc` environment 🙂

### LINKED ISSUE #205: Windows native: resolveBinaryPath() splits PATH on ':' — always null on win32, installer emits bare hook commands, falsely warns 'not on PATH', and platform switch never upgrades hooks
## Summary

`src/install/schedule.ts` `resolveBinaryPath()` splits `process.env.PATH` on a hard-coded `':'`:

```ts
const pathDirs = (process.env.PATH ?? '').split(':').filter(Boolean);
```

On Windows the PATH delimiter is `';'` and every entry contains a drive-letter colon (`C:\...`), so the split produces garbage fragments (`C`, `\Users\x;D`, …) and the function returns `null` on every native-Windows install.

## Impact (native Windows, non-WSL `preflight install`)

1. **False warning** — the installer always prints the "preflight is not on your PATH" warning (`printPathWarning()`), even when the npm global bin dir is on PATH and `preflight` works fine.
2. **Absolute-path protection silently lost** — `generateHookEntries()` / `generateMcpServerEntry()` fall back to bare `preflight-collector` / `preflight` commands instead of the intended absolute paths. This happens to work when Claude Code inherits a PATH containing the npm global bin dir, but breaks with per-shell version managers (nvm-windows switching, volta pinning) or when the desktop app is launched with a different environment — and it fails silently, since hook errors are swallowed.
3. It also affects anything else keying off `resolveBinaryPath()` (e.g. `verifyBinaryOnPath()`).

## Real-world consequence we hit: platform switch (WSL → native) never upgrades hooks

`mergeSettings()` has two branches: with a resolved `binPath` it removes stale NR entries and re-adds the current form ("re-install always upgrades"); with `binPath === null` it deliberately leaves any existing NR-matching entry untouched. Because `resolveBinaryPath()` is always null on win32, a native-Windows `preflight install` can never take the upgrade branch — so when migrating from a `wsl-windows-cc` install, the stale `wsl.exe -e ... preflight-collector pre-tool` hooks (which match `NR_HOOK_RE`) are silently preserved and the platform switch does not actually switch the hooks. The user is then pushed toward hand-editing settings.json (encouraged by the false "not on your PATH" warning), which is where unquoted-backslash mistakes happen.

## Verified behavior (Windows 11, Git Bash hooks, @newrelic/preflight 1.4.38)

Claude Code on native Windows executes hook commands via Git Bash, so the POSIX-escaping in `generateHookEntries()` is actually consistent — we confirmed all three generated forms run correctly under `bash -c`:

| hook command form | result |
|---|---|
| `preflight-collector pre-tool` (bare, current win32 output) | works (exit 0) |
| `"C:\\Users\\...\\preflight-collector" pre-tool` (escaped absolute, intended output) | works (exit 0) |
| unquoted `C:\Users\...\preflight-collector pre-tool` | **fails**: `bash: C:Users...: command not found` |

So the escaping design is fine; the bug is only that the absolute-path branch is unreachable on win32.

## Suggested fix

1. Split PATH with `path.delimiter` instead of `':'`.
2. On win32, also probe PATHEXT-style candidates (`preflight.cmd`, `preflight.ps1`, bare sh shim) since the npm shims carry extensions; `accessSync(X_OK)` is a no-op on Windows, so extension probing is the meaningful check.
3. Optional hardening: when embedding a win32 absolute path into a hook command, normalize to forward slashes (`p.replaceAll('\\', '/')`). Git Bash and the Win32 API both accept forward slashes, which removes the JSON→shell double-escaping hazard entirely and keeps the command readable in settings.json.

## Environment

- @newrelic/preflight 1.4.38 (same code on current `main`)
- Windows 11, native Claude Code (hooks executed via Git Bash)
- Node v22.x (native Windows npm global install)

chrisdhaan:
## Summary

- Fixed the `wsl-windows-cc` collector failing to read Windows-style transcript paths sent by desktop Claude Code, which silently dropped all token/model/cost collection for every WSL-hosted install.
- Fixed `resolveBinaryPath()` splitting `PATH` on `:` instead of the platform delimiter and never probing npm's `.cmd`/`.ps1` shim extensions, which always returned null on native Windows — causing a false "not on PATH" warning, loss of absolute hook-path resolution, and a stalled hook upgrade when switching from WSL to native Windows installs.

Fixes https://github.com/newrelic-experimental/preflight/issues/193
Fixes https://github.com/newrelic-experimental/preflight/issues/205

## Test plan

- [x] New regression tests for the WSL path translation and PATH-delimiter fixes
- [x] Full suite passing, lint clean
- [x] `package-lock.json` version synced to 1.4.43

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.43] - 2026-07-14
+
+### Fixed
+
+- Fixed the WSL-hosted collector failing to read Windows-style transcript paths sent by desktop Claude Code, which silently dropped all token/model/cost collection for `wsl-windows-cc` installs.
+- Fixed `resolveBinaryPath()` splitting `PATH` on `:` instead of the platform delimiter and never probing npm's `.cmd`/`.ps1` shim extensions, which always returned null on native Windows — causing a false "not on PATH" warning, loss of absolute hook-path resolution, and a stalled hook upgrade when switching from WSL to native Windows installs.
+
 ## [1.4.42] - 2026-07-14
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.42",
+  "version": "1.4.43",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.42",
+      "version": "1.4.43",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.42",
+  "version": "1.4.43",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/hooks/collector-script.test.ts
+++ b/src/hooks/collector-script.test.ts
@@ -13,6 +13,7 @@ import {
   collectTranscriptTokens,
   readLastAssistantUsage,
   getTranscriptPath,
+  translateWslPath,
   getBufferPath,
   writePpidBreadcrumb,
   getLinuxAncestorPids,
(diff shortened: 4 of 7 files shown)


corinnekunze:

Issue #183, opened 2026-07-13
AlertBanner.test.tsx: stale assertion asserts raw values instead of formatNumber()'s 1-decimal tier output


chrisdhaan:
## Description

`AlertBanner.test.tsx`'s `'renders title, description, value, and threshold'` test has been failing on `main` independent of any recent feature work — confirmed present, byte-identical, across three consecutive unrelated release branches (v1.4.30, v1.4.31, v1.4.32).

## Root cause

The test's `makeAlert()` fixture sets `value: 12.34, threshold: 10`, and asserts:

```ts
expect(screen.getByText(/12\.34 \/ 10/)).toBeInTheDocument();
```

This assumes `AlertBanner.tsx` echoes the raw numeric props verbatim. It doesn't — the value/threshold Pill (`AlertBanner.tsx`) renders both numbers through `formatNumber()` (`src/web/lib/format.ts:96-102`), which has a deliberate "1-decimal tier" for numbers `>= 10` (comment: "Smooth 1-decimal tier prevents jump at 100 boundary"):

```ts
export function formatNumber(n: number): string {
  if (!Number.isFinite(n)) return '—';
  if (Math.abs(n) >= 100) return n.toFixed(0);
  if (Math.abs(n) >= 10) return n.toFixed(1); // Smooth 1-decimal tier prevents jump at 100 boundary
  if (Number.isInteger(n)) return String(n);
  return n.toFixed(2);
}
```

So `formatNumber(12.34)` → `"12.3"` and `formatNumber(10)` → `"10.0"` — the component genuinely renders "12.3 / 10.0", which the test's stale regex never matches. `git blame` on this tiering logic points to `89e5ca5` ("Fix: full codebase review — ~180 bug fixes... #117", 2026-06-07) — the test's expectation was apparently never updated to match, and has silently failed in the suite ever since.

The component itself is correct; only the test literal is wrong.

## Impact

A real, currently-failing test has been sitting in the suite for over a month with nobody noticing — it doesn't block CI (the project's test command doesn't currently fail the build on this), but it means this file's actual coverage of the value/threshold rendering path has been zero for that entire time, and any real regression in that rendering would go undetected.

## Fix

[...]

PR #208, opened 2026-07-15
v1.4.44: Content-based instruction drift hashing

### LINKED ISSUE #123: nr_observe_get_instruction_drift correlates on Read call arguments, not CLAUDE.md file content
## Description

`nr_observe_get_instruction_drift` (fixed in #116 to actually receive data) is meant to correlate CLAUDE.md / system prompt *content* changes with session outcomes. It currently correlates on something narrower: the arguments of the `Read` tool call, not the content of the file being read.

## Root cause

The shared hashing utility's `Read` handling in `recordToolCall()` computes `hashInput(data.tool_input)` — a hash of the tool call's arguments (file path, offset, limit). Two reads of the same file at the same offset/limit hash identically even if the file's content changed between reads; conversely, re-reading the same unchanged file with a different offset/limit produces a different hash even though nothing about the instructions changed.

Net effect: the tool detects read/path drift (did the agent read CLAUDE.md differently), not CLAUDE.md content drift (did CLAUDE.md's actual text change) — which is what its description promises.

## Why this isn't trivial

PreToolUse hooks don't have file content in scope — only the tool call arguments. Fixing this for real likely needs a PostToolUse content capture (or similar) so the collector can hash the file's actual bytes instead of the call's arguments. This is a real design effort, not a one-line fix.

## Scope note

Found as an out-of-scope side-finding during the `instruction_drift` wiring fix (#116). The user explicitly deferred it at the time rather than blocking that fix on it, and the tool shipped with its description left aspirational (matches the eventual goal, not today's narrower implementation) — same precedent as `nr_observe_report_feedback` (#114).

chrisdhaan:
## Summary
- `nr_observe_get_instruction_drift` correlated CLAUDE.md/instruction-file drift on the arguments of a `Read` tool call (file path, offset, limit) instead of the file's actual content, producing false drift signals on re-reads at a different offset/limit and missing real content changes read at an identical offset/limit.
- Fixed by reusing the existing `ClaudeMdTracker.computeFileHash()` helper to hash the file's real on-disk content instead, wrapped in a try/catch for the unreadable-file case.
- Version bumped to 1.4.44 with a CHANGELOG entry.

Fixes #123

## Test plan
- [x] New tests using real temp files prove both the false-positive (same content, different `inputHash`) and false-negative (different content, same `inputHash`) bugs are fixed
- [x] New test covers a previously-set `promptHash` surviving a subsequent failed/unreadable read
- [x] Full suite passes in this clone (4123 passed, 3 skips, 0 failures)
- [x] Lint clean
- [x] Vitest shows only the pre-existing, unrelated `AlertBanner.test.tsx` failure (tracked as #183)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.44] - 2026-07-15
+
+### Fixed
+
+- Fixed instruction/prompt drift tracking correlating on the arguments of a `Read` tool call (file path, offset, limit) instead of the target file's actual content, which produced false drift signals when re-reading an unchanged file at a different offset or limit, and missed real content changes read at an identical offset/limit.
+
 ## [1.4.43] - 2026-07-14
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.43",
+  "version": "1.4.44",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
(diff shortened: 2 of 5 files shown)


corinnekunze:

PR #209, opened 2026-07-15
Decision-tree real reasoning extraction

### LINKED ISSUE #124: nr_observe_get_decision_tree's reasoning field is a rule-based template, not extracted model chain-of-thought
## Description

`nr_observe_get_decision_tree`'s description promises "reasoning and action extraction per turn." In practice, the `reasoning` field on every `DecisionBranch` is always one of exactly three hardcoded template strings — never the model's actual reasoning.

## Root cause

`DecisionTracker.recordToolCall(record: ToolCallRecord)` (`src/metrics/decision-tracker.ts:59`) takes only tool-call metadata — there is no parameter carrying model reasoning text. The three `reasoning` values it produces are fixed templates: `` `recovery after ${lastToolName} failure` `` (line 67), `'delegating to user'` (line 76), and `` `retrying ${toolName} on ${filePath} (${count} attempts)` `` (line 91). It's also not "per turn" — a branch is only recorded on 3 narrow triggers (previous call failed, `AskUserQuestion` called, or a 3rd+ same-tool-same-file call), so most ordinary successful turns produce zero branches.

This is now disclosed in the tool's registered description and response payload (see the docs/payload correction that closes this issue's sibling work), but the underlying limitation is still real.

## This data is obtainable

Verified directly against a real transcript JSONL (not assumed): Claude Code's transcripts contain separate `thinking` and `text` content-block entries immediately preceding each `tool_use` entry, in file order — one block type per assistant JSONL line, e.g. `('thinking',) -> ('tool_use',)`. The model's actual reasoning **is present on disk**, readable via the same `transcript_path` mechanism `readLastAssistantUsage()` (`src/hooks/collector-script.ts`) already uses in production for token usage.

## Why this isn't trivial

Building a real fix requires:
1. Extending the transcript reader to capture `thinking`/`text` blocks, not just the `usage` object it reads today.
2. Correlating each captured reasoning block to the specific tool call that follows it (likely via `tool_use_id`, which needs verifying against the PreToolUse/PostToolUse hook payload's own id field).
3. Reworking `DecisionTracker.recordToolCall()`'s signature to accept the extracted reasoning text, and updating all call sites/tests.

This is comparable in scope to the deferred message-counting fix (#122) — a real subsystem needing its own brainstorming cycle, not a quick fix.

## Scope note

Found during the MCP tools audit's "unsure" classification for this tool, then independently re-verified against a real transcript during design of the docs/payload correction. Deferred rather than bundled into that correction, matching how #122 and #123 were split out from their sibling fixes.

chrisdhaan:
## Summary

- `nr_observe_get_decision_tree`'s `reasoning` field was always one of 3 hardcoded template strings, never the model's actual reasoning, despite the tool's description promising real extraction.
- Passes `transcript_path` through the hook pipeline unconditionally (mirrors the existing `cwd` field): `src/hooks/collector-script.ts`, `src/hooks/event-processor.ts`.
- New isolated module `src/metrics/transcript-reasoning.ts` tail-reads a transcript JSONL, groups content blocks by turn, and extracts the model's real `thinking` text (falling back to visible `text`) for a given `tool_use` id — returns `null` on any miss, redacts via the existing `redactSensitive()` before returning.
- `DecisionTracker` now calls this extraction at its 3 existing narrow trigger sites (recovery-after-failure, `AskUserQuestion` delegation, 3rd+ same-file retry), gated behind a new `recordContent` option (default off), falling back to the prior template strings when disabled or on a miss.
- Wires the real config's `recordContent` (already `highSecurity`-gated) into `DecisionTracker`'s construction, and rewords the tool's disclosure text to describe the new behavior accurately.
- Version bump to 1.4.45 + CHANGELOG entry.

Fixes #124

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 131/132 suites passed (1 pre-existing skip), 4137/4140 tests passed (3 pre-existing skips)
- [x] `npm run lint` — clean, 0 errors/0 warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.45] - 2026-07-15
+
+### Fixed
+
+- Fixed `nr_observe_get_decision_tree`'s `reasoning` field always being one of 3 hardcoded template strings instead of the model's actual reasoning. When content recording is enabled and the underlying model exposes plaintext thinking or visible text for that turn, the field now reflects the model's real reasoning (passed through the existing redaction filter), falling back to the prior rule-based label when the model exposes no plaintext reasoning.
+
 ## [1.4.44] - 2026-07-15
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.44",
+  "version": "1.4.45",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.44",
+      "version": "1.4.45",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.44",
+  "version": "1.4.45",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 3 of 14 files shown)


corinnekunze:

PR #210, opened 2026-07-15
Close OTLP forward-agent leak, fix fetch test cleanup, correct AlertBanner regex

### LINKED ISSUE #173: OtlpReceiver: undici Agent not closed on stop() + pre-existing globalThis.fetch test-pollution bug
## Description

Two small, unrelated follow-ups surfaced during the v1.4.24 SSRF/DNS-rebinding fix — bundled here rather than filed separately given their size.

## Items

1. **`OtlpReceiver.stop()` doesn't close the `undici.Agent` it constructs.** The DNS-rebinding fix added a `forwardDispatcher: Agent | undefined` field (`src/proxy/otlp-receiver.ts:45`), constructed in the constructor whenever `forwardEndpoint` is configured (`otlp-receiver.ts:50`) and used as the `dispatcher` for every forwarded request (`otlp-receiver.ts:376`). `stop()` (`otlp-receiver.ts:79-86`) closes the HTTP server but never calls `this.forwardDispatcher?.close()` / `.destroy()`. No observed impact today — `OtlpReceiver` is constructed once per proxy process and `stop()` is only reached at process shutdown, so idle keep-alive sockets are reclaimed at process exit regardless — but a future caller that repeatedly constructs/discards `OtlpReceiver` instances (e.g. a config-reload path) would leak idle sockets until GC/idle-timeout.

   **Fix:** add `await this.forwardDispatcher?.close();` to `stop()`, alongside the existing server-close logic.

2. **Pre-existing test-pollution bug: `otlp-receiver.test.ts` clobbers `globalThis.fetch` instead of restoring it.** Two `describe` blocks' cleanup — `describe('forward', ...)`'s `afterEach` (`otlp-receiver.test.ts:194`) and `describe('error message sanitization', ...)`'s `beforeEach` (`otlp-receiver.test.ts:695`) — both set `(globalThis as { fetch?: unknown }).fetch = undefined` as their idea of "un-mocking," rather than restoring the real native `fetch`. Since Jest runs a file's `describe` blocks in declaration order within one worker process, any *later* sibling block that needs a working real `fetch` will find `globalThis.fetch` permanently `undefined` by the time its tests run — regardless of whether that sibling avoids the *mocking* blocks' own `beforeEach` (it does; the bug is in the *cleanup*, not the mock setup). Found and worked around locally (not fixed at the source) while adding the v1.4.24 DNS-rebinding regression tests — see `otlp-receiver.test.ts:268-270`'s comment for the full diagnosis; the workaround captures the real `fetch` reference at module-load time and restores it in a scoped `beforeEach`/`afterEach`, but the two original blocks are still buggy for the next person who adds a test after them.

   **Fix:** in both `afterEach`/`beforeEach` blocks, replace `(globalThis as { fetch?: unknown }).fetch = undefined;` with restoring a captured reference to the real, pre-mock `fetch` (the same pattern already used in the new `'forward — DNS rebinding protection'` describe block).

## Scope note

Found during the v1.4.24 proxy SSRF/DNS-rebinding fix (2026-07-12). Neither item blocked that release — both are correctly-scoped, non-blocking hygiene follow-ups per that PR's final review.

### LINKED ISSUE #183: AlertBanner.test.tsx: stale assertion asserts raw values instead of formatNumber()'s 1-decimal tier output

chrisdhaan:
## Summary
- `OtlpReceiver.stop()` now closes its outbound `undici.Agent` dispatcher (previously left open, leaking the connection pool for the life of the process), and never hangs even if closing that dispatcher rejects.
- Fixed a test-cleanup bug in `otlp-receiver.test.ts` where two `afterEach` blocks set `globalThis.fetch = undefined` instead of restoring the real implementation, creating file-order test pollution that a third describe block had to work around.
- Corrected a stale regex in `AlertBanner.test.tsx` that no longer matched the component's real number formatting (a 1-decimal rounding tier for values ≥10).
- Version bump to 1.4.46.

Fixes #173. Fixes #183.

## Test plan
- [x] `npm test` — 131/132 suites passed (1 pre-existing skip), 4141/4144 tests passed (3 pre-existing skips), no failures
- [x] `npm run test:web` — 23/23 files, 279/279 tests passed
- [x] `npm run build` — clean
- [x] `npm run lint` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.46] - 2026-07-15
+
+### Fixed
+
+- The OTLP receiver now closes its outbound connection pool when stopped, and no longer hangs if closing that pool fails, instead of leaving the pool open for the life of the process.
+- Corrected a test assertion in the alert banner suite that no longer matched the component's real number formatting.
+
 ## [1.4.45] - 2026-07-15
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.45",
+  "version": "1.4.46",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.45",
+      "version": "1.4.46",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.45",
+  "version": "1.4.46",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/proxy/otlp-receiver.test.ts
+++ b/src/proxy/otlp-receiver.test.ts
@@ -19,6 +19,11 @@ import { gzipSync, deflateSync, brotliCompressSync } from 'node:zlib';
 import { OtlpReceiver } from './otlp-receiver.js';
 import type { OtlpReceiverOptions } from './otlp-receiver.js';
 
+// Captured once at module load, before any test mocks globalThis.fetch, so any
+// describe block can restore the real implementation instead of leaving it
+// undefined for whichever block runs next.
+const realFetch = globalThis.fetch;
+
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
@@ -213,7 +218,7 @@ describe('forward', () => {
   });
 
   afterEach(() => {
-    (globalThis as { fetch?: unknown }).fetch = undefined;
+    globalThis.fetch = realFetch;
   });
 
   it('calls fetch with the forward URL and api-key header', async () => {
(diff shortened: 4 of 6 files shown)


corinnekunze:

PR #211, opened 2026-07-15
Fix anti-pattern REST hydration mapping

### LINKED ISSUE #186: useLiveEvents.ts: /api/anti-patterns REST hydration pushes target/count as undefined into the live store

chrisdhaan:
## Summary
- `useLiveEvents.ts`'s `hydrateFromApi()` REST-hydration path force-cast the fetched anti-pattern objects into a shape they don't have (`{ type, target, count }`), so `target`/`count` were always `undefined` in the dashboard until the first live SSE event arrived.
- Replaced the cast with an explicit field mapping (`file`/`command` → `target`, `iterations`/`readCount`/`repeatCount`/`editCount`/`agentCount` → `count`), identical to the mapping `Today.tsx` already uses for the same data shape.
- Added a regression test exercising the real hook against a mocked `/api/anti-patterns` response.
- Version bump to 1.4.47 + CHANGELOG entry.

Fixes #186

## Test plan
- [x] New test: `src/web/hooks/useLiveEvents.test.tsx` — REST hydration maps real `AntiPattern` fields onto the store correctly
- [x] `npm test` — 4141/4144 passed (3 pre-existing skips), 0 failures
- [x] `npm run test:web` — 280/280 passed
- [x] `npm run build` — clean
- [x] `npm run lint` — clean
- [x] `npx tsc --noEmit -p tsconfig.web.json` — no new errors vs. baseline

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.47] - 2026-07-15
+
+### Fixed
+
+- The live dashboard's pre-connection anti-pattern hydration now maps the server's anti-pattern fields (file/command, iteration/read/repeat/edit/agent counts) onto the display shape correctly, instead of leaving the target file and count blank until the first live event arrives.
+
 ## [1.4.46] - 2026-07-15
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.46",
+  "version": "1.4.47",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.4.46",
+      "version": "1.4.47",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.4.46",
+  "version": "1.4.47",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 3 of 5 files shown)


corinnekunze:

PR #214, opened 2026-07-16
Chore: v1.5.1 — watcher hardening, GanttTimeline perf/a11y, Sessions test coverage

### LINKED ISSUE #212: watcher hardening, GanttTimeline perf/a11y, workflow-consolidation test coverage
## Summary

`1.5.0` ("subagent + workflow observability — close 12.93x cost undercount") went through two rounds of independent review (5 parallel slices each, before and after a round of fixes). No Critical or correctness-blocking issues were found in either round, and the PR merged. This issue tracks the real, non-blocking follow-up items the second review pass surfaced — reliability hardening, one perf issue, one a11y gap, and one test-coverage gap on the PR's own headline feature. None of these are regressions; they're pre-existing conditions in the merged code worth addressing as a fast-follow.

## 1. Silent token-loss risk if the subagent transcript filename format ever changes

`src/hooks/subagent-watcher.ts`'s agent-id regex is a fixed-length shape:

```ts
const AGENT_ID_RE = /^a[a-f0-9]{16}$/;
```

Any agent transcript file whose id doesn't match this exact 17-character shape is silently skipped during discovery — no `discovery_skipped` health event, no log, nothing. If a future Claude Code release changes the id format even slightly, this watcher would quietly stop capturing subagent tokens again, reintroducing the exact undercount this PR was built to close, with no observability into why.

For contrast, the sibling `workflow-watcher.ts` deliberately uses a prefix-only match (`^wf_`, documented "R7: no fixed-suffix-length assumption") for exactly this reason.

**Suggested fix:** relax `AGENT_ID_RE` to a prefix/looser pattern, or (simpler, lower-risk) emit a health event when a file under `subagents/` doesn't match the expected shape, so drift is observable instead of silent.

## 2. `WorkflowWatcher`'s in-memory maps are never evicted

`src/hooks/workflow-watcher.ts`'s `seenMtime`, `emittedRuns`, and `topologyCache` maps grow for the lifetime of the process — there's no equivalent of `SubagentWatcher`'s `evictStalePartials()`. In a long-lived `--local` dashboard process scanning `~/.claude/projects`, these grow linearly with total historical workflow-run count, not active count.

**Suggested fix:** evict entries for paths/runs no longer present in the current discovered set, mirroring `SubagentWatcher.evictStalePartials()`.

## 3. Both watchers do a full recursive rescan of `~/.claude/projects` every poll, forever

`discoverFiles()` in both `subagent-watcher.ts` and `workflow-watcher.ts` walks every project → every session → every file on each poll cycle (every 2s by default), with the mtime/session filters applied *after* the directory walk rather than before. For a developer with months of accumulated project history, this is unbounded I/O that scales with total historical data, not active session count, repeated forever for the life of the process. All the filesystem calls are synchronous, so this also briefly blocks the event loop on every poll.

**Suggested fix:** cache the project/session directory listing between polls and only refresh it periodically, and/or short-circuit immediately to the `parentSessionId` filter when one is set (the common case for a per-session `--stdio` server instance) before doing any broader directory walk.

## 4. `GanttTimeline` re-sorts and rebuilds on every mouse hover

`src/web/components/GanttTimeline.tsx`'s `sorted`, `originalToSorted`, and `segmentAt` are recomputed on every render, and hovering any bar triggers a state update (`setHoveredIndex`) that re-renders the component — so hovering over a bar in a lane with many tool calls runs a full sort + segment-map rebuild on every mouse move across bars. This PR made the problem materially worse by embedding `GanttTimeline` repeatedly inside `SessionTrace` (parent lane + every expanded agent's call list), for traces that can have "thousands of tool-call rows" per the component's own comments.

**Suggested fix:** wrap `sorted`/`originalToSanded`/`segmentAt` in `useMemo` keyed on `entries`/`segments`, and consider virtualizing very large row counts.

## 5. `GanttTimeline` bars have no accessible name and are mouse-only

Unlike its sibling `AgentSwimlanes.tsx` (whose bars get `role="img"` + `aria-label`), `GanttTimeline`'s bar elements carry only `onMouseEnter`/`onMouseLeave` — no `role`, no `aria-label`, not focusable. All tooltip content (tool name, duration, success, anti-pattern) is invisible to keyboard and screen-reader users. Since `GanttTimeline` is now the primary trace visualization embedded throughout `SessionTrace`, this gap is more consequential than when it was used in fewer places.

**Suggested fix:** add `role="img"` + an `aria-label` mirroring the tooltip content, matching `AgentSwimlanes`'s existing pattern.

## 6. The PR's headline feature (workflow consolidation into Sessions) is essentially untested

`src/web/views/Sessions.tsx` gained ~359 net new lines implementing the PR's stated purpose — consolidating workflow runs into the sessions-centric view — but `Sessions.test.tsx` added exactly one new test, and that test's fetch mock returns session-shaped data for the `/api/workflows` endpoint, so the KPI aggregation math it's nominally checking never actually executes real logic. There is zero behavioral coverage for: run-level filters scoping the visible row list, the 3-level master-list tree (session → runs → agents) expansion, real KPI aggregation math over actual run data, the `?session=`/`?id=` query-param handling, the in-place workflow-run detail drawer, or `SessionTraceSection`'s shared-window merge logic.

**Suggested fix:** a focused test pass on `Sessions.tsx`'s workflow-consolidation logic, with a correctly-shaped `/api/workflows` mock.

## Also worth a quick pass (Minor, from the same review round)

- `src/web/components/AgentSwimlanes.tsx` imports `'../lib/format'` without the `.js` extension — every sibling new file in this PR uses `.js` per this repo's ESM convention; Vite resolves it fine today but it's a convention slip.
- `src/web/api/client.ts`'s `fetchObservabilityHealth` still returns `Promise<unknown>` (out of scope for the typed-contract cleanup that touched the other 5 fetchers), and `Today.tsx`/`Settings.tsx` each carry their own slightly-divergent local `ObservabilityHealthApiResponse` interface consuming it via a cast. Low-consequence (both locals are safe structural subsets of the real shape) but worth consolidating into one exported type in `client.ts`.
- Duplicated grouping/palette/tick-math logic between `AgentSwimlanes.tsx` and `SessionTrace.tsx` (`groupAgents`, `GROUP_BAR_COLORS`, tick-interval algorithm, span-label formatters) — both are still live (rendered in different contexts), so not dead code, but now two implementations to keep in sync by hand.

chrisdhaan:
## Summary

Closes #212.

- The subagent transcript watcher now emits a health event instead of silently skipping a transcript file whose filename doesn't match the expected agent-id shape, so a future change to that format would be observable instead of silently reintroducing lost subagent cost tracking.
- The workflow watcher no longer accumulates unbounded internal bookkeeping for the life of the process; stale tracking entries are evicted on each poll instead of persisting indefinitely.
- Both the subagent and workflow watchers now go straight to the active session's own directory when tracking a single session, instead of listing every session on disk on every poll.
- The session trace's Gantt timeline no longer re-sorts and rebuilds its highlight state on every mouse movement, and its bars now expose an accessible name to assistive technology.
- Added test coverage for the Sessions view's workflow-consolidation logic: KPI aggregation across workflow runs, run-source and status filtering, the session/run/agent expansion tree, and the in-place workflow-run detail view.
- Fixed a missing file extension on an internal module import.
- Deduped agent-grouping logic that was previously copy-pasted between two components, and consolidated an untyped API response into a shared, exported type.

## Test plan

- [x] `npm run build && npm test` — all suites pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.5.1] - 2026-07-16
+
+### Fixed
+
+- The subagent transcript watcher now emits a health event instead of silently skipping a transcript file whose filename doesn't match the expected agent-id shape, so a future change to that format would be observable instead of silently reintroducing lost subagent cost tracking.
+- The workflow watcher no longer accumulates unbounded internal bookkeeping for the life of the process; stale tracking entries are evicted on each poll instead of persisting indefinitely.
+- Both the subagent and workflow watchers now go straight to the active session's own directory when tracking a single session, instead of listing every session on disk on every poll.
+- The session trace's Gantt timeline no longer re-sorts and rebuilds its highlight state on every mouse movement, and its bars now expose an accessible name to assistive technology.
+- Fixed a missing file extension on an internal module import.
+
+### Added
+
+- Added test coverage for the Sessions view's workflow-consolidation logic: KPI aggregation across workflow runs, run-source and status filtering, the session/run/agent expansion tree, and the in-place workflow-run detail view.
+
 ## [1.5.0] - 2026-07-15
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.5.0",
+  "version": "1.5.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.5.0",
+      "version": "1.5.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.5.0",
+  "version": "1.5.1",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/alerts/alert-snapshot-collector.ts
+++ b/src/alerts/alert-snapshot-collector.ts
@@ -4,7 +4,7 @@ const logger = createLogger('alert-snapshot-collector');
 
 // ---------------------------------------------------------------------------
 // Snapshot type — owns the shape that LocalAlertEngine.evaluate() consumes.
-// (Re-exported from local-alert-engine.ts to keep Phase 1 imports stable.)
+// (Re-exported from local-alert-engine.ts to keep imports stable.)
 // ---------------------------------------------------------------------------
 
 /**
--- a/src/alerts/local-alert-engine.test.ts
+++ b/src/alerts/local-alert-engine.test.ts
@@ -411,7 +411,7 @@ describe('LocalAlertEngine — clock + state housekeeping', () => {
 });
 
 // ---------------------------------------------------------------------------
-// Phase 2 — non-budget rule types
+// Non-budget rule types
 // ---------------------------------------------------------------------------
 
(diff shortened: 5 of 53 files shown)


corinnekunze:

PR #215, opened 2026-07-16
Refactor: tighten unknown typing across backend (ApiHandlerDeps, adapters, config, transcripts, storage)

### LINKED ISSUE #216: Tighten unknown typing across the backend
## Summary

Several parts of the backend use `unknown` (or re-cast already-validated data back to `Record<string, unknown>`) in places where the real shape is either already known at compile time or already validated at runtime — e.g. dependency-injection fields typed as `unknown` when the concrete tracker/store class's return type is known, config loading that validates against a zod schema and then discards the inferred type, and ad-hoc `as Record<string, unknown>` casts on JSON this codebase itself writes and reads back.

None of this is a functional bug, but it weakens compile-time safety and makes some code harder to follow than it needs to be.

## Goal

Audit `unknown` usage across the backend `src/` and replace it with concrete types, `Partial<T>`, generics, or discriminated unions wherever doing so doesn't change runtime behavior — while leaving `unknown` in place wherever it's the honest type (unvalidated external API responses, generic JSON-parsing helpers, etc.).

## Scope

- MCP server dependency-injection interfaces
- Platform adapter event normalization (adds real runtime type guards where previously missing)
- Config loading / zod schema round-tripping
- Transcript parsing (Claude Code JSONL)
- Storage round-trip types (session store, local store)
- NerdGraph / OTLP / Slack Block Kit response shapes

chrisdhaan:
## Summary

Removes unnecessary `unknown` typing across the backend `src/` codebase, in a series of independent, mostly type-only refactors:

- Retypes `ApiHandlerDeps`'s dependency-injection fields with the real return types of the tracker/store classes they wrap, and removes redundant `Record<string, unknown>` re-casts of already-typed MCP tool args in `session-stats.ts`.
- Adds runtime type guards to 7 platform adapters' `normalizeToolCall`, which previously did an unchecked cast that would throw on `null`/non-object input — no production code currently calls these methods, so this is a low-risk, forward-looking hardening.
- Retypes `config.ts`/`setup-wizard.ts`'s zod-validated config loading to return the real inferred schema type instead of re-widening validated data back to `Record<string, unknown>`.
- Introduces shared transcript-parsing types (`transcript-types.ts`) for Claude Code transcript JSONL parsing, reused across the subagent watcher, subagent timeline store, and hook collector.
- Names storage round-trip shapes (`session-store.ts`, `local-store.ts`) to remove duplicate ad-hoc casts on JSON this codebase itself wrote.
- Names NerdGraph/OTLP/Slack Block Kit/settings.json response shapes across `cross-session-tools.ts`, `workflow-tools.ts`, `key-validator.ts`, `diagnostics.ts`, `otlp-receiver.ts`, `deploy-alerts.ts`, and `digest-formatter.ts`.
- Replaces a `MetricAggregator` monkey-patch (cast-based method reassignment) with a proper subclass override in `nr-ingest.ts`.

All changes are type-only refactors with zero runtime behavior change, except the platform adapter guards noted above (intentional, low-risk).

## Test plan

- [x] `npx tsc -b .` — 0 errors

Fixes #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.5.2] - 2026-07-16
+
+### Changed
+
+- Tightened internal typing across the MCP server, platform adapters, config loading, transcript parsing, and storage layers — replacing loosely-typed values with concrete types wherever the underlying data was already validated or structurally known. No user-visible behavior change.
+
 ## [1.5.1] - 2026-07-16
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.5.1",
+  "version": "1.5.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.5.1",
+      "version": "1.5.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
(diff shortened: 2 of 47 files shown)


corinnekunze:

PR #218, opened 2026-07-16
Fix 3 deferred unknown-type audit items

### LINKED ISSUE #217: Fix 3 deferred unknown-type audit items
Three follow-up items deferred from the earlier unknown-type PR (#215):

1. **Anti-pattern shape mismatch**: the live-session dashboard route returned per-occurrence anti-pattern objects instead of the `{type,count}[]` aggregate the frontend expects, via an unsafe type cast.
2. **Readonly-field mutation**: turn attribution (`turn_id`/`turn_number`) was attached to a tool-call record by mutating a field declared `readonly`, via a forced cast.
3. **Hook event buffer typing**: the internal hook event buffer used a single loosely-typed shape for every event mode, requiring an unchecked cast to narrow it for one specific mode in the event-processing pipeline.

None of these were functional bugs — the PR deliberately deferred them because fixing them safely needed more investigation than a mechanical batch fix allows.

chrisdhaan:
## Summary
- Fix the anti-pattern detector shape mismatch in the live-session dashboard route (aggregate by type instead of an unsafe cast).
- Replace a readonly-field mutation used to attach turn attribution to tool-call records with a properly-typed record construction.
- Turn the internal hook event buffer's type into a real discriminated union (shared base + 6 per-mode variants), removing an unchecked cast from the event-processing pipeline.

Fixes #217

## Test plan
- [x] `npm run build && npm test && npm run lint` all clean
- [x] New test proving anti-pattern aggregation
- [x] New test proving turn attribution still propagates correctly
- [x] Existing hook event processor and local store tests pass with an unchanged pass count after the type retype

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.5.3] - 2026-07-16
+
+### Fixed
+
+- The dashboard's live-session view now aggregates anti-pattern occurrences by type before returning them, instead of returning a mismatched shape the frontend couldn't render correctly.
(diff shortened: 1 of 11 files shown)


corinnekunze:

PR #222, opened 2026-07-17
Fix: serve live workflow-run detail instead of 404

### LINKED ISSUE #223: Workflow run detail view shows "Failed to load run details" for runs still in progress
## Summary

Opening the detail view for a workflow run that is still in progress shows "Failed to load run details" instead of live progress. The session-trace swimlane makes every run lane clickable as soon as its subagent activity exists — well before the run finishes — but the detail view can't render anything until the run terminates.

## Root cause

The workflow-run detail endpoint only reads the run's on-disk rollup file, and that file is written once at termination (completed, killed, or failed) — never while the run is active. So clicking the most recent (still-running) run finds no rollup, the endpoint returns 404, and the drawer falls into its error state.

## Expected behavior

Opening the detail drawer for a still-running run should show live per-agent progress — tokens, tool calls, timing, and the declared topology/name — and automatically upgrade to the final rollup summary once the run completes.

## Suggested fix

When no rollup exists yet, assemble a "running" detail from the run's live subagent transcripts (and its persisted orchestration script, for topology/name), returned in the same shape as the completed-run response so the existing UI renders it without any changes.

chrisdhaan:
## Summary

Fixes #223.

A running workflow has no on-disk rollup file until it terminates, but the session-trace swimlane makes every run lane clickable from its live subagent transcripts, which exist mid-run. Opening the detail drawer for the most recent (still-running) run therefore always hit the terminal-only rollup lookup, got nothing back, and showed "Failed to load run details."

This adds a live fallback: when the rollup lookup finds nothing, the detail route synthesizes a running-status detail from the run's live subagent transcripts — per-agent tokens/tool-calls/timing, summed totals, and the declared topology/name resolved from the persisted orchestration script. It emits the same wire shape as the completed path, so the drawer renders unchanged and automatically upgrades to the final summary once the run terminates.

Also includes two follow-up fixes from review:
- The live view's reported default model is now chosen deterministically (the most recently active agent) rather than depending on unspecified directory-listing order.
- The run's on-disk location is cached so repeated polls of a still-running run's detail don't repeat the full directory scan.

## Test plan

- [x] `npm run build` clean
- [x] `npm run lint` — no new issues
- [x] `npm run format:check` clean
- [x] `npm test` — full suite green
- [x] New unit tests for the live-detail assembly (cost/token/tool-call correctness, deterministic default model, run-location cache reuse/invalidation, script/topology resolution, malformed/unknown run ids)
- [x] New route tests for the live-fallback branch (rollup path, live fallback, name fallback, 404, missing dependency)
- [x] New end-to-end integration test wiring the real stores against a real temp filesystem tree

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.5.5] - 2026-07-17
+
+### Fixed
+
+- Opening a workflow run's detail view while the run was still in progress showed "Failed to load run details" instead of live progress. The detail view now shows in-progress agent activity and automatically upgrades to the final summary once the run completes.
+
 ## [1.5.4] - 2026-07-17
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.5.4",
+      "version": "1.5.5",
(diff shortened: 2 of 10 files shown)


corinnekunze:

Issue #225, opened 2026-07-17
Centralize synthetic session-id filtering at the SessionStore/LiveSessionRegistry boundary


chrisdhaan:
## Summary

Synthetic session ids (`local-*`, `proxy-*`, `pending-*` — internal bookkeeping ids from `--local` / proxy modes, not real Claude Code sessions) need to be filtered out of session data wherever it reaches the dashboard, but that filtering is currently applied ad hoc at each call site via `isSyntheticSessionId()` rather than once at a shared producer boundary.

## Context

Noted as a follow-up while fixing #224 (the Today concurrency chart/headline peak mismatch): that fix had to apply `isSyntheticSessionId()` filtering in two separate places within the `/api/concurrency` route handler alone (once for today's bucket sessions, once for the all-time peak sessions) rather than relying on a single already-filtered source. Every new consumer of session data has to remember to apply the same filter itself, which is easy to forget.

## Goal

Move synthetic-session-id filtering into `SessionStore` / `LiveSessionRegistry` (or another shared producer boundary) so downstream consumers get already-filtered data by default.

## Scope

- `SessionStore.loadTodaySessions()` / `loadAllSessions()` / `loadSessionsOverlappingToday()`
- `LiveSessionRegistry.getLiveSessions()`
- Every existing call site in `src/dashboard/routes/api-handler.ts` that currently calls `isSyntheticSessionId()` itself

PR #226, opened 2026-07-17
Fix: Today concurrency chart peak disagreeing with headline peak

### LINKED ISSUE #224: Today concurrency chart peak can disagree with the headline "today peak" KPI
## Summary

The Today page's "Concurrent Sessions" card could show a headline peak (e.g. `4`) lower than the tallest column in its own chart (e.g. `7`).

## Root cause

The headline and the chart bars were computed from two different concurrency models:

- **Headline** used a 3-minute activity-window model (`mergeActivityWindows`), matching `LiveSessionRegistry`'s liveness definition.
- **Chart buckets** used whole-session `[startTime, endTime]` spans, with live sessions extended to `now`.

An idle-but-not-yet-ended session was counted as continuously concurrent under the span model, so a session that had gone quiet (but not yet ended) inflated recent chart buckets above the true activity-window peak.

## Fix

Replace the span-based interval builder with an activity-window builder that unions each session's persisted timeline with its live tool-call-buffer activity, using the same 3-minute activity-window model the headline already used. The headline peak is now derived directly from the same chart buckets, so the two numbers can never disagree by construction.

Also fixed as part of the same change:
- A bucket-boundary tiebreak bug where one session's close coinciding exactly with another's open (on a 15-minute bucket grid line) could undercount the momentary overlap.
- The headline no longer folds in `LiveSessionRegistry`'s never-reset lifetime-max `peak`, which could reintroduce the same class of disagreement.
- Synthetic session ids (`local-*`, `proxy-*`) are now filtered consistently from both the chart buckets and the all-time peak.

chrisdhaan:
## Summary

- The Today page's "Concurrent Sessions" card could show a headline peak (e.g. `4`) lower than the tallest column in its own chart (e.g. `7`), because the two numbers were computed from different concurrency models: the headline used a 3-minute activity-window model (`mergeActivityWindows`), while the chart buckets used whole-session `[startTime, endTime]` spans with live sessions extended to `now` — so an idle-but-not-yet-ended session was counted as continuously concurrent and inflated recent buckets above the true peak.
- Replaces `collectTodaySessionIntervals` (span model) with `collectTodayActivityWindows`, which builds the same activity-window model the headline already used, unioning each session's persisted timeline with its live tool-call-buffer activity per session id.
- Derives the headline `historicalPeak` directly from the chart buckets (`buckets.reduce(...)`) rather than from an independent computation, so the two numbers can never disagree by construction.
- Also fixes a bucket-boundary tiebreak bug where one session's close coinciding exactly with another's open (on a 15-minute bucket grid line) could undercount the momentary overlap, and filters synthetic session ids (`local-*`/`proxy-*`) consistently from both the chart buckets and the all-time peak.

Fixes #224.

A follow-up to centralize the ad hoc synthetic-session-id filtering used here is tracked separately in #225.

## Test plan

- [x] `npx tsc -b .` — clean
- [x] `npm run lint` — no new issues
- [x] `npx jest -- src/dashboard/routes/api-handler.test.ts` — all passing (rewrote the tests that encoded the old whole-session-span model onto the activity-window model)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.5.6] - 2026-07-17
+
+### Fixed
+
+- The Today page's "Concurrent Sessions" chart could show a taller peak bar than the headline peak number above it. Both are now computed from the same activity-window model, so an idle-but-not-yet-ended session no longer inflates the chart above what the headline reports.
+
 ## [1.5.5] - 2026-07-17
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.5.5",
+  "version": "1.5.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.5.5",
+      "version": "1.5.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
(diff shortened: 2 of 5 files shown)


corinnekunze:

PR #227, opened 2026-07-18
Feat: track cost per 1M tokens (session-blended + per-model)

### LINKED ISSUE #228: Track cost per 1M tokens (session-blended + per-model)
## Problem

Raw session/model cost totals aren't comparable across sessions or models with different token volumes — a session that used 10x the tokens will naturally cost more, even if it's not less efficient per token.

## Proposal

Add a normalized cost-rate metric, cost per million tokens, so spend can be compared across sessions and models independent of raw token volume:

- A session-blended rate (across input, output, thinking, cache-read, and cache-creation tokens) from `nr_observe_get_cost_breakdown`.
- A per-model rate (input+output tokens) from `nr_observe_get_model_usage`, surfaced in the dashboard's Today and History views.

chrisdhaan:
Fixes #228

## Summary
- Adds a normalized cost-rate metric, cost per million tokens, so spend can be compared across sessions and models independent of raw token volume.
- `CostTracker`: session-blended rate across all token types (input/output/thinking/cache-read/cache-creation), exposed via `nr_observe_get_cost_breakdown` (`cost_per_million_tokens` field). Not shipped as its own NR metric — it's a pure ratio of the cost/token metrics already emitted, so it's more flexibly computed in NRQL at query time.
- `ModelUsageTracker`: per-model rate (input+output tokens) on `ModelStats`, exposed via `nr_observe_get_model_usage` and shown in the dashboard's Today and History views.
- Updated `docs/COMMANDS_TABLE.md`, including a correction to the `nr_observe_get_model_usage` entry's documented response shape to match the handler's actual output.

## Test plan
- [x] `npm run build && npm test` — full suite passes
- [x] `npx vitest run` — web test suite passes
- [x] New/updated unit tests for the blended-rate math, the per-model rate, and the live-session cost/token accounting edge case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.0] - 2026-07-17
+
+### Added
+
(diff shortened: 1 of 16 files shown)


corinnekunze:

PR #229, opened 2026-07-18
Fix #122: real message counts for nr_observe_get_collaboration_profile

### LINKED ISSUE #122: nr_observe_get_collaboration_profile returns fixed fallback values — message counts never populated
## Description

`nr_observe_get_collaboration_profile` is meant to classify a developer's working style (specificity, autonomy, correction rate) from their session history. In practice every developer gets the same fixed fallback values: `specificity: 0.5`, `autonomy: 0.5`, `correctionRate: 1`, classification always `'Collaborative'`.

This also degrades two other tools that consume the same data: `nr_observe_get_claudemd_impact`'s `avgCorrectionRate` sub-field, and `nr_observe_get_recommendations`'s prompt-engineering category (via `CollaborationProfiler`).

## Root cause

`buildSessionSummary()` in `src/storage/session-store.ts` hardcodes `userMessages: 0`, `assistantMessages: 0`, and `userCorrections: 0` on every persisted session summary. Nothing populates real values, so every downstream computation (`computeSpecificity`, `computeAutonomy`, `computeCorrectionRate` in `src/metrics/collaboration-profile.ts`) hits its zero-input fallback branch every time.

## This data is obtainable

Claude Code's hook payloads include a `transcript_path` field pointing to a JSONL file with the full conversation. `src/hooks/collector-script.ts`'s `readLastAssistantUsage()` already tail-reads this file and parses `type: 'assistant'` entries to extract token usage — the mechanism is proven and in production. The same file should contain `type: 'user'` entries, meaning real message counts are computable from data already flowing through the system, not something that needs new instrumentation from zero.

## Open design questions

1. Where should transcript parsing happen — incrementally in `collector-script.ts` (new hook-event type, similar to the existing `'token'` mode event), or one-shot in `buildSessionSummary()` at session-persist time?
2. What heuristic defines a "user correction"? No existing code in the codebase has this heuristic today — needs real design, not a threshold tweak.
3. `transcript_path` is a Claude Code hook field. Other platforms (Cursor, Windsurf, etc.) don't have an equivalent today — the fix must degrade gracefully (stay at 0) on platforms without this data, not regress anything.

## Scope note

This was split out from the platform-field bug (same root cause in `buildSessionSummary()`) once it became clear the two halves had very different weight — the platform field was trivial wiring (shipped), this is a real subsystem needing its own design pass.

chrisdhaan:
## Summary

`nr_observe_get_collaboration_profile` returned the same fixed fallback values for every developer because `buildSessionSummary()` hardcoded `userMessages`/`assistantMessages`/`userCorrections` to `0` on every persisted session. This also degraded `nr_observe_get_claudemd_impact`'s `avgCorrectionRate` and `nr_observe_get_recommendations`'s prompt-engineering category.

- Adds `TranscriptMessageTracker`, which incrementally tails the Claude Code transcript JSONL file and classifies each `type: 'user'`/`type: 'assistant'` line as a real human/model message vs. one of several synthetic entry types the harness also writes (tool-result echoes, local-command caveats, context-compaction summaries, task notifications, subagent sidechain turns, and a few content-prefix-matched cases with no structural marker).
- Wires the tracker into `buildSessionSummary()` and the MCP server's session lifecycle, refreshing on the existing 30s periodic checkpoint and at shutdown — not per tool call.
- Platforms without `transcript_path` (Cursor, Windsurf, etc.) see unchanged all-zero behavior — no regression.
- Includes a fix for a stall bug found during review: the transcript reader would permanently stop making progress if a single transcript line exceeded the 1MB read cap (realistic for large tool outputs); it now skips forward past an oversized line instead.

Bumps the version to 1.6.1 and adds a CHANGELOG entry.

Fixes #122

## Test plan

- [x] `npm run build && npm test && npm run lint` — clean, 0 lint errors/warnings
- [x] New unit tests for `TranscriptMessageTracker` covering every synthetic-entry pattern (tool-result echoes, `isMeta`, `isCompactSummary`, `task-notification` origin, `isSidechain`, content-prefix cases, correction regex, incremental refresh, file rotation, oversized-line skip-mode, reset)
- [x] Extended `buildSessionSummary` tests for tracker present/absent cases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.1] - 2026-07-18
+
+### Fixed
+
+- Developer collaboration profiles previously showed the same values for every developer because real conversation activity wasn't being counted. Message and correction counts are now derived from the actual session transcript, so specificity, autonomy, and correction-rate scores reflect real usage. This also corrects the correction-rate figure surfaced in CLAUDE.md impact analysis and in prompt-engineering recommendations.
+
 ## [1.6.0] - 2026-07-17
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.6.0",
+      "version": "1.6.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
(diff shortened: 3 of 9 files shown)


corinnekunze:

PR #230, opened 2026-07-18
Fix #225: centralize synthetic session-id filtering at the producer boundary

### LINKED ISSUE #225: Centralize synthetic session-id filtering at the SessionStore/LiveSessionRegistry boundary

chrisdhaan:
## Summary

- Filters synthetic session IDs (`local-*`, `proxy-*`, `pending-*`) inside `SessionStore.loadAllSessions()` instead of at each call site — `loadTodaySessions()` and `loadSessionsOverlappingToday()` both delegate to it, so they're fixed for free.
- Adds an `includeSynthetic` option to `LiveSessionRegistry.getLiveSessions()`, defaulting to filtered; internal peak/count/time-series tracking (`touch()`, `getConcurrentCount()`, `startSampling()`) explicitly opts into `{ includeSynthetic: true }` so those numbers are unchanged.
- Fixes the orphan-buffer GC pass in `src/index.ts` to request `{ includeSynthetic: true }`, so it keeps protecting still-live synthetic-ID session buffers from deletion now that the default is filtered.
- Removes the ad-hoc `isSyntheticSessionId()` filtering from 5 of 6 call sites in `src/dashboard/routes/api-handler.ts`, now that the producers guarantee clean data. The 6th (the current-live-session push in `GET /api/sessions`) reads `SessionTracker`'s own unfiltered sessionId — not backed by either centralized producer — and is left as-is with an explanatory comment.

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [x] `npm run lint`

Fixes #225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.2] - 2026-07-18
+
+### Fixed
+
+- Internal bookkeeping session IDs from `--local` and proxy mode could occasionally still show up in the session list, live sessions view, or concurrency chart, since each dashboard endpoint filtered them independently and inconsistently. That filtering is now applied once, centrally, wherever session data is loaded, so it can no longer be missed for any given view.
+
 ## [1.6.1] - 2026-07-18
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.1",
+  "version": "1.6.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.6.1",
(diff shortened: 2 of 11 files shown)


corinnekunze:

PR #250, opened 2026-07-20
Fix: Unify secret-redaction patterns, evictedKey leak, recordContent gate (v1.6.7)

### LINKED ISSUE #241: Secret-redaction pattern list has drifted between config.ts and the hook collector
## Problem

`src/config.ts`'s `DEFAULT_REDACTION_PATTERNS` (16 regexes) and `src/hooks/collector-script.ts`'s `REDACTION_PATTERNS` (8 regexes, commented "mirrors config.ts DEFAULT_REDACTION_PATTERNS") have drifted. The hook-collector's copy is missing 8 pattern categories present in the canonical list: DB connection strings with embedded credentials, Stripe live/test keys, PyPI tokens, HuggingFace tokens, Twilio SIDs, Azure SAS tokens, and generic Vercel/Heroku/Datadog tokens.

`collector-script.ts` is the first point of contact for raw tool-call content (Bash commands, file paths, etc.) — so any of these 8 secret shapes typed into a tool call is captured to the local buffer file unredacted at the earliest possible point.

Two related gaps in the same area:

1. **`evictedKey` diagnostic is silently blanked.** `src/hooks/event-processor.ts` logs `logger.warn('Evicting non-orphan pre-event due to capacity overflow', { evictedKey })`. The field name `evictedKey` matches `SECRET_KEY_RE` in `src/shared/redact.ts` (any key-name containing "key"), so the logger's automatic redaction replaces the value with `'***'` — defeating the diagnostic's purpose. The existing test only asserts the message text, not that the value survives.
2. **`highSecurity → recordContent=false` is implemented independently in multiple places** (`src/config.ts`, `src/hooks/collector-script.ts`, and the vendored `src/shared/config.ts`) with inconsistent env var names, and no shared gate a new content-reading feature is guaranteed to inherit.

## Proposed fix

- Extract the canonical redaction pattern list into one module imported by both `config.ts` and `collector-script.ts`; add a test asserting they stay in sync.
- Rename `evictedKey` to a name that doesn't match `SECRET_KEY_RE`; add a regression test asserting the value (not just the message) survives in the log output.
- Consolidate the two locally-owned `recordContent` gate implementations behind one shared helper.

## Affected files

- `src/config.ts`
- `src/hooks/collector-script.ts`
- `src/hooks/event-processor.ts`
- `src/shared/redact.ts`

chrisdhaan:
## Summary

- Unified two independently-maintained secret-redaction pattern lists into a single shared module, `src/redaction-patterns.ts`. The hook collector — the highest-volume raw-content capture point — was missing coverage for DB connection strings, Stripe/PyPI/HuggingFace tokens, Twilio SIDs, and Azure SAS tokens.
- Fixed `evictedKey` diagnostic field being silently redacted to `***` by the logger's key-name-based secret detection (same bug class fixed for `unknownKeys`→`unknownFields` in 1.6.6). Renamed to `evictedPairingId`.
- Consolidated two independently-implemented copies of the `highSecurity → recordContent=false` gate behind one shared helper, `src/record-content-gate.ts`.
- Version 1.6.6 → 1.6.7 + CHANGELOG entry.

Closes #241

## Test plan

- [x] `npx tsc -b .` — no errors
- [x] Full suite passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.7] - 2026-07-20
+
+### Security
+
+- The hook collector binary's secret-redaction pattern list had fallen out of sync with the canonical list used elsewhere in the server, missing coverage for database connection strings with embedded credentials, Stripe/PyPI/HuggingFace tokens, Twilio SIDs, and Azure SAS tokens. Both now import from a single shared pattern list, so the two can no longer drift.
+
+### Fixed
+
+- The "Evicting non-orphan pre-event due to capacity overflow" warning was masking the actual pending-event identifier behind `***`, for the same reason as the 1.6.6 fix — the log field's own name was treated as sensitive by the secret-redaction logic. The warning now shows the real identifier.
+- The `recordContent` gate (which `highSecurity` forces off) was implemented independently in two places with duplicated logic. Both now call one shared function so the behavior can't diverge between them.
+
 ## [1.6.6] - 2026-07-19
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


corinnekunze:

PR #252, opened 2026-07-20
Refactor: Give src/tools/ a per-feature MCP registration seam

### LINKED ISSUE #243: src/tools/ has no per-feature MCP registration seam — everything funnels through one switch
## Problem

`src/tools/session-stats.ts` registers all 40+ MCP tools through a single `registerTools()` function containing a ~1400-line switch statement. The sibling files (`cost-tools.ts`, `workflow-tools.ts`, `cross-session-tools.ts`, `analytics-tools.ts`, `extended-analytics-tools.ts`) only export bare constants and handler functions — none of them registers its own tools. A new feature area cannot add MCP tools without editing that one shared file's tools array, its `ToolRegistrationOptions` type (already 30 optional fields), and the switch.

Two smaller issues in the same file compound this:

- The "tracker not available" error-response block is duplicated near-verbatim ~40 times instead of being factored into a helper.
- Conditional tool registration is done via ~30 hand-written, undocumented `&&` combinations of tracker dependencies, with no declarative model — plus at least one dead-but-live case (a tool intentionally excluded from the registered list but still reachable through the switch with a hardcoded "not currently functional" response).

## Proposed fix

- Introduce a `registerXTools(server, deps)` pattern so each tool-surface file owns its own registration, composed by `session-stats.ts` rather than centralized there.
- Extract a shared `requireTracker(tracker, name)` helper for the not-available response.
- Encode tracker-dependency requirements as a small declarative table instead of inline `&&` chains.

## Affected files

`src/tools/session-stats.ts`, `cost-tools.ts`, `workflow-tools.ts`, `cross-session-tools.ts`, `analytics-tools.ts`, `extended-analytics-tools.ts`

chrisdhaan:
## Summary

MCP tool registration was centralized in one 1476-line file: a ~30-condition if-chain building the `tools/list` array, and a ~760-line `switch` dispatching `tools/call`, both referencing consts/handlers scattered across five sibling files that had no registration logic of their own.

- New `src/tools/tool-registry.ts` — a small `ToolSpec`/`buildToolSet`/`mergeToolSets` seam, plus `requireTracker`/`requireAvailable` helpers.
- Each sibling file (`cost-tools.ts`, `workflow-tools.ts`, `cross-session-tools.ts`, `analytics-tools.ts`, `extended-analytics-tools.ts`) now exports its own `registerXTools(deps): RegisteredToolSet` — a declarative array of its tools, their availability conditions, and their handlers.
- `session-stats.ts`'s `registerTools()` shrinks to composing all six sets (five sibling files + its own local `registerCoreTools()`) into the server's single `tools/list`/`tools/call` handler pair.
- The ~40 near-identical "tracker not available" error blocks are replaced by two shared helpers instead of copy-pasted per tool.
- `nr_observe_get_latency_decomposition` (permanently disabled since it has no real backing tracker today) gets an explicit comment marking it as intentionally off, distinct from "not wired yet."

No behavior change: every tool's `tools/list` availability condition and dispatch error-message text is preserved exactly.

## Test plan

- [x] `npm run build && npm test` — 4631 tests passing, 0 failures
- [x] `npm run lint` and `npm run format:check` — clean

Closes #243

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.9] - 2026-07-20
+
+### Changed
+
+- The MCP server's tool registration was centralized in one 1476-line file: a single function built the `tools/list` array from a ~30-condition if-chain and dispatched `tools/call` through a ~760-line switch, referencing tool definitions and handlers scattered across five sibling files that had no registration logic of their own. Each sibling file now exports its own `registerXTools()` — a small declarative list of its tools, their availability conditions, and their handlers — and the main registry composes all of them into the server's single `tools/list`/`tools/call` handler pair. No user-visible behavior change.
+- The ~40 near-identical "tracker not available" error blocks in that dispatch switch are now built from two shared helpers (`requireTracker`/`requireAvailable`) instead of being copy-pasted per tool.
+
 ## [1.6.8] - 2026-07-20
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.8",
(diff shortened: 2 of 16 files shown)


corinnekunze:

PR #253, opened 2026-07-20
Fix: Give proxy-mode tool calls a real per-connection session id

### LINKED ISSUE #244: Proxy mode collapses all concurrent clients into one session and skips the disk audit log
## Problem

Proxy mode is meant to multiplex many concurrent client sessions to shared upstream MCP servers, but several parts of the implementation only handle a single logical session:

- `ProxyManager.emitToolCallRecord()` hardcodes `sessionId: null` on every proxied tool call, and `main()`'s proxy branch computes one process-level `sessionTraceId` for the whole proxy's uptime. Two developers proxied through the same instance are indistinguishable in New Relic telemetry and in the audit trail.
- `ai.session.*` gauges (duration, unique files read/written) are emitted unconditionally every harvest tick, but the only code that populates them (`SessionTracker.recordToolCall()`) is never called from the proxy code path — so proxy mode reports a session whose duration grows for the life of the process with permanently-zero file activity.
- `AuditTrailManager`'s disk persistence is a no-op whenever it's constructed without a `LocalStore`, and the proxy branch of `main()` never passes one in — so the documented real-time on-disk audit log doesn't exist for anything proxied.
- `StdioUpstream` opens exactly one child-process connection per upstream and reuses it for every concurrent client, which would conflate sessions for any upstream MCP server that keeps per-connection state.

## Proposed fix

- Thread a real per-connection/per-request session id through `ProxyManager` instead of the hardcoded `null`.
- Feed proxied `ToolCallRecord`s into `SessionTracker` (or gate the session gauges by mode) so the metrics are accurate.
- Pass a `LocalStore` into the proxy branch's `NrIngestManager` construction, matching the stdio/local branch.
- Document (or fix) the single shared stdio-upstream connection limitation.

## Affected files

`src/index.ts`, `src/proxy/proxy-manager.ts`, `src/proxy/upstream-stdio.ts`, `src/proxy/types.ts`, `src/security/audit-trail.ts`, `src/transport/nr-ingest.ts`

chrisdhaan:
## Summary

- Proxy mode multiplexes many concurrent developers' MCP clients through one process, but collapsed all of them into a single fake session — every `ProxyToolCallRecord` hardcoded `sessionId: null`, and `AiMcpToolCall` events / per-tool metrics were tagged with one process-level `sessionTraceId` instead.
- `ProxyManager` now resolves a real per-client session id (the MCP `Mcp-Session-Id` request header when present, otherwise one generated per TCP connection via a `WeakMap<Socket, string>`) and threads it through `AiMcpToolCall` events and per-tool metrics.
- The `ai.session.*` gauges (previously reporting a never-populated, single-session `SessionTracker` in proxy mode — e.g. a week-long process reporting a multi-day "session duration" with zero file activity) are now gated off via a new `trackSessionGauges` option.
- Proxy mode now constructs an unscoped `LocalStore` (mirroring `--local` mode) so the on-disk security audit log is actually persisted for proxied traffic, instead of silently no-op'ing.
- `StdioUpstream`'s shared-connection-across-all-clients limitation is documented (class-level comment), not fixed with pooling — a larger design change explicitly descoped for this PR.
- Deliberately left the generic `AiToolCall` event's `session_id` unchanged (still `sessionTraceId`) — two pre-existing tests lock in that field as a cross-mode "single source of truth"; the proxy-specific `AiMcpToolCall` event and metrics carry the real per-connection id instead.

Closes #244

## Test plan

- [x] New tests in `proxy-manager.test.ts`: `mcp-session-id` header passthrough, per-connection fallback id generation, distinct ids across distinct connections.
- [x] New/updated tests in `nr-ingest.test.ts`: `proxyToolCallToNrEvent` prefers `record.sessionId`; `ingestToolCall` scopes the metric-dimension fix to proxy records only (regression guard confirms stdio/local unaffected); `trackSessionGauges: false` suppresses the three `ai.session.*` gauges.
- [x] `npm run build && npm test` — full suite green (4638 passed, 3 pre-existing skips).
- [x] `npm run lint` — no new lint issues (1 error/5 warnings, all pre-existing in unrelated `scripts/*.ts` files).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.10] - 2026-07-20
+
+### Fixed
+
+- Proxy mode assigned every concurrent client the same fake shared session id (`sessionId: null` on every proxy tool call, one process-level id reused across all developers). Each proxy client's `AiMcpToolCall` events and per-tool metrics now carry a distinct per-connection session id, resolved from the MCP `Mcp-Session-Id` request header when present and otherwise generated per TCP connection.
+- Proxy mode emitted `ai.session.duration_ms`/`unique_files_read`/`unique_files_written` gauges reflecting a single never-populated session tracker (e.g. a week-long proxy process reporting a multi-day session duration with zero file activity). These gauges are no longer emitted in proxy mode.
+- The on-disk security audit log was never populated in proxy mode because no local store was wired up for it; proxied tool calls are now persisted to the audit log like stdio/local traffic.
+- Documented that the proxy's stdio upstream transport shares one child-process connection across all concurrent clients — safe for stateless MCP servers, not for stateful ones.
+
 ## [1.6.9] - 2026-07-20
 
 ### Changed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.9",
+  "version": "1.6.10",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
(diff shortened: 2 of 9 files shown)


corinnekunze:

PR #255, opened 2026-07-20
Fix: Decouple maintenance GC from dashboard-bind, add retention defaults and file-tracker caps

### LINKED ISSUE #247: Session/buffer retention is off by default and doesn't run on a schedule
## Problem

Several storage cleanup paths don't run under normal conditions:

- `retainSessionsDays` defaults to `null` (retention disabled), and even when configured, `purgeOldSessions()` only runs once at process startup — never on a recurring interval — and never touches `weekly_summaries/` at all. A long-running installation accumulates session JSON files indefinitely.
- Orphan buffer-file garbage collection (`LocalStore.gcOrphanBuffers()`) only runs when the `--local` dashboard successfully binds its HTTP port. The config loader's default mode is `'cloud'` — the standard stdio-MCP setup — so in the default deployment, any abnormally-terminated session (crash, `kill -9`, terminal close) leaves its buffer file and heartbeat file on disk forever.
- `ContextWindowTracker.fileReadCounts` and `SessionTracker.filesRead`/`filesWritten` are uncapped Maps/Sets, unlike every sibling in-memory structure in the same files (which are explicitly capped with a comment explaining why). A session that touches many thousands of distinct files (a large repo-wide operation) grows these structures without bound.

## Proposed fix

- Default `retainSessionsDays` to a sane non-null value and re-run the purge on a periodic interval; extend it to sweep `weekly_summaries/` too.
- Decouple orphan-buffer GC from dashboard-bind success so it runs regardless of `mode`.
- Cap the two uncapped per-session structures with the same head-drop/LRU eviction style already used elsewhere in the tracker layer.

## Affected files

`src/storage/retention.ts`, `src/storage/local-store.ts`, `src/storage/weekly-summary.ts`, `src/metrics/context-window-tracker.ts`, `src/metrics/session-tracker.ts`, `src/index.ts`, `src/config.ts`

chrisdhaan:
## Summary

- Retention (`retainSessionsDays`) now defaults to 90 days instead of being off by default (an explicit `null` in config.json still disables it — a deliberate opt-out, distinct from the key being absent). The purge now also sweeps `weekly_summaries/`, not just `sessions/`, and re-runs every 6 hours instead of once at startup.
- Orphan-buffer/breadcrumb/dead-instance GC previously only ran when the `--local` dashboard won its port bind, but the default `mode` is `cloud` (the documented stdio-MCP setup) — a crashed session in the default configuration leaked `buffer-<id>.jsonl` and heartbeat files forever. GC now runs unconditionally in every mode. `gcOrphanBuffers()` is rename-based and already tolerates concurrent execution across processes (a second attempt on an already-moved file just logs a benign warning), so no new leader-election primitive was needed.
- `ContextWindowTracker.fileReadCounts` and `SessionTracker.filesRead`/`filesWritten` were uncapped, unlike every sibling capped structure in the same files. Both now cap at 10,000 unique file paths per session with insertion-order head-drop eviction.

As a consequence of decoupling GC from the dashboard-bind lifecycle, `setupDashboardPostBind()` shrank to just PID-file-writing + the `openOnStart` warning, and `startDashboardRepoll()`'s now-unnecessary `onTakeover` callback was removed.

Closes #247

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — full suite green (4666 passed, pre-existing unrelated skips)
- [x] `npm run lint` — 0 new errors/warnings (1 pre-existing error + 5 warnings, both in `scripts/`, confirmed present on `main` before this branch)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.12] - 2026-07-20
+
+### Added
+
+- `retainSessionsDays` now defaults to 90 days instead of retention being off by default; set it to `null` explicitly in config.json to disable retention.
+- Retention now also sweeps `weekly_summaries/`, not just `sessions/`, and re-runs every 6 hours instead of once at startup.
+- `ContextWindowTracker.fileReadCounts` and `SessionTracker.filesRead`/`filesWritten` are now capped at 10,000 unique file paths per session, evicting the oldest-touched file first — previously unbounded, unlike every sibling capped structure in the same files.
+
+### Fixed
+
+- Orphan-buffer/breadcrumb/dead-instance GC only ran when the `--local` dashboard won its port bind, but the default `mode` is `cloud` (the documented stdio-MCP setup) — a crashed session in the default configuration leaked `buffer-<id>.jsonl` and heartbeat files forever. GC now runs unconditionally in every mode, independent of dashboard-bind status.
+
 ## [1.6.11] - 2026-07-20
 
 ### Added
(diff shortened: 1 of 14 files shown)


corinnekunze:

PR #256, opened 2026-07-21
Chore: Sync v1.6.13 — config validation gaps and nested OTLP fields

### LINKED ISSUE #248: Numeric config bounds are only enforced on the env-var path, not CLI/config-file values
## Problem

`envInt(key, defaultValue, bounds)` only clamps a value read from an environment variable — when the env var is unset, it returns `defaultValue` unchecked. Several call sites pass a config-file-derived or CLI-derived value as that `defaultValue` (`harvestEventsMs`, `harvestMetricsMs`, `port`, `otlpReceiverPort`), so a typo'd config file value (e.g. `harvestEventsMs: 1`) or a stray CLI flag (`--port -1`) reaches the harvest scheduler or HTTP server completely unclamped, bypassing the documented min/max bounds entirely.

Related: budget and retention fields (`sessionBudgetUsd`, `dailyBudgetUsd`, `weeklyBudgetUsd`, `retainSessionsDays`) validate positivity only on the env-var path with no logging on rejection — a negative value in the config file passes schema validation untouched and silently disables the cost guardrail, with no warning, unlike other malformed-input paths in the same file which all call `logger.warn`.

Two smaller structural issues in the same file: the 8 OTLP-related config fields are flat top-level fields while `dashboard`/`alerts` (added around the same time) are nested objects — two competing precedents for how new config should be added — and there's no `configVersion` field or migration mechanism, which has only worked so far because every config change has been additive.

## Proposed fix

- Apply bounds-clamping uniformly regardless of source (env/CLI/file), not just the env-var path.
- Add the same positivity check plus a `logger.warn` on rejection for budget/retention fields sourced from the config file.
- Group the 8 OTLP fields into a nested `otlp: {...}` object, matching the `dashboard`/`alerts` precedent.
- Add an optional `configVersion` field now, before any non-additive config change ships.

## Affected files

`src/config.ts`

chrisdhaan:
## Summary

Fixes 4 config-validation/structure gaps:

- Numeric config values sourced from a config file or CLI flag (harvest intervals, the local OTLP receiver port, the MCP server port, alert evaluation interval/log retention) bypassed the same min/max bounds already enforced on the equivalent environment variable. A `clampNumber()` helper is now applied uniformly regardless of source.
- `sessionBudgetUsd`/`dailyBudgetUsd`/`weeklyBudgetUsd`/`retainSessionsDays` validated positivity only on the env-var path with zero logging on rejection — a negative value in config.json silently disabled the guardrail. A `validateFilePositiveNumber()` helper now validates the file-sourced branch too and logs a warning on rejection.
- The 8 OTLP-related config fields were flat top-level config while `dashboard`/`alerts` were nested objects. They're now grouped under a nested `otlp: {...}` object on both the config-file schema and the resolved server config. The config-file schema still accepts the legacy flat keys for backward compatibility (non-breaking, accept-both) — using one now logs a one-time deprecation warning naming exactly the legacy keys consulted.
- Added an optional `configVersion` field (defaults to 1) with a documented convention for when to bump it.

## Test plan

- [x] `npm run build && npm test` — full suite green (4682 passed, 3 pre-existing skips)
- [x] `npm run lint` — clean (matches pre-existing baseline, unrelated to this change)
- [x] `package-lock.json` regenerated locally via `npm install --package-lock-only`

Closes #248


corinnekunze:

PR #257, opened 2026-07-21
Fix: SubagentWatcher UTF-8 boundary decode + dashboard-health test coverage

### LINKED ISSUE #249: SubagentWatcher's fixed-size read can split multi-byte UTF-8 at the poll boundary
## Problem

`SubagentWatcher.processFile()` reads a fixed 64KB window per poll and immediately calls `.toString('utf-8')` on it before any newline-based line splitting. Real transcripts routinely contain multi-byte UTF-8 (emoji, non-ASCII text, box-drawing characters). When the 64KB cutoff lands in the middle of such a character, the bytes are decoded with a replacement character and the read cursor advances past them — the corruption is permanent for that line, since the split bytes are never re-read.

If the split falls inside a JSON string value used for dedup/drift-detection fields, the parsed value silently contains replacement characters; if it falls at a structurally sensitive spot, the line fails to parse and is indistinguishable from any other malformed line. This can corrupt or drop subagent cost/token attribution data on long-running or high-throughput sessions with non-ASCII content, and there's no test exercising this path.

Separately, `src/install/dashboard-health.ts` — which drives the pass/fail signal shown to users after `preflight update` and the setup wizard — has zero test coverage despite being a maintained, logic-bearing file.

## Proposed fix

- Decode with a boundary-safe approach (e.g. `node:string_decoder`'s `StringDecoder`, or track incomplete tail bytes across polls) and add tests for a multi-byte character straddling the read boundary.
- Add `dashboard-health.test.ts` covering the healthy/mismatch/timeout/connection-refused-then-recovers cases with a mocked `fetch`.

## Affected files

`src/hooks/subagent-watcher.ts`, `src/install/dashboard-health.ts`

chrisdhaan:
## Summary

- `SubagentWatcher.processFile()` decoded each 64KB poll read with a one-shot `toString('utf-8')`. A multi-byte character (emoji, non-ASCII) whose bytes straddled the read boundary got silently replaced with U+FFFD and permanently lost, since the byte cursor had already advanced past those bytes — a real risk to subagent cost/token attribution. Decoding now goes through a per-file `StringDecoder` (`node:string_decoder`), which buffers an incomplete trailing sequence and reconstructs it on the next poll.
- `dashboard-health.ts` had zero test coverage despite gating the pass/fail signal `preflight update` and the setup wizard's daemon-install flow show the user. Added `dashboard-health.test.ts` covering `waitForHealthyDashboard()` (healthy match, version-mismatch timeout, connection-refused-then-recovers, always-fails timeout, `expectedVersion=null` short-circuit) and `getDashboardAddress()` (success and config-load-throws branches).

Version bumped 1.6.13 → 1.6.14 (the decode fix changes shipped runtime behavior).

Closes #249

## Test plan

- [x] New regression test in `subagent-watcher.test.ts` computes the exact byte offset needed to split a 4-byte emoji across the 64KB poll boundary and asserts the reconstructed `messageId` is byte-for-byte correct
- [x] New `dashboard-health.test.ts` (7 cases) — all pass
- [x] Full suite: `npm run build && npm test && npm run lint && npm run format:check` — 4690 tests pass, lint/format clean (pre-existing `scripts/` lint baseline unchanged)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.14] - 2026-07-21
+
+### Fixed
+
+- `SubagentWatcher` decoded each 64KB poll read with a one-shot `toString('utf-8')`; a multi-byte character (emoji, non-ASCII) whose bytes straddled the read boundary was silently replaced with the Unicode replacement character and permanently lost, since the byte cursor had already advanced past those bytes. Decoding now goes through a per-file `StringDecoder`, which buffers an incomplete trailing byte sequence and reconstructs the character correctly on the next poll.
+
+### Added
+
+- Test coverage for `dashboard-health.ts` (`waitForHealthyDashboard()` and `getDashboardAddress()`), which had none despite gating the pass/fail signal shown by `preflight update` and the setup wizard's daemon-install flow.
+
 ## [1.6.13] - 2026-07-20
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.6.13",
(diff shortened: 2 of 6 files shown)


corinnekunze:

PR #264, opened 2026-07-21
Fix: Thread validateSsrfUrl into key-validator.ts and cross-session-tools.ts

### LINKED ISSUE #265: SSRF validation gap in key-validator.ts and cross-session-tools.ts outbound fetch
## Summary

Two outbound-`fetch()` dispatch sites don't run their target URL through the SSRF validation (`validateSsrfUrl`, `src/security/ssrf.ts`) already applied to the proxy's outbound HTTP dispatch (`src/proxy/upstream-http.ts`, `src/proxy/otlp-receiver.ts`):

- `src/install/key-validator.ts` — `validateLicenseKey()` and `validateApiKey()` fetch a URL built from a fixed map of NR hostnames with no SSRF check.
- `src/tools/cross-session-tools.ts` — `handleGetTeamSummary()`'s NerdGraph query (`runNrql()`) does the same.

## Why this matters

Both dispatch sites build their URL from a small hardcoded map of real NR hostnames with a safe fallback, so this isn't a currently exploitable path — but it's an inconsistency against the codebase's own invariant that every outbound HTTP call passes through SSRF validation before dispatch.

## Fix

Thread `validateSsrfUrl` into both call sites, matching the existing pattern in `proxy/upstream-http.ts`.

chrisdhaan:
Fixes #265

## Summary

- License-key/API-key validation and the team-summary NerdGraph query now run their outbound URL through the same SSRF validation (`validateSsrfUrl`) already applied to the proxy's outbound HTTP dispatch.
- Both call sites build their URL from a fixed map of real NR hostnames with a safe fallback, so this closes a defense-in-depth/consistency gap rather than a currently exploitable path — no behavior change for legitimate calls.

## Test plan

- [x] Added tests asserting `fetch` is never called when SSRF validation rejects the target URL, for both `key-validator.ts` and `cross-session-tools.ts`.
- [x] `npm run build && npm test` — full suite passes.
- [x] `npm run lint` — no new lint issues introduced.
- [x] Version bumped to 1.6.18 and CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.18] - 2026-07-21
+
+### Fixed
+
+- License-key/API-key validation (`install/key-validator.ts`) and the team-summary NerdGraph query (`tools/cross-session-tools.ts`) now run their outbound URL through the same SSRF validation (`validateSsrfUrl`) already applied to the proxy's outbound HTTP dispatch. Both build their URL from a fixed map of real NR hostnames, so this closes a consistency gap rather than a currently exploitable path.
+
 ## [1.6.17] - 2026-07-21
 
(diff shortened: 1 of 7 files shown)


corinnekunze:

PR #267, opened 2026-07-22
Fix: Drop dead sessionId field from client-side AlertEvent type

### LINKED ISSUE #232: AlertEvent type diverges between src/web/api/client.ts and src/web/store/liveStore.ts
## What

Two client-side copies of the `AlertEvent` type exist:

- `src/web/api/client.ts` — `AlertEvent` includes an optional `sessionId` field.
- `src/web/store/liveStore.ts` — `AlertEvent` does not have `sessionId`. This is intentional per the comment on that type (alerts aren't session-scoped client-side — every firing alert is shown regardless of which session triggered it), verified against the server-side type in `src/dashboard/live-event-bus.ts`.

The two copies of the same conceptual shape have diverged, which could surprise a future reader who diffs one against the other expecting them to match.

## Suggested next step

Either drop the now-effectively-unused `sessionId` from `client.ts`'s `AlertEvent` if nothing actually needs it, or add a short comment there explaining why it's retained despite `liveStore.ts` not needing it (e.g. if some other consumer of `client.ts`'s type relies on it).

chrisdhaan:
## Summary
- `src/web/api/client.ts`'s `AlertEvent` type declared an optional `sessionId` field mirroring the server-side `AlertEvent` shape (`src/dashboard/live-event-bus.ts`), but `AlertLog`'s runtime schema (`src/alerts/alert-log.ts`, `AlertEventSchema`) never included `sessionId` — Zod's default object parsing strips unknown keys, so `/api/alerts/recent` never actually returns it.
- Verified the only consumer of this type (`RecentAlertsPanel` in `src/web/views/Today.tsx`) never reads `.sessionId`, so no behavior changes.
- `src/web/store/liveStore.ts` already has its own local `AlertEvent` (used for the SSE-driven live-alert store) correctly omitting `sessionId`, with a comment explaining why. Added an analogous comment to `client.ts`'s copy.

Fixes #232

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.web.json` — clean
- [x] `npm test` — 4709 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.6.20] - 2026-07-21
+
+### Fixed
+
+- The client-side `AlertEvent` type in `src/web/api/client.ts` declared an optional `sessionId` field mirroring the server-side `AlertEvent` shape, but `AlertLog`'s runtime schema (`src/alerts/alert-log.ts`) never included `sessionId`, so `/api/alerts/recent` never actually returned it. Removed the dead field; `src/web/store/liveStore.ts` already had its own `AlertEvent` correctly omitting it, with a comment explaining why.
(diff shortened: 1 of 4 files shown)


corinnekunze:

PR #278, opened 2026-07-23
updated privacy.md with recommendations for 'ai.collaboration.*'

jjang-nr:


chrisdhaan:
Thanks for raising this — the underlying instinct here is a good one: behavioral characterization of individual developers is a distinct, more sensitive category than the rest of this project's telemetry, and it deserved its own explicit call-out in `PRIVACY.md`.

While reviewing this PR, I traced the actual code and found that the premise it (and, it turns out, `PRIVACY.md` itself, in two places predating this PR) was built on doesn't hold: `ai.collaboration.*` and the classification label are never actually transmitted to New Relic, in any mode. `CollaborationProfiler.emitMetrics()` is never called anywhere in the running server — the real harvest loop only emits a few other trackers — and every real use of this data only backs local MCP tools. This data stays entirely on your machine.

Since `PRIVACY.md` already claimed otherwise before this PR, the fix needed to go a bit further than an edit here: #289 corrects both pre-existing inaccurate statements and adds an accurate description of this data in the "Local Only" section, while keeping the part of your contribution that's still worth preserving — that this category deserves its own explicit mention.

Closing this in favor of #289, which has that fix. Thanks again for catching that this needed documentation at all — appreciate the contribution.

PR #282, opened 2026-07-24
Fix: native-Windows --stdio MCP never resolves session_id

### LINKED ISSUE #268: Native Windows: --stdio MCP never resolves session_id (Linux-only ancestor-PID breadcrumb walk via Git Bash) -> sessions/ and History always empty
## Summary

On **native Windows Claude Code** (not WSL), the dashboard History / `sessions/` is permanently empty and `nr_observe_health` returns `session_id: null`, even though hooks fire correctly and the live "current session" view works. No session summary is ever persisted, so History, weekly summaries, and MODEL PERFORMANCE never populate — for every native-Windows user.

Root cause: the `--stdio` MCP server can never resolve its owning Claude Code session id, because the breadcrumb ancestor-PID walk that bridges an interposed shell is **Linux-only** (`/proc`-based), while native-Windows Claude Code always interposes Git Bash between itself and the hook collector.

## Root cause

Session persistence only happens for a **real** (non-synthetic) session id. `persistSession()` in `src/index.ts` skips synthetic ids:

```js
// hooks/session-resolver.js
export function isSyntheticSessionId(id) {
  return id.startsWith('local-') || id.startsWith('proxy-') || id.startsWith('pending-');
}
```

The `--stdio` MCP resolves its real id from a breadcrumb keyed by **its own `process.ppid`**:

```js
// hooks/session-resolver.js
export function resolveFromBreadcrumb(storagePath, ppid) {
  const breadcrumbPath = resolve(storagePath, 'session-by-ppid', `${ppid}.txt`);
  if (!existsSync(breadcrumbPath)) return null;
  ...
}
```

The collector writes that breadcrumb. To cover the case where a shell is interposed between Claude Code and the collector (so the collector's ppid ≠ the MCP's ppid), `writePpidBreadcrumb()` walks the ancestor PIDs and writes a breadcrumb at every one:

```js
// hooks/collector-script.js
const pids = getLinuxAncestorPids(ppid);   // <-- Linux-only
for (const pid of pids) { writeFileSync(resolve(breadcrumbDir, `${pid}.txt`), sessionId, ...); }
```

But `getLinuxAncestorPids()` reads `/proc/<pid>/stat`, which does not exist on Windows, so it immediately falls into its `catch` and returns **only the direct ppid**:

```js
export function getLinuxAncestorPids(startPpid, maxDepth = 5) {
  const pids = [startPpid];
  ...
  const stat = _procFs.readFile(`/proc/${pid}/stat`);  // throws on win32 -> break
  ...
  return pids;   // on Windows: [startPpid] only
}
```

The code comment even states the ancestor walk is what makes interposed-shell setups work, and that "On macOS/Windows the array has exactly one element" — but **native-Windows Claude Code runs hooks through Git Bash**, i.e. it is exactly the interposed-shell case the walk was designed to fix, on the one platform where the walk is a no-op.

So on native Windows the chain is:

- Hook runs as `claude.exe → /usr/bin/bash → node.exe collector-script.js`, so the collector's `process.ppid` is a **transient Git Bash pid** (different on every hook invocation).
- The collector writes only `session-by-ppid/<transient-bash-pid>.txt`.
- The `--stdio` MCP's `process.ppid` is **`claude.exe`'s pid**, which never has a breadcrumb.
- `resolveFromBreadcrumb()` returns null → the MCP stays on a synthetic `pending-<ts>` id → `persistSession()` skips it → nothing is ever written to `sessions/`.

## Evidence (my machine)

- `@newrelic/preflight` **1.6.18**, native Windows install (`platformTarget: native`), Claude Code **2.1.217**, Node v22.23.1, Windows 11. Hooks executed via Git Bash.
- `nr_observe_health` → `"session_id": null` (MCP up 44 min, never resolved).
- `~/.newrelic-preflight/sessions/` is **empty** across 8 days / 11+ sessions; every `weekly_summaries/*.json` has `sessionCount: 0`.
- Hooks ARE firing: `session-by-ppid/` contains many breadcrumbs, all with the correct session UUID — but keyed by transient Git Bash pids (`1540.txt`, `20984.txt`, `26444.txt`, …).
- Process tree confirmed: the `--stdio` MCP (pid 29020) has parent pid **25048 = `claude.exe`**; there is **no `session-by-ppid/25048.txt`**, so the lookup misses.
- `CLAUDE_JOB_DIR` is unset, so the alternate `resolveFromJobDir()` path doesn't help either.
- The live `/api/session/current` view still works only because a separate `--local` dashboard ingests buffers in memory under a synthetic `local-<ts>` id — which is likewise never persisted.

## Repro

1. Native Windows Claude Code (desktop), Preflight installed natively (`preflight install`), hooks run via Git Bash.
2. Use Claude Code normally for several minutes / several sessions.
3. Call `nr_observe_health` → `session_id: null`; check `~/.newrelic-preflight/sessions/` → empty; open the dashboard History → empty.

## Suggested fix

Give `getLinuxAncestorPids()` a Windows code path so the collector can write a breadcrumb at `claude.exe`'s pid (the MCP's ppid). Options:

1. On win32, walk the ancestor chain via `Win32_Process.ParentProcessId` (e.g. a cached `wmic`/PowerShell/`CreateToolhelp32Snapshot` lookup) instead of `/proc`. Keep the hot-path budget small by only walking on the PreToolUse that also writes the breadcrumb.
2. Or resolve the session id without the ppid bridge on Windows — e.g. have the collector drop a breadcrumb keyed by something the MCP can independently derive (the transcript/cwd project dir it already knows), and add a `resolveFromCwd()` fallback alongside `resolveFromBreadcrumb()` / `resolveFromJobDir()`.

Either way, a once-only stderr/`logger.warn` when the MCP falls back to a synthetic id after N seconds would have made this diagnosable immediately instead of looking like a healthy-but-empty dashboard.

Related but distinct: #17 covers the WSL2 / Linux-CC resolution gap; this is the native-Windows (Git-Bash-interposed) variant, which the Linux-only ancestor walk does not cover.

## Environment

- @newrelic/preflight 1.6.18 (same code on current `main`)
- Windows 11, native Claude Code 2.1.217 (hooks executed via Git Bash)
- Node v22.23.1 (native Windows npm global install), `platformTarget: native`, local mode

chrisdhaan:
## Summary

- Native Windows Claude Code sessions (hooks running through Git Bash) never resolved a real `session_id` for the `--stdio` MCP server — the hook collector's PID-keyed breadcrumb never matched the MCP server's own PID, so affected sessions stayed on a synthetic `pending-` ID forever. `sessions/`, dashboard History, weekly summaries, and MODEL PERFORMANCE never populated for those sessions.
- Adds a second breadcrumb, keyed by the sanitized project working directory instead of a PID, written by the hook collector (`writeCwdBreadcrumb`) alongside the existing PID-keyed one. The session resolver (`resolveFromCwd`) falls back to it, on any platform, only when the precise PID-based lookup misses — no behavior change for platforms where PID resolution already works.
- Also fixes a follow-up issue found during review: the cwd-sanitization scheme didn't strip the Windows drive-letter colon (`C:`), which `path.resolve()` can misinterpret as a drive-relative path segment — cross-drive setups (project and storage directory on different drives) could resolve the breadcrumb path outside the intended, permission-secured storage directory. Fixed by stripping `:` too, with regression tests covering the exact scenario.
- Version bumped 1.14.0 → 1.14.1 with a CHANGELOG entry.

Fixes #268

## Test plan

- [x] New unit tests for `writeCwdBreadcrumb()` and `resolveFromCwd()` / fallback-ordering / native-Windows scenario
- [x] Regression tests for the Windows drive-letter colon fix, verified against real `path.win32.resolve` behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.1] - 2026-07-24
+
+### Fixed
+
+- **Native Windows session resolution** — the `--stdio` MCP server now resolves its Claude Code `session_id` on native Windows, where Claude Code interposes Git Bash between itself and the hook collector but launches the MCP server directly. The hook collector writes an additional breadcrumb keyed by the project's working directory (alongside the existing PID-keyed breadcrumb); the MCP server falls back to it, on any platform, whenever the PID-based lookup misses. Previously, affected sessions never resolved a real session ID, so dashboard History, weekly summaries, and model-performance metrics never populated for them.
+
 ## [1.14.0] - 2026-07-23
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.0",
+  "version": "1.14.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.14.0",
+      "version": "1.14.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.0",
+  "version": "1.14.1",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/hooks/collector-script.test.ts
+++ b/src/hooks/collector-script.test.ts
@@ -16,6 +16,7 @@ import {
   translateWslPath,
   getBufferPath,
   writePpidBreadcrumb,
+  writeCwdBreadcrumb,
   getLinuxAncestorPids,
   _procFs,
   readStdinSync,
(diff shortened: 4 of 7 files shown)


corinnekunze:

PR #284, opened 2026-07-24
Fix: GC .subagent-pos-*/.transcript-pos-* cursor files

### LINKED ISSUE #281: Housekeeping: .subagent-pos-* / .transcript-pos-* cursor files are never GC'd (329 files from one session; both sweeps skip them)
## Summary

Cursor/checkpoint files written into the storage root — `.subagent-pos-<parentSessionId>-<agentId>` (SubagentWatcher) and `.transcript-pos-<sessionId>` (collector) — are **never garbage-collected**. They accumulate for the lifetime of the install, and the `.subagent-pos-*` count scales with subagent usage rather than with session count.

This is minor (the files are ~2.6 KB each) and platform-independent — filing it as housekeeping, not a functional bug.

## Neither GC path covers them

`runMaintenanceGcPass()` cleans three things, none of which match these files:

```js
localStore.gcStaleBreadcrumbs();      // session-by-ppid/*.txt
localStore.gcDeadLocalInstances();    // local-instances/*.json
localStore.gcOrphanBuffers(live);     // buffer-<sessionId>.jsonl only
```

`gcOrphanBuffers()` is explicitly prefix/suffix-scoped and skips everything else:

```js
if (!name.startsWith('buffer-') || !name.endsWith('.jsonl')) continue;
```

`runRetentionSweep()` only purges persisted summaries:

```js
const deletedSessionFiles       = purgeOldSessions(deps.storagePath, deps.retainSessionsDays);
const deletedWeeklySummaryFiles = purgeOldWeeklySummaries(deps.storagePath, deps.retainSessionsDays);
```

So `.subagent-pos-*` / `.transcript-pos-*` fall through both sweeps permanently.

## Observed on my install (1.6.18)

```
.subagent-pos-*   329 files, 854 KB   <- all from ONE parent session, created 18:37-21:00 (~2.5h)
.transcript-pos-*  17 files            <- one per session ever run
storage root total 3.4 MB
```

The notable part is the scaling: **329 cursor files from a single 2.5-hour session**, because one file is written per subagent (`agent-{id}`), including workflow-spawned agents. A few heavy subagent/workflow days would put five figures of small files in one flat directory.

Minor knock-on effect: `gcOrphanBuffers()` does `readdirSync(this.storagePath)` on every 5-minute GC pass, so the directory it enumerates keeps growing with entries that sweep can never remove.

## Why pruning them is safe

The watcher's own discovery budget already makes old cursors provably inert — transcripts whose mtime is outside the discovery window are not cold-scanned at all:

> Startup-discovery budget: only files with mtime in the last 24h are eligible for cold scan (configurable via `NR_AI_WATCHER_DISCOVERY_HOURS`)

And even if a cursor is dropped for a file that *is* still in the window, the documented behavior is a harmless re-read:

> On crash mid-emit the next poll re-reads from the previous cursor → potential duplicates which downstream dedupes by (agent_id, message.id)

So deleting a cursor costs at most one re-read that dedup already absorbs — there is no correctness risk in collecting them.

## Suggested fix

Extend the maintenance GC with a cursor sweep, using the same liveness input it already computes:

1. `.subagent-pos-<parentSessionId>-<agentId>` — remove when `parentSessionId` is not in the live set **and** mtime is older than `NR_AI_WATCHER_DISCOVERY_HOURS` (default 24h). Past that window the watcher will never cold-scan the corresponding transcript, so the cursor is dead state by construction.
2. `.transcript-pos-<sessionId>` — remove when the session is not live and has no remaining `buffer-<sessionId>.jsonl`.

Alternatively fold both into `runRetentionSweep()` next to `purgeOldSessions()` / `purgeOldWeeklySummaries()` so they inherit `retainSessionsDays`.

Either way it would also be worth having `preflight doctor` report the storage-root file count, since this is otherwise invisible until someone happens to look in the directory.

## Environment

- @newrelic/preflight 1.6.18 (code path unchanged on current `main`)
- Not platform-specific — the affected code is in `runMaintenanceGcPass()` / `LocalStore` / `runRetentionSweep()`
- Observed on Windows 11, native install, `platformTarget: native`, local mode

chrisdhaan:
## Summary
- Adds `LocalStore.gcWatcherCursors()`, which garbage-collects `.subagent-pos-<parentSessionId>-<agentId>` and `.transcript-pos-<sessionId>` cursor files in the storage root — previously neither the periodic maintenance sweep nor the retention sweep touched them, so they accumulated forever (one `.subagent-pos-*` file per spawned subagent, including workflow-spawned agents).
- Wires the new method into the existing 5-minute maintenance GC pass, reusing the same live-session-id set already built for the sibling orphan-buffer GC.
- Also fixes a pre-existing test-isolation gap: the first `stdio integration` test in `src/index.test.ts` spawned the real compiled binary without a storage-path override, which meant every `npm test` run performed real GC operations against the developer's actual local storage directory. Isolated with the same temp-dir override pattern already used by the neighboring test.

Closes #281

## Test plan
- [x] `npm run build && npm test` passes
- [x] New `gcWatcherCursors()` unit tests cover: live session preserved regardless of mtime; dead session + recent mtime preserved; dead session + old mtime deleted; transcript cursor preserved while its buffer exists; transcript cursor deleted once buffer is gone; transcript cursor preserved while its session is live even with no buffer; malformed filenames ignored; missing storage dir returns zeros
- [x] Confirmed the real local storage directory's file count is unchanged before/after `npm test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.3] - 2026-07-24
+
+### Fixed
+
+- **Subagent and transcript cursor files were never garbage-collected** — `.subagent-pos-<sessionId>-<agentId>` (written once per spawned subagent, including workflow-spawned agents) and `.transcript-pos-<sessionId>` files in the storage root accumulated indefinitely; heavy subagent or workflow usage could leave thousands of small files behind with no cleanup path. The periodic maintenance sweep now removes a subagent cursor once its parent session is no longer active and the cursor is older than the watcher's cold-scan window, and removes a transcript cursor once its session is no longer active and no companion buffer file remains for it.
+
 ## [1.14.2] - 2026-07-24
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.2",
+  "version": "1.14.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.14.2",
+      "version": "1.14.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.29.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.2",
+  "version": "1.14.3",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -758,6 +758,10 @@ describe('stdio integration', () => {
       JSON.stringify({ linkScanPath: '/tmp/stdio-test-session.jsonl' }),
     );
 
+    // Isolate storage to a temp directory to prevent GC operations from
+    // affecting the real ~/.newrelic-preflight storage.
+    const tmpStoragePath = mkdtempSync(join(tmpdir(), 'nr-stdio-storage-'));
+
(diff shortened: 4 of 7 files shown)


corinnekunze:

PR #296, opened 2026-07-29
Fix: Reject SSRF-vulnerable hosts on outbound NR ingest requests

### LINKED ISSUE #297: Outbound NR ingest requests aren't validated against SSRF
Outbound HTTP requests to the configured New Relic ingest endpoint (`sendWithRetry` in the shared HTTP transport) are sent without validating the destination host first. The OTLP transport already validates its endpoint against private/loopback and cloud-metadata hosts, but the New Relic ingest path does not.

If `collectorHost` (or another config value that feeds into the ingest URL) resolves to a private-network address or a cloud-metadata service endpoint, the request would be sent as-is — a classic SSRF surface for a misconfigured or attacker-controlled config value.

**Fix:** apply the same SSRF guard used for OTLP endpoint validation to the New Relic ingest send path, rejecting cloud-metadata and private/loopback hosts before the request goes out.

chrisdhaan:
Fixes #297

## Summary
- Adds an SSRF guard applied to outbound New Relic ingest requests (previously only OTLP endpoints had one) — rejects cloud-metadata and private/loopback hosts before sending; plaintext `http://` to a private host now throws instead of only warning
- Unifies the `highSecurity` → `recordContent` gate behind a single `resolveRecordContent()` helper (no behavior change), and updates `BatchLogRecordProcessor` construction for the current OpenTelemetry Logs SDK constructor shape
- The SDK update unblocks the OTel logs/exporter family bump (`sdk-logs`, `exporter-logs-otlp-http`, `exporter-metrics-otlp-http`, `exporter-trace-otlp-http`: 0.219.0 → 0.221.0) that was previously deferred
- Bumps package version to 1.14.5 with a CHANGELOG entry

## Test plan
- [x] `npm run build` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — 158/159 suites, 5156 tests passing (1 suite skipped, expected)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.5] - 2026-07-29
+
+### Security
+
+- Outbound HTTP requests to the configured New Relic ingest endpoint are now validated against the same SSRF guard used for OTLP endpoints, rejecting cloud-metadata and private/loopback hosts before sending.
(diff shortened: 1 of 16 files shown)


corinnekunze:

PR #302, opened 2026-07-30
Fix: Subagent spend aggregation and disabled-watcher banner

### LINKED ISSUE #303: Subagent spend on the Today dashboard undercounts across concurrent sessions
## Summary

The "subagent spend" figure on the Today dashboard doesn't reliably reflect actual subagent/workflow activity when more than one Claude Code session is running at the same time, and a related banner can misattribute the cause when it shows zero.

## Details

- The dashboard's aggregate spend endpoint only read subagent cost from the single process currently serving the dashboard's HTTP requests. With multiple concurrent sessions (a common pattern when using subagents/workflows), only one of those sessions' processes can hold the dashboard port at a time — so subagent activity from every other session was silently excluded from the total.
- When subagent cost tracking showed as inactive, the dashboard displayed a banner blaming `NR_AI_ENABLE_SUBAGENT_WATCHER=0` — even in cases where that variable was never set and the real reason was that a background/standalone dashboard process doesn't run its own subagent watcher by default.
- Opting a background dashboard process into its own watcher (`NR_AI_WATCHER_MODE=local`) could cause subagent spend it discovered to be double-counted against spend already recorded by the session it originated from.

## Expected behavior

Subagent spend on the Today dashboard should reflect activity across all sessions, and any banner explaining why subagent tracking looks disabled should name the actual cause.

chrisdhaan:
Fixes #303

## Summary
- Sum subagent spend from every persisted session's own total on the Today dashboard's aggregate endpoint, instead of reading only the single process currently serving the dashboard — fixes subagent spend silently excluding activity from other concurrent sessions.
- Distinguish, in the "subagent cost tracking is disabled" banner, between the watcher being explicitly disabled (`NR_AI_ENABLE_SUBAGENT_WATCHER=0`) and a background dashboard process simply not running its own watcher by default — the old message blamed the env var even when it was never set.
- Add `isUnscopedAggregatorSessionId()` and use it to stop a background dashboard process's own live subagent total from being added on top of an already-persisted session's total, closing a double-counting path that opting such a process into its own watcher (`NR_AI_WATCHER_MODE=local`) would otherwise open.
- Bump to v1.14.6, CHANGELOG entry.

## Test plan
- [x] `npm test` (Jest) — 5175 passed
- [x] `npx vitest run` (web) — 34 passed in `Today.test.tsx`
- [x] `npm run lint` — clean
- [x] `npx tsc -b .` and `npx tsc --noEmit -p tsconfig.web.json` — clean
- [x] `npm run build` — clean
- [x] New tests added TDD-first for: `todayPortionRatio`, persisted `subagentCostUsd` round-trip, cross-session aggregate summation, `isUnscopedAggregatorSessionId`, and the banner's three states (env_var / mode_mismatch / absent-field fallback)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.6] - 2026-07-30
+
+### Fixed
+
+- **Subagent spend on the Today dashboard only reflected whichever session happened to be serving the dashboard** — with multiple Claude Code sessions running concurrently (a common pattern with subagents and workflows), only one process's own in-memory tracker ever backed the "subagent spend" figure, silently excluding subagent activity from every other session. It's now aggregated from each session's persisted totals, the same way overall spend already was.
+- **The "subagent cost tracking is disabled" banner named the wrong cause for background dashboard processes** — a dashboard running in standalone/background mode doesn't run its own subagent watcher by default (by design), which the banner incorrectly attributed to the `NR_AI_ENABLE_SUBAGENT_WATCHER` environment variable even when that variable was never set. The banner now distinguishes the two cases and gives accurate guidance for each.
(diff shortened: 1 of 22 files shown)


corinnekunze:

Issue #305, opened 2026-07-30
Pi's --local-only dashboard daemon never tracks subagent cost by default


chrisdhaan:
## Summary

Pi has no `--stdio` MCP client support at all (confirmed platform design choice, see `docs/ADAPTERS.md`), so its entire setup path is the macOS `launchd` dashboard daemon (`installDashboardDaemon()` in `src/install/schedule.ts`) running `--local`, with no `--stdio` counterpart ever running alongside it.

The subagent/workflow transcript watchers only auto-start under `--stdio` by default (`NR_AI_WATCHER_MODE=stdio`). As of the recent fix in #324 (v1.14.6), a `--local` process not running its own watcher is correctly treated as expected — the Today view's spend already aggregates every session's persisted totals, and the dashboard shows an accurate banner explaining the default and how to opt in (`NR_AI_WATCHER_MODE=local`).

That framing assumes there's *some* `--stdio` session contributing persisted totals for the watcher to defer to. For Pi, there never is one — so Pi's dashboard daemon permanently gets zero subagent cost tracking unless the operator manually discovers and sets `NR_AI_WATCHER_MODE=local` themselves, with no Pi-specific guidance pointing them at it.

## What this needs

Something in Pi's own setup path (`installDashboardDaemon()`, ADAPTERS.md's Pi setup instructions, or the setup wizard) should either set `NR_AI_WATCHER_MODE=local` automatically for Pi specifically, or explicitly instruct the user to set it — rather than leaving Pi users to discover this env var from source.

Note: it's now safe to set `NR_AI_WATCHER_MODE=local` even when a `--stdio` sibling *does* happen to be running (a follow-up fix makes the unfiltered watcher skip any session with a live `--stdio` heartbeat), so whatever fix lands here doesn't need to worry about that interaction.

## Related
- #324 (v1.14.6 — cross-session subagent aggregation + watcher-disabled banner)

PR #306, opened 2026-07-30
Fix: cross-process races between --local and --stdio (v1.14.7)

### LINKED ISSUE #307: --local and --stdio silently race over the same buffer/cursor files when run concurrently
## Summary

Two related cross-process races when a `--local` dashboard process runs alongside one or more active `--stdio` (Claude Code) sessions:

1. **Buffer-drain race**: `--local`'s cross-session buffer drain and a `--stdio` session's own scoped drain both poll the same per-session buffer file, with no coordination. Whichever process wins a given poll cycle (every 100ms) keeps that batch of tool-call/token events; the other sees nothing, with no error on either side. Silently undercounts that session's own stats/cost, and for a session configured for cloud telemetry, drops those events from New Relic entirely.
2. **Subagent-watcher race**: opting a `--local` process into its own subagent watcher (`NR_AI_WATCHER_MODE=local`) while a `--stdio` sibling is also running causes both to redundantly tail the same subagent transcripts and race over the same cursor files.

## Impact

Anyone running a persistent `--local` dashboard (e.g. a background daemon) alongside active Claude Code sessions in one or more projects — a common and intentionally-supported deployment shape — could see silently undercounted session stats, cost, and (for cloud-configured sessions) missing New Relic telemetry, with no indication anything went wrong.

## Environment
- @newrelic/preflight 1.14.6

chrisdhaan:
Fixes #307

## Summary

- **Buffer-drain race**: a `--local` dashboard process's cross-session buffer drain could race a live `--stdio` session's own scoped drain over the exact same per-session buffer file, with no coordination. Whichever process won a given poll cycle (every 100ms) kept that batch of tool-call/token events; the other saw nothing, with no error on either side — silently undercounting that session's own stats/cost, and for a session configured for cloud telemetry, dropping those events from New Relic entirely. The cross-session drain now skips any per-session buffer whose heartbeat names a live PID.
- **Subagent-watcher race**: opting a `--local` process into its own subagent watcher (`NR_AI_WATCHER_MODE=local`) while a `--stdio` sibling was also running caused both to redundantly tail the same subagent transcripts and race over the same cursor files. The unfiltered watcher now skips any session with a live `--stdio` heartbeat, the same way.
- `docs/ADVANCED.md` updated to describe when to set `NR_AI_WATCHER_MODE=local` yourself, and that it's now safe to combine with concurrently-running `--stdio` sessions.

Related: #305 (a separate, still-open gap — Pi's `--local`-only dashboard has no `--stdio` sibling to ever defer to, so it needs its own fix to enable subagent tracking by default).

## Test plan

- [x] New test coverage for both races
- [x] `npm run lint` — clean
- [x] `npx tsc -b .` — clean
- [x] `npx jest` — all passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.7] - 2026-07-30
+
+### Fixed
+
+- **A `--local` dashboard process running alongside active `--stdio` sessions could silently steal a portion of a session's tool-call and token events.** Both processes poll for new events on the same per-session buffer file; `--local`'s cross-session drain had no way to tell that a `--stdio` process already owned and was draining that exact file itself. Whichever side won a given poll cycle kept the events, and the other side saw nothing, with no error on either side — silently undercounting that session's own stats and cost tracking, and, for a session configured for cloud telemetry, dropping those events from New Relic entirely. `--local` now skips any per-session buffer that has a live `--stdio` owner.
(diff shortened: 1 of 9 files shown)


corinnekunze:

PR #308, opened 2026-07-30
Fix: cross-midnight sessions invisible to Today heatmap/concurrency chart

### LINKED ISSUE #309: Today dashboard heatmap/concurrency chart miss activity from cross-midnight sessions
## Description

The Today dashboard's activity heatmap and concurrency chart can silently miss activity from a session that started before local midnight and is still running.

Both charts read sessions via a helper that filters by a session's *start* date (derived from its persisted filename). A session that started at, say, 11:45pm and is still active after midnight never matches that filter, so its portion of today's activity doesn't show up in either chart — even though the "spend today" figure elsewhere on the same page already accounts for this case correctly via a separate helper.

## Expected behavior

A session that spans midnight should still have its post-midnight activity reflected in the Today heatmap and concurrency chart, consistent with how "spend today" already handles it.

## Actual behavior

The post-midnight portion of a cross-midnight session's activity is missing from the heatmap and concurrency chart.

chrisdhaan:
Fixes #309

## Summary
- The Today dashboard's activity heatmap and concurrency chart both read `sessionStore.loadTodaySessions()`, which filters sessions by *start* date (derived from the persisted filename). A session that started before local midnight and is still running never matches that filter, so its today-portion silently disappeared from both charts — even though the "spend today" KPI already handled this exact case via a separate helper (`loadSessionsOverlappingToday()`).
- Both routes now prefer `loadSessionsOverlappingToday()`, falling back to `loadTodaySessions()` if the store doesn't implement it. Each route already clamps individual timestamps/windows to today's boundary downstream, so widening the input session set can't leak yesterday's activity into today's buckets.
- Bumped to v1.14.8 with a CHANGELOG entry.

## Test plan
- [x] Added a failing-first test to both the `/api/concurrency` and `/api/activity-heatmap` describe blocks in `api-handler.test.ts`, each covering a session that started before local midnight and is still active — confirmed both failed before the fix (bucket count 0 instead of 1).
- [x] Implemented the fix; both new tests pass.
- [x] Full suite, lint, and build all pass in this clone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.8] - 2026-07-30
+
+### Fixed
+
+- **The Today dashboard's activity heatmap and concurrency chart could show a gap for a session that started before local midnight and was still running** — both read only sessions whose _start_ date matched today's, so a session crossing midnight was invisible to them even though its today-portion should have counted (the "spend today" figure already handled this case correctly via a separate helper). Both now include cross-midnight sessions, counting only their activity that actually falls within today.
+
 ## [1.14.7] - 2026-07-30
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


corinnekunze:

PR #310, opened 2026-07-30
Fix: Today dashboard cross-session blind spots (efficiency KPI + anti-pattern banner)

### LINKED ISSUE #311: Today dashboard efficiency KPI and anti-pattern banner miss data from other sessions
## Description

Two panels on the Today dashboard only reflect whichever process happens to be serving the dashboard (for example, a persistent background `--local` daemon), not today's actual coding sessions when those run in separate processes.

1. **Efficiency KPI**: shows no score (or a stale one) even when today's sessions have real efficiency data, because it only reads the dashboard-serving process's own live tracker instead of aggregating across today's sessions.
2. **Anti-pattern detail banner**: the "flags" count at the top of the page can be non-zero while the detail banner underneath it (which pattern fired, on which file) stays blank, because it only checks the dashboard-serving process's own live detection instead of falling back to already-persisted session data.

## Expected behavior

Both should reflect today's activity across every session, the same way the other Today KPIs (spend, tool calls, activity heatmap, concurrency) already do.

## Actual behavior

Both silently show no/incomplete data when the dashboard-serving process didn't itself do the work being measured.

chrisdhaan:
Fixes #311

## Summary
- **Efficiency KPI**: the Today dashboard's "efficiency" figure only reflected whichever process happened to be serving the dashboard (e.g. a persistent `--local` background daemon), not today's actual coding sessions running in separate `--stdio` processes. It's now averaged across today's sessions via `/api/sessions/today/aggregate`'s new `avgEfficiencyScore` field, the same way the other Today KPIs already aggregate.
- **Anti-pattern detail banner**: the banner naming which anti-pattern fired (and on which file) could stay blank even when the flags count above it was non-zero, because it only checked the dashboard-serving process's own live detection. It now falls back to already-persisted session data when neither the live SSE store nor this process's own `/api/anti-patterns` has anything.
- Bumped to v1.14.9 with a CHANGELOG entry covering both.

## Test plan
- [x] Both fixes built via TDD (failing test first, confirmed RED, minimal implementation, confirmed GREEN).
- [x] Each change independently reviewed (spec + quality) and approved, plus a clean whole-branch review.
- [x] `npm run build` — clean.
- [x] `npm run lint` — clean.
- [x] `npm test` (Jest) — 5187 passed, 3 pre-existing skips.
- [x] `npm run test:web` (Vitest) — 419 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.9] - 2026-07-30
+
+### Fixed
+
+- **The Today dashboard's efficiency KPI only reflected whichever process happened to be serving the dashboard** — with multiple Claude Code sessions running concurrently, only one process's own in-memory efficiency tracker ever backed the "efficiency" figure, silently excluding coding productivity data from every other session. It's now aggregated across today's sessions, the same way the other Today KPIs (activity, concurrency, spend) already were.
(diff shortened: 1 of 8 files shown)


corinnekunze:

PR #312, opened 2026-07-30
Fix: cross-process live-session detection (v1.14.10)

### LINKED ISSUE #313: Live-sessions list and concurrency count miss sessions running in other processes
## Description

The live-sessions list and the "current concurrency" count on the Today dashboard only see sessions handled by the exact process serving the dashboard (for example, a persistent background daemon), not sessions actively running in a different concurrent process.

1. **Live-sessions list**: a session that's live in another process never appears, so the Today selector can't default to it and it never shows up as a clickable row.
2. **Concurrency count**: the "current" figure undercounts whenever a live session is only visible via a different process's activity, and is inconsistent with the peak/all-time-peak figures shown right next to it, which already account for cross-process activity.

## Expected behavior

Both should reflect every session that's actually live right now, regardless of which process is serving the dashboard — the same way the other cross-session Today panels already do.

## Actual behavior

A session live in a different process is invisible to both the live-sessions list and the concurrency count, even though its activity is already on disk and visible to other cross-session views.

chrisdhaan:
Fixes #313

## Summary
- The live-sessions list and the "current concurrency" count only saw sessions handled by the exact process serving the dashboard, missing sessions live in a different concurrently-running process.
- Adds a shared helper that unions the per-process live-session registry with session IDs seen in any process's pending activity buffer, reusing the registry's own staleness window and synthetic-session-id filtering.
- Wires the helper into both the live-sessions and concurrency routes. The concurrency count now also excludes synthetic session ids, matching the peak/all-time-peak fields already in that response (previously inconsistent).

## Test plan
- [x] `npm run build && npm test` — 5195 passing, 0 failures
- [x] `npm run test:web` — 419 passing, 0 failures
- [x] New tests cover: union of registry- and buffer-derived session ids, staleness boundary exclusion, synthetic-id exclusion, dedup, both routes' wiring
- [x] Version bumped to 1.14.10, CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.10] - 2026-07-30
+
+### Fixed
+
+- **The live-sessions list and the "current concurrency" count only saw sessions that the dashboard-serving process itself had handled** — with multiple Claude Code sessions running against different concurrent processes, a session live in another process never appeared in the Today selector or counted toward current concurrency, even though its activity was visible on disk. Both now also check every process's pending activity buffer, so a session shows up as live regardless of which process is serving the dashboard.
+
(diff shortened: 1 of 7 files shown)


corinnekunze:

PR #315, opened 2026-07-30
Fix: /api/context sees sessions live in other processes

### LINKED ISSUE #314: ContextBar shows no data for a session live only in a different process
The context-window bar on the Today dashboard reads context-window usage from an in-memory tracker scoped to whichever process is currently serving the dashboard. When a session is actively running in a different process than the one serving the dashboard, that process's own tracker has no data for it, so the context bar goes blank even though the session has real activity.

This is the same class of cross-process blind spot as the other Today dashboard fixes shipped recently (live-session detection, efficiency KPI, anti-pattern detail banner): a per-process in-memory tracker with no fallback to the activity every process can already see on disk.

chrisdhaan:
## Summary
- Extracts the context-window tracker's scoring logic into a pure replay function, so metrics can be computed from a snapshot of events instead of only via incremental live recording.
- The context API now falls back to recomputing metrics from every process's pending activity when the dashboard-serving process's own tracker has no data for the requested session — fixing the context bar going blank for a session live only in a different process.
- No new persistence: everything needed is already present in the existing cross-process activity snapshot.
- Bumps version to 1.14.11 with a matching CHANGELOG entry.

Fixes #314

## Test plan
- [x] `npm run build` — clean
- [x] `npm test` — 5204/5207 passing (3 pre-existing skips)
- [x] `npm run test:web` — 419/419 passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.11] - 2026-07-30
+
+### Fixed
+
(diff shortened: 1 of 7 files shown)


corinnekunze:

PR #317, opened 2026-07-31
Fix: Latency panel reflects tool calls from every session today

### LINKED ISSUE #316: Latency panel only reflects tool calls from the process serving the dashboard
The Today dashboard's Latency panel (p50/p95/p99) is computed from a single process's own in-memory tracker. When multiple Claude Code sessions run concurrently against different processes, only the one process happening to serve the dashboard contributes latency data — tool calls handled by any other concurrent session are silently excluded, even though their activity is visible on disk.

chrisdhaan:
Fixes #316

## Summary
- The Today dashboard's Latency panel (p50/p95/p99) only reflected tool calls from whichever process happened to be serving the dashboard, missing latency data from any concurrently running session in a different process.
- Extends the existing cross-session aggregate route (which already unions every process's undrained buffer plus every persisted session's timeline for other KPIs) to also collect duration/tool samples and compute percentiles via a new pure `computeLatencyPercentiles()` function.
- Updates the Latency panel to read the new field from the aggregate payload instead of querying the separate per-process latency endpoint (unchanged — still used elsewhere).
- No new persistence — everything comes from data the aggregate route already reads.
- Version bump to 1.14.12 + CHANGELOG entry.

## Test plan
- [x] `npm run test:web` — 420/420 passing
- [x] New unit tests: pure percentile function, aggregate route latency field, frontend panel (asserts it does not fall back to the per-process endpoint)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.12] - 2026-07-31
+
+### Fixed
+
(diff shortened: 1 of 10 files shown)


corinnekunze:

PR #319, opened 2026-07-31
Fix: Cache Health panel reflects token usage from every session today

### LINKED ISSUE #318: Cache Health panel only reflects token usage from one process
The Today dashboard's Cache Health panel (hit rate percentage, status, and dollar savings) reads only the in-memory cost tracker of whichever process happens to be serving the dashboard. With multiple concurrent sessions running against different processes, cache activity from every other session is silently excluded — the panel can show a low hit rate or no activity at all even while other sessions are actively benefiting from prompt caching.

The fix aggregates cache hit rate and savings across every session active today, the same way the dashboard's other cross-process KPIs (efficiency, latency, spend, anti-patterns) already work.

chrisdhaan:
## Summary
- The Today dashboard's Cache Health panel (hit rate %, status, dollar savings) previously read only the dashboard-serving process's own in-memory cost tracker, silently excluding cache activity from every other concurrently running session.
- Adds a `cacheHealth` field to the existing cross-session `GET /api/sessions/today/aggregate` route, blending live undrained token events across every process with persisted per-session token totals (pro-rated for cross-midnight sessions), via a new pure `computeCacheHealth()` helper mirroring the existing latency-percentiles aggregation.
- The frontend Cache Health panel now reads hit rate/savings/status from that aggregate field; it keeps its existing per-process query only for the week-over-week trend figure, which has no cross-process bug since it's already derived from persisted weekly summaries.

Fixes #318

## Test plan
- [x] `npm run build` — clean
- [x] `npm run test:web` — full suite passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.13] - 2026-07-30
+
+### Fixed
+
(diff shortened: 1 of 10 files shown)


corinnekunze:

PR #326, opened 2026-07-31
Fix: Forecast · End of Day reflects cross-process activity today

### LINKED ISSUE #325: Forecast · End of Day only reflects one process's own activity
The Today dashboard's "Forecast · End of Day" card only projects spend from whichever process happens to be serving the dashboard — its burn-rate calculation comes from that one process's own in-memory cost tracker, which is empty on a dashboard-only process with no coding activity of its own. The forecast silently fails to account for spend happening in other concurrently running sessions.

This is the same class of cross-process blind spot already fixed for the other Today dashboard KPIs (tool selection score, cache health, latency percentiles, efficiency, anti-patterns).

chrisdhaan:
## Summary
- The Today dashboard's "Forecast · End of Day" card only projected spend from whichever process happened to be serving the dashboard — its burn-rate anchor came from that one process's own in-memory cost tracker, which is empty on a dashboard-only process with no coding activity of its own.
- `/api/sessions/today/aggregate` now derives the earliest today-scoped activity across the live cross-process buffer union, persisted same-day sessions, cross-midnight sessions, and this process's own live tracker, and computes `forecastEndOfDayUsd` from it — the same route the other cross-session Today dashboard KPIs already use.
- The Today view now consumes that field as its fallback when the per-session live push (which reflects only this process's own session, correctly, and stays first-priority) is unavailable, replacing an older client-side delta-math workaround.

Fixes #325

## Test plan
- [x] `npm run build`
- [x] `npm test` (5240/5243 passing, 3 pre-existing skips, 0 failures)
- [x] `npm run test:web` (422/422 passing)
- [x] `npm run lint` (0 errors/warnings)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.15] - 2026-07-31
+
+### Fixed
+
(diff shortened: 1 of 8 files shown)


corinnekunze:

PR #328, opened 2026-07-31
Fix: Model Usage panel reflects cross-process activity today

### LINKED ISSUE #327: Today dashboard: Model Usage panel only reflects the process serving the dashboard
The Today dashboard's Model Usage panel only reflects model/token/cost activity from whichever process happens to be serving the dashboard — its per-model request counts and cost-per-token figures come from that one process's own in-memory tracker, silently excluding activity from every other concurrently running session and resetting whenever that process restarts.

This is the same class of blind spot already fixed for the Latency, Cache Health, Tool Selection, and Forecast · End of Day panels on this dashboard.

chrisdhaan:
## Summary
- The Today dashboard's Model Usage panel only reflected model/token/cost activity from whichever process happened to be serving the dashboard.
- Adds a raw per-model breakdown getter and a combine method that sums raw counters across sources and derives cost-per-token ratios exactly once — never by averaging per-source ratios.
- Persists that raw breakdown per session so completed/checkpointed sessions carry it to disk, with a safe default for legacy files.
- The Model Usage endpoint now combines this process's own live breakdown with every other today-persisted session's breakdown, mirroring the same pattern already used for the Tool Selection Score endpoint on this dashboard.

Fixes #327

## Test plan
- [x] `npm run build && npm test` — full suite passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.16] - 2026-07-31
+
+### Fixed
+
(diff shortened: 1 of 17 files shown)


corinnekunze:

PR #334, opened 2026-08-01
Fix: add cwd breadcrumb fallback to synchronous session-ID resolution

### LINKED ISSUE #320: Native Windows + cloud/both mode: no events are ever sent (NrIngestManager gated on !isProvisional; sync resolution lacks the cwd fallback) - metrics-only, fully silent
## Summary

On **native Windows**, a `mode: 'cloud'` / `mode: 'both'` install silently sends **no events at all** (`AiToolCall`, `AiMcpToolCall`, `AiCodingTask`, `AiAntiPattern`, …). Metrics arrive fine, so the integration looks healthy — logs show `Sent metrics to NR` on schedule, no errors, no warnings — but every NRQL query against the event types returns 0 rows, forever.

Root cause: the `--stdio` MCP's **synchronous** session-id resolution has no cwd-breadcrumb fallback, so on native Windows it always starts provisional; and `NrIngestManager` — the only thing that feeds tool events into the harvest queue — is gated on `!isProvisional` and is never constructed. The later async resolution repairs the session id but the ingest path stays dead for the life of the process.

This is effectively the cloud-transport half of #268: that fix repaired local persistence via a cwd breadcrumb, but only on the async path, so cloud ingest still never initializes.

## Root cause

`resolveFromCwd()` (added in #268 / PR #282) is used by the async breadcrumb poller, but the **synchronous** startup resolution still only tries job-dir + **PID-keyed** breadcrumb — the exact lookup that never matches on native Windows, since hooks run through Git Bash and the collector's ppid is a transient shell pid, not `claude.exe`:

```js
// src/index.ts — startup, synchronous
const synchronouslyResolved =
    resolveFromJobDir(process.env.CLAUDE_JOB_DIR ?? null) ??
    resolveFromBreadcrumb(config.storagePath, process.ppid);   // ← no resolveFromCwd()
```

So `sessionTraceId` becomes `pending-<ts>` and:

```js
const isProvisional = options.stdio && sessionTraceId.startsWith('pending-');
...
if (config.mode !== 'local' && !isProvisional) {
    nrIngest = new NrIngestManager({ ... });   // ← never runs on native Windows
    capturedNrIngest = nrIngest;
}
```

Every tool record then hits an optional-chained no-op, which is why nothing is logged:

```js
capturedNrIngest?.ingestToolCall(record, auditRecord);   // capturedNrIngest === undefined
```

Metrics keep flowing because they go through a separate path that doesn't depend on `NrIngestManager`, which is what makes this look like a working cloud integration.

There *is* a late `NrIngestManager` construction in the async-resolution block, but it doesn't rescue this case: the process resolves via the "immediate" cwd path and reports `Session ID resolved, full initialization complete`, yet no events are ever queued or sent.

## Evidence (1.14.13, native Windows 11, mode: both, region us)

**Broken (normal startup):**

```
level=info  msg="Session ID not yet available; using provisional ID, dashboard will start early"
level=info  msg="Session ID resolved, full initialization complete"  sessionTraceId=5a0c277c-…
level=debug msg="Sent metrics to NR"  scope=metrics  batchSize=14
```
→ `scope=events` never appears, across repeated runs and with fresh hook events generated while the process was alive. Config validates clean (`mode: both`, numeric accountId, `NRAL`-suffixed ingest key, region us), and the collector writes buffer entries normally.

**Control (force synchronous resolution via `CLAUDE_JOB_DIR`, same version/config):**

```
level=info  msg="Session ID resolved synchronously"  sessionTraceId=5a0c277c-…
level=debug msg="Sent events to NR"   scope=events   batchSize=17   ← only ever appears here
level=debug msg="Sent metrics to NR"  scope=metrics  batchSize=19
```

The only variable changed between the two is whether startup resolution succeeded synchronously — which is exactly the `isProvisional` gate.

## Repro

1. Native Windows Claude Code, Preflight installed natively, hooks via Git Bash.
2. `preflight install --license-key <ingest key> --account-id <numeric id>`, then set `"mode": "both"` in `config.json`.
3. Restart Claude Code, use it for a while.
4. NRQL: `SELECT count(*) FROM AiToolCall SINCE 1 hour ago` → **0**, indefinitely. `preflight --validate` is clean and the debug log shows only `scope=metrics` harvests.

## Suggested fix

Add the cwd fallback to the synchronous resolution so native Windows starts non-provisional in the first place:

```js
const synchronouslyResolved =
    resolveFromJobDir(process.env.CLAUDE_JOB_DIR ?? null) ??
    resolveFromBreadcrumb(config.storagePath, process.ppid) ??
    resolveFromCwd(config.storagePath, process.cwd());
```

Worth pairing with defence in depth, since the failure is completely silent:

- Make the late (post-async-resolution) `NrIngestManager` construction authoritative for every path that leaves provisional state, and re-attach `capturedNrIngest` so already-drained records aren't dropped.
- Log a one-time `warn` when `mode !== 'local'` but `capturedNrIngest` is undefined at the time a tool record arrives — that single line would have made this a two-minute diagnosis instead of a source dive.
- Consider surfacing "events sent: N" in `nr_observe_health`, so a cloud install can be verified without NRQL.

## Environment

- @newrelic/preflight 1.14.13 (also reproduced on 1.14.5; code path unchanged on current `main`)
- Windows 11, native Claude Code 2.1.217, Node v22.23.1, `platformTarget: native`, `mode: both`, region `us`
- New Relic Free tier account

chrisdhaan:
## Summary

- Synchronous startup session-ID resolution only tried `resolveFromJobDir()` and `resolveFromBreadcrumb()` (PPID-keyed), skipping the cwd-keyed breadcrumb the async resolver already falls back to. On setups where the PPID breadcrumb never matches (e.g. native Windows, where the hook collector's PPID is a transient shell process rather than the real parent), every session paid an avoidable delay in the provisional window. Synchronous resolution now matches the async resolver's fallback order.
- Added a one-time warning log when a tool record arrives while cloud mode is configured but ingest hasn't initialized yet, so a stuck provisional window is now visible in the logs instead of silently dropping events.
- Version bump to 1.14.18 with a CHANGELOG entry.

Addresses part of #320 — see my comment there for what this PR does and doesn't cover. The issue's originally stated root cause (that cloud ingest never initializes at all) didn't hold up under a live repro: the async rescue path already recovers once the cwd breadcrumb resolves. This PR closes the one confirmed gap (missing cwd fallback in the *synchronous* path) plus a defense-in-depth log line — it does not close #320 outright, since the deeper "zero events on native Windows" symptom (if still real) likely needs a separate fix. Leaving #320 open pending confirmation from the reporter.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — full suite passing (added a new integration test: `resolves session_id synchronously from the cwd breadcrumb when no CLAUDE_JOB_DIR or PPID breadcrumb is available`, which fails without the fix)
- [x] `npm run lint` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.18] - 2026-08-01
+
+### Fixed
+
+- **The `--stdio` MCP's synchronous session-ID resolution at startup only tried the job-dir and PPID-keyed breadcrumb, skipping the cwd-keyed breadcrumb that the async resolver already falls back to** — on setups where the PPID breadcrumb never matches (e.g. native Windows, where the hook collector's PPID is a transient shell process rather than the real parent), this meant every session started in the provisional window and paid an avoidable delay before cloud ingest wired up. Synchronous resolution now tries the cwd breadcrumb too, matching the async resolver's fallback order. Also added a one-time warning log when a tool record arrives while cloud mode is configured but ingest hasn't initialized yet, so a stuck provisional window is now visible in the logs instead of silently dropping events.
+
 ## [1.14.17] - 2026-07-31
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.17",
+  "version": "1.14.18",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.14.17",
+      "version": "1.14.18",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.14.17",
+  "version": "1.14.18",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,5 +1,13 @@
 import { jest, describe, it, expect, beforeEach, afterEach } from '@jest/globals';
-import { mkdirSync, mkdtempSync, writeFileSync, rmSync, existsSync, utimesSync } from 'node:fs';
+import {
+  mkdirSync,
+  mkdtempSync,
+  writeFileSync,
+  rmSync,
+  existsSync,
+  utimesSync,
+  realpathSync,
+} from 'node:fs';
 import { tmpdir } from 'node:os';
 import { resolve, join } from 'node:path';
 import {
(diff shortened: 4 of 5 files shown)


corinnekunze:

PR #335, opened 2026-08-01
Fix: parent-session cost tracking dropped turns with no tool call

### LINKED ISSUE #336: Parent-session cost tracking silently drops turns with no tool call
## Description

Token/cost capture for the parent session only runs when a PreToolUse/PostToolUse hook fires, and even then only reads the single most recent assistant-turn usage from the transcript tail.

Any assistant turn that produces no tool call never triggers a scan at all, and if more than one new turn appears between two consecutive hook firings, only the last one is captured — every earlier one is silently and permanently lost, with no error or warning anywhere.

## Impact

Measured against real session transcripts: roughly half of assistant turns produce no tool call, and they account for roughly half of real token cost in a mostly-conversational session. Cost/token dashboards and reports can significantly undercount actual spend as a result — worse in sessions with more back-and-forth conversation relative to tool use.

## Fix

Replaces the hook-triggered scanner with a poller that tails each session's own transcript independently of tool-call timing, using a durable byte cursor so every real turn is captured regardless of whether it called a tool.

chrisdhaan:
Fixes #336

## Summary

- Token/cost capture for the parent session previously only ran when a PreToolUse/PostToolUse hook fired, and even then captured only the single most recent assistant-turn usage from the transcript tail. Any turn that produced no tool call never triggered a scan, and any turn superseded by another before the next hook fired was silently and permanently lost — measured against real transcripts, roughly half of turns and cost in a mostly-conversational session.
- Replaces that mechanism with `ParentTranscriptWatcher`, a poller in the long-running server that tails each session's own transcript independently of tool-call timing via a durable byte cursor with partial-line carryover, mirroring the existing `SubagentWatcher` pattern. Also fixes an adjacent bug: the old scanner never skipped sidechain (subagent) transcript entries, so subagent turns inlined into the main transcript could be double-attributed as parent cost.
- Adds `messageId`-based dedup for cursor-replay safety, and garbage collection for the new cursor files.
- Deliberately not gated behind the same mode flag `SubagentWatcher`/`WorkflowWatcher` use — parent-token capture ran unconditionally in every mode before this change, and gating it the same way would silently zero out cost tracking for standalone deployments.

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [x] `npm run lint`
- [x] New test suite includes a direct regression test: a tool-call / text-only / tool-call sequence, asserting all three turns are captured


corinnekunze:

PR #337, opened 2026-08-01
Fix: parent-token cost events weren't stamped with their real timestamp for day-bucketing

### LINKED ISSUE #338: Spend Today inflates and fluctuates after upgrading to v1.14.19+
## Description

After upgrading, the dashboard's "Spend Today" figure can climb continuously and jump between wildly different values (e.g. oscillating between a low, correct-looking number and one several times higher) instead of settling on a stable total.

## Root cause

`ParentTranscriptWatcher` (added in v1.14.19) feeds parent-session token events into the cost tracker without an explicit event timestamp, so day-bucketing falls back to arrival time instead of the turn's real transcript timestamp. This is harmless in steady state, but the watcher also backfills up to 24h of transcript history in `--local` mode — every replayed historical turn gets mis-attributed to *today's* spend while the backlog is processed, inflating and destabilizing the figure until the backfill catches up.

## Fix

Stamp parent-token cost events with their real transcript timestamp before recording, matching how the sibling subagent-token path already does it.

chrisdhaan:
Fixes #338

## Summary
- `ParentTranscriptWatcher` (added in v1.14.19) feeds parent-session token events into `CostTracker.recordTokenUsage()` without an explicit `timestampMs`, so day-bucketing falls back to arrival time instead of the turn's real transcript timestamp — the sibling subagent-token path already passes `timestampMs` correctly.
- This was latent while nothing replayed history, but became a real bug once the watcher started backfilling up to 24h of transcript history in `--local` mode: every replayed historical turn got counted as *today's* spend, inflating and destabilizing the "Spend Today" figure while the backlog was processed.
- Fix: pass `{ timestampMs: tokenEvent.timestamp }` at the parent-token call site, matching the existing subagent-token wiring.

## Test plan
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm test` (full suite passes)
- [x] `ctx.timestampMs`-based day-bucketing is already thoroughly covered by `cost-tracker.test.ts`; this fix corrects a wiring omission, not tracker logic

Co-authored-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.20] - 2026-08-01
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


corinnekunze:

PR #339, opened 2026-08-01
Fix: self-correct a stale cwd-fallback session ID adoption

### LINKED ISSUE #340: Stale cwd-fallback session ID can be adopted permanently at startup
## Bug

A `--stdio` MCP process resolves its own session identity at startup from two breadcrumb files written by the hook collector: a PPID-keyed one (precise — no cross-session collision) and a cwd-keyed one (a fallback, shared by every session that has ever run in that working directory).

If the PPID breadcrumb hasn't been written yet when startup resolution runs, the process can fall back to the cwd breadcrumb — which may hold a stale value left over from an unrelated prior session in the same directory. That stale value gets adopted permanently, with no way to self-correct once the real PPID breadcrumb appears moments later.

Every metric, checkpoint, and tool call for that session is then misattributed to the wrong session ID.

## Fix

The MCP now watches for a PPID-breadcrumb resolution in the background whenever it had to fall back to the cwd breadcrumb, and re-adopts the correct session identity the moment one appears — preserving all metrics already recorded under the earlier one.

chrisdhaan:
Fixes #340

## Summary
- A `--stdio` MCP process could permanently adopt the wrong session identity when startup resolution fell back to the cwd-keyed breadcrumb before the precise PPID-keyed one was available — the cwd breadcrumb is shared by every session that has ever run in that directory, so a stale value from an unrelated prior session could win the race with no way to self-correct.
- `resolveSessionId()` now reports which source resolved it, and a new `watchPpidBreadcrumb()` keeps watching in the background whenever the initial resolution came from the cwd fallback.
- Extracted the existing pending→real adoption logic into a reusable `adoptRealSessionId()` (now safe to call a second time — stops any already-live `NrIngestManager` first) and wired the background watch from both entry points that can resolve via cwd: the synchronous startup path and the async pending-poller path.
- Accumulated metrics survive the correction; only the metrics-carrying identity changes.
- Version bump to 1.14.21 + CHANGELOG entry.

## Test plan
- [x] New unit tests in `src/hooks/session-resolver.test.ts`: `onResolutionSource` reporting for all three sources, `watchPpidBreadcrumb()` (resolves on ppid, never trusts cwd, aborts via signal), and a regression test for the race itself
- [x] New spawned-process integration tests in `src/index.test.ts` covering both trigger sites (synchronous cwd resolution, and pending→async-poller cwd resolution), each confirming the stale id is corrected once a differing PPID breadcrumb appears and metrics survive
- [x] `npm run build && npm test && npm run lint && npm run format:check` all clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.21] - 2026-08-01
+
+### Fixed
+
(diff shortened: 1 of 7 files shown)


corinnekunze:

PR #341, opened 2026-08-01
Fix: stop stale cwd-fallback session ids from contaminating persisted data

### LINKED ISSUE #342: Stale cwd-fallback session ids can contaminate persisted session data
## Description

When a `--stdio` MCP process resolves its session identity via the collision-prone cwd fallback before the precise PPID breadcrumb confirms it, the brief unconfirmed window can cause lasting damage even after the existing self-correction mechanism runs:

1. **Orphaned ghost files** — if the periodic 30-second checkpoint fires before the real PPID-derived identity is confirmed, it writes a real, non-synthetic, zero-activity session file to disk under the stale identity. That file is never cleaned up.
2. **Cost/token contamination** — the parent-transcript watcher (scoped to whichever identity is currently live) can backfill an unrelated old session's cost/token data into the live cost/model-usage trackers during that same unconfirmed window. Since the existing self-correction only relabels the tracker in place, that contamination rides along uncorrected into the real session's own totals once the PPID breadcrumb confirms the correct identity.

## Fix

Checkpointing is now suppressed for as long as a session's identity is unconfirmed (capped at two minutes so a session that never gets a PPID breadcrumb doesn't lose checkpointing indefinitely), and a real correction now resets the affected cost/token trackers before continuing so nothing from the wrong identity carries forward.

chrisdhaan:
Fixes #342

## Summary

- When a `--stdio` MCP process resolved its session identity via the collision-prone cwd fallback before the precise PPID breadcrumb confirmed it, the brief unconfirmed window could cause lasting damage even after the existing self-correction landed: the periodic 30-second checkpoint could write a real, non-synthetic, zero-activity session file to disk under the stale identity, orphaning it, and any cost/token data backfilled from an unrelated old session under that stale identity rode along uncorrected into the real session's own totals once corrected, since the correction only relabeled the tracker in place.
- Checkpointing is now suppressed for as long as a session's identity is unconfirmed (capped at two minutes so a session that never gets a PPID breadcrumb doesn't lose checkpointing indefinitely), and a real correction now resets the affected cost/token trackers before continuing so nothing from the wrong identity carries forward.
- Version bump to 1.14.22 + CHANGELOG entry.

## Test plan

- [x] `npm run build && npm test` — full suite green
- [x] 6 new integration tests covering: checkpoint suppression while unconfirmed, cap-timeout fallback, immediate-clear on confirmed-same-id, tracker reset on a real correction, prompt checkpoint resume after a real correction, and shutdown still persisting the terminal save mid-pending

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.22] - 2026-08-01
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


corinnekunze:

PR #358, opened 2026-08-05
Fix: preflight install --mode now actually defaults to local

### LINKED ISSUE #359: preflight install --mode help text claims a default that wasn't enforced
## Description

`preflight install`'s `--mode <mode>` option (added in #355) has help text claiming a default of `local`:

```
--mode <mode>        Telemetry mode: cloud, local, or both (default: local)
                     (choices: "cloud", "local", "both")
```

That default was never actually enforced. Omitting `--mode` never wrote a `mode` key to `config.json`, so the server's own config-loading fallback resolved the absence to `cloud` instead — the opposite of what the help text claims, and of this project's own documented product default ("Skip [the New Relic account] to run local-only (the default)" in the README).

## Impact

Anyone running the non-interactive install with cloud credentials but without `--mode` would silently end up in cloud mode, sending telemetry to New Relic when they may have expected local-only operation.

## Fix

Fixed in #358 (synced from an internal follow-up): `handleInstall` now writes `mode: 'local'` explicitly for installs that don't already have a saved mode or saved credentials, without disturbing the mode on an install that already has one saved.

chrisdhaan:
## Summary

Follow-up to #355: that PR's `--mode` option help text claimed a default of `local`, but nothing enforced it — omitting `--mode` never wrote a `mode` key to `config.json`, so the server's own fallback resolved the absence to `cloud` instead.

- `preflight install` now defaults to `mode: 'local'` when `--mode` is omitted and no mode is already saved — and only when there's no cloud credentials already saved either, so an existing non-interactive cloud install is never silently flipped to local the next time `install` runs for any other reason.
- 5 new tests covering the fresh-install default, preserving an existing saved mode, explicit `--mode` overriding existing state, and the regression case (existing credentials + no mode key + bare re-install leaves the mode untouched).
- CHANGELOG now documents `--mode` itself under `### Added` (this is the first release to ship it), alongside the `### Fixed` entry for the default-text fix.
- Version bumped 1.14.31 → 1.14.32 (patch).

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [x] `npm run lint`

Fixes #359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.32] - 2026-08-05
+
+### Added
+
+- **`preflight install` now accepts `--mode <cloud|local|both>`** — sets the telemetry mode non-interactively, matching what the interactive setup wizard already offers. Defaults to `local` for fresh installs; an install that already has a saved mode is never overwritten by omitting the flag.
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


corinnekunze:

PR #360, opened 2026-08-06
Fix small bugs: BudgetTracker UTC, event-send visibility, Node-version guard, cloud-mode docs

### LINKED ISSUE #320: Native Windows + cloud/both mode: no events are ever sent (NrIngestManager gated on !isProvisional; sync resolution lacks the cwd fallback) - metrics-only, fully silent

### LINKED ISSUE #324: Docs don't explain where cloud-mode data goes or how to view it
## Feedback (from a help-channel question)

> I am trying to use preflight in cloud mode pointed at us-production. How do I find the data? it looks like there should be a dashboard or app of some kind but I can't find it.
>
> The docs I could find don't really explain where the data goes in cloud mode or how to view it so that might need improved.

Cloud mode ships telemetry to New Relic as custom events (`AiToolCall`, `AiCodingTask`, `AiAntiPattern`, etc.), but nothing creates a dashboard automatically — that's a separate `preflight deploy-dashboards --all` step requiring a **User API key** (`NRAK-...`), which is distinct from the license key used for ingest.

The README's "Connect New Relic" section (lines ~111-139) does mention `deploy-dashboards`, but it's presented as one more setup step in a sequence, not called out as "cloud mode has no UI until you do this" — a user who's already run `preflight install --mode cloud` and is now looking for their data has no clear signal that a dashboard doesn't exist yet, or where else to look for the raw data in the meantime.

## Ask

Make it explicit, ideally right where cloud/both mode is introduced, that:
1. Cloud mode ships data with **no dashboard** until `deploy-dashboards` is run separately (and that it needs a *different* key — a User API key, not the license key).
2. Until then (or as an alternative), the raw data is queryable directly via NRQL, e.g. `SELECT * FROM AiToolCall SINCE 1 hour ago` in New Relic's Query Builder.
3. Where the deployed dashboards actually show up (New Relic account → Dashboards).

## Files likely involved

- `README.md` (the "Connect New Relic" section and the mode comparison table)
- `docs/ADVANCED.md` (if a more detailed cloud-mode walkthrough belongs there instead)

### LINKED ISSUE #356: BudgetTracker daily threshold test fails in UTC+ timezones (uses local date instead of UTC)
## Bug                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  `src/metrics/budget-tracker.test.ts` — "re-arms daily thresholds after midnight" fails when                                                                                                                                                        
  the test machine's local timezone is UTC+1 or later (confirmed in JST/UTC+9).                                                                                                                                                                      
                                                                                                                                                                                                                                                     
  ## Root cause                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                     
  `budget-tracker.ts` builds the daily period key using local-time methods:                                                                                                                                                                          
                                                                                                                                                                                                                                                     
  ```ts                                                                                                                                                                                                                                            
  const year  = now.getFullYear();   // local                                                                                                                                                                                                        
  const month = now.getMonth() + 1;  // local                                                                                                                                                                                                      
  const day   = now.getDate();       // local            
```                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  The test advances fake time to 2026-05-27T20:00:00Z expecting "still the same day" and                                                                                                                                                             
  asserting events.toHaveLength(1). In UTC+9 that timestamp is 2026-05-28T05:00:00+09:00                                                                                                                                                             
  — already the next day locally — so the threshold re-arms and fires a second event, giving                                                                                                                                                         
  toHaveLength(2).                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                     
### Suggested fix                                                                                                                                                                                                                                         
  Replace local-time getters with UTC equivalents in getPeriodKey():                                                                                                                                                                                 
                                                                                                                                                                                                                                                   
  ```ts                                                                                                                                                                                                                                            
  const year  = now.getUTCFullYear();                                                                                                                                                                                                                
  const month = String(now.getUTCMonth() + 1).padStart(2, '0');                                                                                                                                                                                      
  const day   = String(now.getUTCDate()).padStart(2, '0');                                                                                                                                                                                           
```                                                                                                                                                                                                                                                     
  The weekly key already uses Date.UTC(...) correctly — daily should match.                                                                                                                                                                          
                                                                                                                                                                                                                                                     
 ### Reproduction                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                   
  Run tests on a machine set to any UTC+ timezone:                                                                                                                                                                                                   
  TZ=Asia/Tokyo npx jest -- src/metrics/budget-tracker.test.ts                                                                                                                                                                                     
                                                                                                                                                                                                                                                     
 ### Affected test                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                     
  src/metrics/budget-tracker.test.ts › BudgetTracker › re-arms daily thresholds after midnight

chrisdhaan:
## Summary

- **Fixed** `BudgetTracker`'s daily and weekly period keys — both previously read local-time date components before computing their period identifier, so a threshold could re-arm a day (or ISO week) early in UTC+ timezones. Fixes #356.
- **Added** `event_send_status`/`consecutive_event_send_failures` to the `nr_observe_health` MCP tool, so a cross-account or misconfigured license key (events silently rejected while metrics keep flowing via a different transport) now has a visible signal.
- **Added** a runtime Node.js version check at MCP startup, so a too-old Node binary (most commonly a stale `nvm` default resolved by an MCP client) fails with a clear diagnostic instead of an opaque deep-stack error.
- **Documentation**: README now explains that cloud mode ships no dashboard until `deploy-dashboards` is run separately with a different credential, and shows the NRQL fallback for querying raw data in the meantime. Fixes #324.

Ships as v1.14.33 — see `CHANGELOG.md`.

## Test plan

- [x] `npm run build && npm test` — 162 suites passed, 5346 tests passed, 3 skipped (pre-existing, unrelated), 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.14.33] - 2026-08-05
+
+### Added
+
+- **`nr_observe_health` now reports whether telemetry event sends are succeeding** — new `event_send_status` (`ok`/`failing`) and `consecutive_event_send_failures` fields surface consecutive failures sending events to New Relic (for example, from a misconfigured or cross-account license key), which previously had no visible signal — metrics could keep flowing successfully over a different transport while events silently failed. Both fields are omitted when event-send health tracking isn't wired up for the current server mode.
+- **The MCP server now checks the running Node.js version at startup** and exits with a clear diagnostic instead of failing deep in the stack when a too-old Node binary is resolved — most commonly seen when an MCP client resolves a stale `nvm` default.
+
+### Fixed
+
+- **`BudgetTracker`'s daily and weekly period keys now use UTC consistently** — both previously read local-time date components before computing their period identifier, so in UTC+ timezones a budget threshold could re-arm a full day (or, for the weekly key, a full ISO week) early.
+
+### Changed
+
+- **README now explains what happens after enabling cloud mode** — cloud mode ships telemetry immediately, but no dashboard exists until `deploy-dashboards` is run separately with a different credential (a user API key, not the license key). The README now calls this out explicitly and shows how to query the raw data via NRQL in the meantime.
+
 ## [1.14.32] - 2026-08-05
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -136,6 +136,14 @@ NEW_RELIC_API_KEY=NRAK-... NEW_RELIC_ACCOUNT_ID=12345 \
 
 You'll need a **license key** (telemetry ingest) and your **account ID**, plus a **user API key** (`NRAK-…`) to deploy dashboards and alerts. See [ADVANCED.md](docs/ADVANCED.md) for alerts, OTLP export to other backends, and Terraform.
 
+> **No dashboard until you run `deploy-dashboards`.** Cloud mode ships telemetry to New Relic as soon as it's configured, but nothing creates a dashboard automatically — that's the separate step above, and it needs a **different** credential (a user API key, not your license key). Until you run it, there's no UI to look at. In the meantime, query the raw events directly in New Relic's **Query Builder**, e.g.:
+>
+> ```sql
+> SELECT * FROM AiToolCall SINCE 1 hour ago
+> ```
+>
+> Once `deploy-dashboards` succeeds, find the dashboards under your New Relic account's **Dashboards** section.
+
 > **Data ingest note:** Telemetry sent to New Relic counts against your account's data ingest. On paid plans, standard ingest rates apply. Monitor your usage under **NR One → Data Management → Data Ingestion**.
 
 ---
--- a/docs/COMMANDS_TABLE.md
+++ b/docs/COMMANDS_TABLE.md
@@ -30,13 +30,15 @@ Check server health and connection status.
   "connected_at": "2026-06-03T10:00:00.000Z",
   "uptime_seconds": 3600,
   "hooks_installed": true,
-  "setup_required": false
+  "setup_required": false,
+  "event_send_status": "ok",
+  "consecutive_event_send_failures": 0
 }
 ```
 
 **Data source:** Server startup metadata
 
-**How it works:** Returns the current server version, the resolved developer name, how long the server has been running (`uptime_seconds`), the current session ID, and the ISO timestamp of when the MCP connection was established. Use this to confirm the MCP server is responsive and to verify the expected developer identity is being used. When a hook-detection function is available (e.g. in a Claude Code environment), also reports `hooks_installed` and `setup_required` so the caller can detect an incomplete setup — for example, right after a Smithery-driven install that only wired up the MCP server — and prompt to call `nr_observe_install_hooks`. Both fields are omitted when hook detection isn't wired up for the current server mode.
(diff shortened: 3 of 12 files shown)


corinnekunze:

PR #437, opened 2026-08-10
Feat: GitHub Copilot full-hooks support + token-exact cost

### LINKED ISSUE #436: GitHub Copilot: self-reported only; add full-hooks capture + token-exact cost
## Summary

The GitHub Copilot adapter is currently classified `self-reported`: it only ingests events pushed over HTTP to `localhost:9847` by a user-supplied VS Code extension, with no first-party capture. In practice that means a Copilot user gets almost nothing out of the box.

But VS Code Copilot has since shipped two capabilities that make first-party, high-fidelity capture possible:

1. **Native agent hooks** (`PreToolUse`/`PostToolUse`, plus 6 other lifecycle events) that fire on every built-in tool call and pass JSON on stdin - the same uniform `hook_event_name`/`tool_name`/`tool_input`/`session_id` envelope `collector-script.ts` already parses. VS Code even reads Claude-format hook files (`~/.claude/settings.json`) by default. Ref: <https://code.visualstudio.com/docs/copilot/customization/hooks>
2. **Per-request token + credit data** written to the workspace debug log (`workspaceStorage/<hash>/GitHub.copilot-chat/debug-logs/<sessionId>/main.jsonl`), keyed by the same session id the hooks deliver.

## Proposal

- Reclassify Copilot from `self-reported` to `full-hooks`, handling the two documented deltas from Claude Code's envelope: camelCase `tool_input` keys (`filePath`, `oldString`) and VS Code's own tool names (`create_file`, `replace_string_in_file`, `run_in_terminal`), sourced from `microsoft/vscode`'s `ToolName` enum. Keep the HTTP-push path as a legacy fallback.
- Add a usage watcher that tails the debug log and emits `mode:'token'` events, giving Copilot sessions token-exact cost - the analog of the existing Claude Code parent-transcript watcher.

## Judgment calls worth maintainer input

1. **Platform reclassification is a product decision** per `docs/ADAPTERS.md`. This changes Copilot's `visibilityLevel`, which downstream consumers (platform-comparison dashboard, weekly digest) use to caveat cross-platform comparisons.
2. **The debug-log format is not a stable VS Code API** (the hooks doc says so explicitly). It's the same stability tier as the Claude Code transcript format the parent/subagent watchers already depend on, and the watcher degrades to estimation-based cost (never a crash) if the schema drifts - but it's a real dependency choice.

## Status

Implementation is complete and dogfooded against a live New Relic account (events landing, `platform: copilot` confirmed, token-exact cost verified end-to-end). Happy to open the PR if the approach looks good, or adjust scope first.

Thetanner:
## Summary

Upgrades GitHub Copilot from `self-reported` to a first-class, `full-hooks` platform with token-exact cost, matching the fidelity Claude Code already gets.

- **Full-hooks capture**: VS Code Copilot agent hooks (`PreToolUse`/`PostToolUse`) send the uniform envelope `collector-script.ts` already parses. Handles the two documented deltas from Claude Code: camelCase `tool_input` keys and VS Code's own tool names (`COPILOT_TOOL_MAP`, sourced from the `ToolName` enum in `microsoft/vscode`). Copilot CLI's lowerCamelCase event names work via the existing case normalization. The legacy HTTP-push path is retained as a fallback.
- **Token-exact cost** (`CopilotUsageWatcher`): tails VS Code's per-session debug log (`workspaceStorage/<hash>/GitHub.copilot-chat/debug-logs/<sessionId>/main.jsonl`), whose `llm_request` records carry exact `inputTokens`/`outputTokens`/`cachedTokens` per request, and emits `mode:'token'` events into the same session buffer the hooks write to - the Copilot analog of `ParentTranscriptWatcher`. `responseId` is the dedup key; cursor/tailing mechanics mirror the existing watchers.
- **Cache-billing fix**: VS Code's `inputTokens` is cache-inclusive (verified against real logs), while `CostTracker` follows the Anthropic convention where input excludes cache reads. The watcher emits only the uncached remainder so cached tokens aren't double-billed.
- **`preflight local` dispatch fix**: the `local` subcommand wasn't routed through the top-level dispatcher, so `preflight local --clean` (the fix command `preflight doctor` recommends) errored with "unknown option". Now routed like the other subcommands.
- Adds `AGENTS.md` and a dogfooding `.github/hooks/preflight.json`.

## Judgment calls flagged for review

1. This changes Copilot's `visibilityLevel` (`self-reported` -> `full-hooks`), which `docs/ADAPTERS.md` notes is a product decision. See the linked issue for context.
2. `CopilotUsageWatcher` depends on VS Code's debug-log format, which VS Code's docs call not-a-stable-API. It's the same stability tier as the Claude Code transcript the existing watchers tail, and degrades to estimation-based cost (never a crash) on schema drift.

## Notes for the reviewer

- Every `COPILOT_TOOL_MAP` entry and setup instruction traces to Copilot's own docs or `microsoft/vscode` source, per the adapter sourcing rule; citations are in code comments.
- One commit (`Chore: update dependencies and format imports`) bumps a few transitive deps and reorders imports. Happy to split it out if you'd prefer a Copilot-only diff.
- The pre-existing `setup-wizard.test.ts` failure (it reads the developer's real `~/.newrelic-preflight/config.json`) is unrelated to this change and present on `main`.

## Validation

- TDD throughout (red -> green); new co-located tests for the adapter, collector deltas, and usage watcher.
- Dogfooded against a live New Relic account: events landing, `platform: copilot` confirmed via session summaries and process env, `event_send_status: ok`, token-exact cost verified end-to-end.

Closes #436


chrisdhaan:
### Code review

Following up on two items from an earlier deep-dive review — one confirmed as a real gap worth addressing in this PR, one an update to a Finding worth documenting/hardening.

**1. Cache-creation tokens are folded into base-rate `inputTokens` instead of being billed at the premium cache-creation rate — please fix in this PR**

`cacheCreationTokens` is hardcoded to `0`, and any real cache-write tokens end up counted as plain (non-premium) input tokens instead:

https://github.com/newrelic-experimental/preflight/blob/fc516456d0eebfacacb239b498d7921246baf030/src/hooks/copilot-usage-watcher.ts#L326-L339

VS Code's Claude-backed requests do set `cache_control` breakpoints and report real `cache_creation_input_tokens` upstream (confirmed against `microsoft/vscode`'s `messagesApi.ts`/`anthropicProvider.ts`), and this project's own pricing model already has a distinct premium rate for exactly this (`cacheCreationPerMTok` in `src/shared/pricing.ts`). Right now the debug-log schema doesn't appear to surface cache-creation separately from cache-read, so this may need mapping through whatever field VS Code exposes for it (or, if it genuinely isn't exposed, at least tracking it as a known estimation gap rather than silently absorbing it into the base rate). Net effect today is a small, consistent under-billing on any turn with a fresh cache write.

**2. The debug-log source citation and its (undocumented) prerequisite**

Verified independently: the file path and JSON schema you're relying on are real — traced to `microsoft/vscode`'s `extensions/copilot/` (the former `vscode-copilot-chat`, now merged and archived), specifically `chatDebugFileLoggerService.ts` and its `otel-data-flow.html` doc. Worth citing that directly instead of "verified against a real main.jsonl" going forward:

https://github.com/newrelic-experimental/preflight/blob/fc516456d0eebfacacb239b498d7921246baf030/src/hooks/copilot-usage-watcher.ts#L7-L20

The more important finding: this debug log is **off by default**, gated behind `github.copilot.chat.agentDebugLog.fileLogging.enabled` — a setting tagged `advanced`/`experimental` (hidden from the normal Settings UI), still labeled "(Preview)," and requiring a VS Code window reload after toggling. For most users this file will simply never exist, and `CopilotUsageWatcher` will silently observe nothing with no indication why:

https://github.com/newrelic-experimental/preflight/blob/fc516456d0eebfacacb239b498d7921246baf030/docs/ADAPTERS.md#L555-L559

Two asks:
- Document the required setting + reload step explicitly in the ADAPTERS.md section above (and any user-facing setup docs), so users know why they'd see zero Copilot cost data.
- Consider a lightweight runtime check that warns rather than silently reporting nothing when no debug-log directory ever appears — otherwise this reads as a broken integration instead of a missing prerequisite.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Thetanner:
Thanks for the thorough review. All three points are addressed in 438e504. I also rebased the branch onto current `main` (it was 5 commits behind).

**1. Cache-creation tokens.** Confirmed the under-billing you flagged, and dug into the schema before fixing it. VS Code's debug log exposes only a single cache-READ figure (`cachedTokens`) per `llm_request` record; there is no cache-creation / cache-write field at all (verified against a live `main.jsonl`, where `cachedTokens` tracks the previous request's input). So there is genuinely nothing to map it through. Taking your fallback: instead of silently absorbing it, it is now documented as a known, bounded estimation gap. Cache-write tokens on a fresh-cache turn are unavoidably billed at the base input rate rather than `cacheCreationPerMTok`, a small consistent under-bill. Called out both at the emit site and in `docs/ADAPTERS.md`, with a note to revisit if VS Code ever exposes the field.

**2a. Source citation.** Updated the header comment and `docs/ADAPTERS.md` to cite `chatDebugFileLoggerService.ts` / `otel-data-flow.html` in `microsoft/vscode` `extensions/copilot/`, instead of "verified against a real main.jsonl".

**2b. Off-by-default prerequisite.** Documented the required `github.copilot.chat.agentDebugLog.fileLogging.enabled` setting plus the window reload in `docs/ADAPTERS.md` (both the adapter reference and the setup steps). Also added the runtime check you suggested: when a VS Code workspaceStorage root exists but no `debug-logs` directory does, `CopilotUsageWatcher` logs a one-time warning naming the setting and surfaces `copilotDebugLoggingDisabled` in the dashboard observability-health snapshot, so a zero-cost session reads as a missing prerequisite rather than a broken integration.

chrisdhaan:
### Maintainer follow-up

Verified all three items from the earlier review are addressed — thanks, this looks solid.

While running the full suite locally to confirm, I found and pushed two small fixes on top of your last commit (maintainer-edit is enabled on this PR):

1. **`git-efficiency-tracker.test.ts`**: the new repo-attribution test hardcoded `expect(event?.repo).toBe('Thetanner/preflight')` — it only passed on your machine, since `RepoNameResolver` resolves against whatever `git remote` the test happens to run under. Switched it to a hermetic temp repo with a controlled fake remote.

2. **`GIT_OPTS` in `local-session-aggregator.ts`** (new in this PR): none of the `spawnSync('git', ...)` calls cleared `GIT_DIR`/`GIT_WORK_TREE`, so `-C <dir>` gets silently overridden whenever those env vars are set — notably inside git hook subprocesses (which is how I actually caught it: the test above passed standalone but failed under `husky`'s `pre-push` hook, which is exactly the environment where this would misattribute repo names in real usage too). Matches the existing defensive pattern already in `src/install/cli.ts`.

Both verified with a full `npm run build && npm test` pass, including a run with `GIT_DIR`/`GIT_WORK_TREE` explicitly set to simulate the hook environment. Pushed as two plain commits on top of yours (no rebase/force-push) — `96270fa8..e572c8a3`.

Once this lands, I'll follow up with the version bump + CHANGELOG entry in a separate commit against `main`.

🤖 Generated with [Claude Code](https://claude.ai/code)

chrisdhaan:
One more small maintainer push (`e572c8a3..56c1c3f4`, no rebase/force-push): `package-lock.json` had 10 `libc`-field removals on optional native binary entries (rolldown/tailwindcss bindings) and 3 transitive dev-dep bumps (`brace-expansion`, `js-yaml`, `nanoid`) with no corresponding `package.json` change — consistent with a different local npm version regenerating the lockfile rather than an intentional update. None of it is needed for this PR's own work, so restored it to keep the lockfile diff minimal. Verified with a full `npm run build && npm run lint && npm test` pass.

🤖 Generated with [Claude Code](https://claude.ai/code)

chrisdhaan:
Brought the branch up to date with `main` (`56c1c3f4..38f7d88f`) via a merge commit rather than a rebase — no history rewrite, just two commits caught up (cost-deep-dive forecast/Cost-by-Tool panel, activity-heatmap timezone fix). Clean merge, no conflicts. Verified with a full `npm run build && npm run lint && npm test` pass.

🤖 Generated with [Claude Code](https://claude.ai/code)

chrisdhaan:
Now that the branch is caught up with main, the version bump can safely land here without a conflict. Bumped to v1.16.0 (minor: new Copilot full-hooks capability) with a CHANGELOG entry (`38f7d88f..226ff817`). Verified with a full build/lint/test pass.

chrisdhaan:
Hi @Thetanner — following up on the fixes above: from our side, we believe this PR is ready to merge.

We noticed it's now marked as a draft — are you still actively working on something here, or would you like us to do a final review and merge? If it's the latter, marking it "Ready for review" would help move it forward.

Thanks again for the contribution — solid piece of work.

chrisdhaan:
Following up again, since it's now been a few more days with no response.

Brought this branch fully up to date with `main` (`226ff817..4e663132`) — main had shipped 4 patch releases (v1.15.1 through v1.15.4) since the last sync. Merge commit rather than a rebase, so nothing on the branch was rewritten. Resolved the resulting conflicts (version bump / CHANGELOG entries, and a couple of spots where main had since extracted shared helpers — `checkNodeVersion`, `isNewerVersion`/`fetchLatestNpmVersion` — that this branch still had inlined). Regenerated the lockfile cleanly.

Verified with a full build + lint + format + test pass: 168/169 suites, 5667/5670 tests (same one pre-existing unrelated skip as before) — no regressions.

We also did a full adversarial security pass over the complete diff given this introduces a new file-watcher that reads VS Code's Copilot Chat debug logs: no network calls beyond the existing New Relic/npm endpoints, no shell/command injection (all git calls use an argv array, never a shell string), no credential access beyond three specific documented env vars, zero new dependencies, and the debug-log watcher extracts only token counts/model/message-id — never the actual chat payload. Nothing concerning found.

From our side, this PR is ready to merge as-is. @Thetanner — since it's still marked draft, let us know if you'd like to mark it ready for review, or if you'd rather we just take it from here.

chrisdhaan:
No pressure at all if you're just away — but to set clear expectations: if we haven't heard back by **2026-08-31** (two weeks from today), we'll take over this PR ourselves (preserving your authorship on the commits) and merge it, since it's fully synced, tested, and reviewed clean on our end.

If you're still working on something here, or want more time, just let us know and we'll happily wait longer.

Thetanner:
Hi @chrisdhaan - first, apologies for the long silence, and thank you for the patience (and for the maintainer-side syncs, the security pass, and the version bump while I was heads-down). That was above and beyond and it's appreciated.

The reason for the delay: I went back to rework the piece that wasn't yet covered here. As it stood, this PR gave Copilot **Chat** (VS Code) full-hooks + token-exact cost, but not the **GitHub Copilot CLI / SDK agent-host protocol** - a genuinely separate runtime with its own session-id space (`~/.copilot/session-state/<id>/`, no `workspaceStorage`) and its own tool vocabulary (`bash`/`edit`/`grep` rather than VS Code's `run_in_terminal`/`replace_string_in_file`). Rather than land this and immediately follow up, I'd rather fold that support in now so Copilot arrives complete in one release.

I've pushed the additional work on top of your last commit (`4e66313..95ca752`, plain fast-forward, no rebase/force-push):

- **`copilot-sdk` platform adapter** - a distinct adapter for the Copilot SDK / agent-host runtime (the CLI is the confirmed host). Tool calls arrive via the host's own `PreToolUse`/`PostToolUse` hooks through the same uniform envelope `collector-script.ts` already parses. The tool-name map is sourced from GitHub's CLI command reference ("Tool availability values"), with introspection/messaging tools deliberately left unmapped - mirroring the existing `COPILOT_TOOL_MAP` precedent.
- **Token-exact cost via a Copilot SDK extension** (`copilot-sdk-extension/extension.mjs`) that subscribes to the SDK's `assistant.usage` event and emits the same `mode:'token'` buffer-line contract `CopilotUsageWatcher` already uses. Usage-only, so it never double-counts against the hooks. (SDK extensions must be plain JS, so it hand-mirrors the tested logic in `copilot-sdk-usage-mapper.ts` - noted in-code to keep the two in sync.)
- **Pricing overlay** (`copilot-pricing/pricing.json`) that gap-fills Copilot-visible model IDs the vendored table doesn't resolve. I re-cut this against current `main`: your sync had since added `claude-opus-5` and the `gpt-5.6` variants to `src/shared/pricing-data.ts` (with fuller values), so I trimmed those from the overlay to honor its own no-duplication contract; 8 genuinely-missing IDs remain.
- Full setup documented in the new Copilot SDK section of `docs/ADAPTERS.md`.

All green locally against current `main`: `npm run build && npm run lint && npm test` - 172 suites, 5719 tests passing (same one pre-existing unrelated skip).

Happy to split the SDK adapter into its own PR if you'd rather land the original scope of #437 as-is first - your call. Otherwise this is ready from my side.

chrisdhaan (approved):

PR #20, opened 2026-07-02
Chore: v1.0.9 — sync pricing tables 2026-07-01

chrisdhaan:
Pricing tables drift quietly — a model gets released, rates get adjusted, and cost estimates in your dashboard start drifting from reality. This PR brings the built-in pricing table up to date as of 2026-07-01.

### New models

| Model | Input | Output | Notes |
|---|---|---|---|
| `claude-fable-5` | \$10 / MTok | \$50 / MTok | |
| `claude-opus-4-8` | \$5 / MTok | \$25 / MTok | Replaces `claude-opus-4-7` as current-gen |
| `claude-sonnet-5` | \$2 / MTok | \$10 / MTok | Introductory price through Aug 31, 2026; standard \$3/\$15 after |
| `gemini-3.5-flash` | \$1.50 / MTok | \$9 / MTok | |
| `mistral-medium-latest` | \$1.50 / MTok | \$7.50 / MTok | |
| `anthropic.claude-sonnet-5` (Bedrock) | \$2 / MTok | \$10 / MTok | |
| `anthropic.claude-opus-4-8` (Bedrock) | \$5 / MTok | \$25 / MTok | |

### Corrected rates

| Model | Field | Was | Now |
|---|---|---|---|
| `mistral-large-latest` | input | \$2.00 | \$0.50 |
| `mistral-large-latest` | output | \$6.00 | \$1.50 |
| `mistral-small-latest` | input | \$0.10 | \$0.15 |
| `mistral-small-latest` | output | \$0.30 | \$0.60 |
| `codestral-latest` | input | \$0.20 | \$0.30 |
| `codestral-latest` | output | \$0.60 | \$0.90 |

### Alias update

`claude-opus-4` (family alias) now resolves to `claude-opus-4-8`. The older `claude-opus-4-7` entry is retained as a legacy record for historical cost backfill.

`gemini-3.5` alias added, resolves to `gemini-3.5-flash`.

## Impact

If you use any of the corrected Mistral models, historical cost estimates in your dashboard will remain as originally recorded; only new sessions pick up the updated rates. No data migration is needed.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — all tests passing
- [x] `npm run lint` — 0 errors/warnings
- [x] `npm run format:check` — clean

PR #445, opened 2026-08-14
Chore: add claude-opus-5 to the built-in pricing table

byson2562:
Claude Opus 5 shipped after the 2026-07-01 pricing sync (#20), so it has no entry in `DEFAULT_PRICING_TABLE`.

`resolveModelPricing()` logs `Unknown model, pricing not available` and returns `null`, so every cost surface reports **$0.00** for sessions on that model — Spend Today, Model Usage, Cost by Tool, Cost per Outcome, and the forecast chips. Token counts, tool-selection scores, and every other metric record normally, so the dashboard looks healthy while cost silently reads zero. Budget alerts (`budget.session` / `budget.daily` / `budget.weekly`) also never fire for these sessions.

Reproduce by running a session on `claude-opus-5` and opening the dashboard: `MODEL USAGE` shows `$0.00/1M tok`.

### New models

| Model | Input | Output | Notes |
|---|---|---|---|
| `claude-opus-5` | \$5 / MTok | \$25 / MTok | Cache read \$0.5, cache creation \$6.25, 1M context |
| `anthropic.claude-opus-5` (Bedrock) | \$5 / MTok | \$25 / MTok | No cache rates, per the existing Bedrock convention |

Rates are from Anthropic's public pricing page and match the existing `claude-opus-4-8` entry — Opus 5 is priced identically to Opus 4.8. Cache rates follow the same multipliers used throughout the table (read 0.1x input, creation 1.25x input for the 5-minute TTL).

### Verification comment

The header comment keeps its `2026-07-01` date for the other vendors and notes that only Anthropic was re-checked (2026-08-14) for these entries — I didn't re-verify Google, OpenAI, Cohere, Mistral, or the rest of the Bedrock table, so bumping the shared date would have overstated it.

## Notes

- No alias change. `claude-opus-4` still resolves to `claude-opus-4-8`; Opus 5 is a new family and matches exactly.
- Historical sessions already recorded at \$0 stay as recorded — only new sessions pick up the rates. Same behavior as #20.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — added a `claude-opus-5` case to `pricing-data.test.ts` alongside the existing per-model rate assertions; 5506 passing

One pre-existing failure in `src/dashboard/routes/api-handler.test.ts` (`does not inflate the next bucket when a session ends exactly at a bucket boundary`) is unrelated to this change — it reproduces on a clean `main` checkout when the full suite runs, and passes when that file is run on its own, so it looks order- or clock-dependent rather than caused by this PR.

### DIFF
--- a/src/shared/pricing-data.test.ts
+++ b/src/shared/pricing-data.test.ts
@@ -2,6 +2,17 @@ import { DEFAULT_PRICING_TABLE } from './pricing-data.js';
 
 describe('DEFAULT_PRICING_TABLE', () => {
   describe('Anthropic models', () => {
+    it('has claude-opus-5 (current gen) with correct rates', () => {
+      const p = DEFAULT_PRICING_TABLE['claude-opus-5'];
+      expect(p).toBeDefined();
+      expect(p.inputPerMTok).toBe(5);
+      expect(p.outputPerMTok).toBe(25);
+      expect(p.thinkingPerMTok).toBe(25);
+      expect(p.cacheReadPerMTok).toBe(0.5);
+      expect(p.cacheCreationPerMTok).toBe(6.25);
+      expect(p.contextWindow).toBe(1_000_000);
+    });
+
     it('has claude-opus-4-7 (current gen) with correct rates', () => {
       const p = DEFAULT_PRICING_TABLE['claude-opus-4-7'];
       expect(p).toBeDefined();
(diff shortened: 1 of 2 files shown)


chrisdhaan:
Closing in favor of #447, which supersedes this — it syncs the same `claude-opus-5` pricing entry (direct API + Bedrock) as part of a broader 6-vendor pricing audit, tracked in #448. That sync also adds Bedrock cache rates for `anthropic.claude-opus-5` that this PR left out (per the older "undocumented" convention), so merging this afterward would conflict on the same table key with less complete data.

Thanks for catching the `claude-opus-5` gap and for the clear repro — #448 credits the underlying bug this PR describes.

PR #471, opened 2026-08-20
Fix #464-#467: session-seeding, ppid-breadcrumb, platform-precedence, dedup-ring fixes

### LINKED ISSUE #464: SessionTracker has no restart-seeding, so its scalar counters under-report after a restart
Follow-up from review of PR #437 (Copilot full-hooks + token-exact cost).

That PR adds `seedFromPersisted()` to six trackers so a resumed session's pre-restart totals survive a process restart: `CostTracker`, `ModelUsageTracker`, `BudgetTracker`, `QualityProxyTracker`, `TaskDetector`, `EfficiencyScorer` (see `rehydrateTrackersIfResumed()` in `src/index.ts`). `SessionTracker` has no equivalent — its own scalar counters (`toolCallCount`, `linesAdded`, `linesRemoved`, `bashCommandCount`, `testRunCount`/`testPassCount`, `buildRunCount`/`buildPassCount`) are never seeded from a prior checkpoint.

`SessionStore.mergeSummaries()` merges these scalars by `Math.max(existing, incoming)`, an approach that's correct when the two writers see *overlapping* views of the same activity (the concurrent engine+dashboard case the merge was designed for) but under-reports when they're genuinely *disjoint* — exactly what pre-restart vs. post-restart activity is. Concretely: a session with 6 pre-restart tool calls and 2 post-restart tool calls (true total 8) persists as `toolCallCount: 6` (the max of the two), even though `toolBreakdown` (a map, correctly union-merged) recovers the true total of 8 across its keys.

**Impact is bounded, not a crash/data-loss bug:** every ratio-based consumer we checked (`prompt-feedback.ts`, `cost-per-outcome.ts`, `efficiency-score.ts`) computes per-key values over `toolCallCount`, and since a merged per-key value can never exceed the merged count, none of those ratios can exceed 1.0 because of this gap. The under-reporting only affects the raw scalar counts themselves (dashboard displays, any absolute-count-based analytics), not derived ratios.

**Suggested fix:** add a `SessionTracker.seedFromPersisted()` mirroring the pattern already used by the six seeded trackers, and call it from `rehydrateTrackersIfResumed()` in `src/index.ts` alongside the others.

Files likely involved: `src/metrics/session-tracker.ts`, `src/index.ts` (`rehydrateTrackersIfResumed()`), `src/storage/session-store.ts` (`mergeSummaries()`).

### LINKED ISSUE #465: Watcher cursor reset on repo-rename (project-dir switch) can exceed the 4096-entry global dedup ring for long-lived sessions
Follow-up from reviewing https://source.datanerd.us/cdehaan/preflight/pull/368 (not yet merged/synced to this repo).

That PR fixes a duplicate-session double-count: `ParentTranscriptWatcher` and `SubagentWatcher` byte-cursors are keyed by `sessionId` (and `agentId`) alone, but the same ID can exist under two `~/.claude/projects/<slug>/` dirs after a repo rename, with the old dir's copy left behind. The fix keeps only the newest-mtime copy, and — when the canonical copy for an ID switches project dirs mid-stream — resets that ID's persisted byte cursor to 0 rather than applying a stale offset to an unrelated file (the stale offset previously caused `processFile` to early-return and silently drop every post-rename turn: a total-data-loss bug, strictly worse than what this introduces).

The reset relies on `HookEventProcessor`'s in-memory dedup ring (`tokenDedup`/`subagentDedup` in `src/hooks/event-processor.ts`, keyed by `sessionId|messageId` / `agentId|messageId`) to avoid double-counting turns shared between the old file and the newly-canonical one. That ring is capped at 4096 entries **globally, across every session in the process** — sized for its original purpose (deduping a handful of re-delivered lines after a crash mid-write). A rename-triggered reset re-reads a project-dir switch's full shared history from byte 0; for a long-lived or heavily-resumed session (the linked PR cites one with ~680M cache-read tokens accumulated over weeks) whose pre-rename shared history exceeds 4096 turns, the ring evicts old entries before the re-read catches up, and the tail of that shared history gets double-counted.

**Impact is bounded:** the duplicated turns carry their own real (past) timestamps, so they land on that session's lifetime total and whatever historical day-bucket they actually belong to — not on "today," so it doesn't reproduce the dashboard bug the linked PR fixes. But it can inflate lifetime cost and a past day's `costByDayUsd` for the affected session, and only manifests on the rare rename+very-long-history combination.

**Possible fixes to consider:** scope the dedup ring per-session (or per session+day) instead of globally, size it to something closer to a realistic max transcript length, or track cursor-reset events explicitly rather than depending entirely on the shared ring.

Files likely involved: `src/hooks/event-processor.ts` (dedup ring), `src/hooks/parent-transcript-watcher.ts` / `src/hooks/subagent-watcher.ts` (cursor reset on project-dir switch).

### LINKED ISSUE #466: PPID breadcrumb has no staleness check — a recycled PID re-binds an MCP to a session_id from weeks earlier
On native Windows, an MCP process resolved its `session_id` to a session that had ended **16 days earlier**, in a different directory. The result is a session split: hook-path tool calls land on the correct id while MCP-path token/cost lands on the stale one.

## Observed (v1.15.6, native Windows, Claude Code Desktop)

| record | toolCallCount | cost | requestCount |
|---|---|---|---|
| `8ad7d98d` (real session, started 2026-08-19 18:00 JST in `…/vault`) | 88 audit events, **no daily session file for 08-19** | — | — |
| `fa319898` (session from **2026-08-03**, different cwd) | **0**, empty `toolBreakdown` | **$10.07** | 81 |

`active-fa319898….meta.json` has `cwd` = the vault directory and an mtime of 08-19 18:00 — i.e. the 08-03 session id was adopted by an MCP started on 08-19.

## Why the cwd breadcrumb isn't the culprit

`session-by-cwd/…vault-vault.txt` contains `8ad7d98d` — the *correct* id. Had resolution fallen through to the cwd breadcrumb, it would have returned the right session. So a higher-priority source returned the stale id. `CLAUDE_JOB_DIR` was not set, which leaves the PPID breadcrumb.

## Root cause: neither side validates breadcrumb age

- `resolveFromBreadcrumb()` only checks the content against `SESSION_ID_RE`. No mtime, no creation-time, no cross-check against the resolved session's own start time.
- `gcStaleBreadcrumbs()` defines stale purely as `!isPidAlive(pid)`. **Once a PID is recycled by the OS, the breadcrumb becomes permanently un-collectable** — the GC sees a live PID and keeps it forever.

This is not a rare race on Windows: the MCP's `process.ppid` is a short-lived `cmd.exe` wrapper (confirmed locally — `node.exe pid=16936 → parent cmd.exe pid=12144`), and short-lived processes are exactly the ones whose PIDs get recycled fastest.

The module doc comment says of the PPID breadcrumb: *"Precise: no cross-session collision risk when it resolves."* That holds only if PIDs are never reused.

## Two downstream symptoms, same root cause

(I was about to file these separately.)

1. A session record with cost/token data but `toolCallCount: 0` and empty `toolBreakdown`.
2. The genuinely-active session gets no daily session file for that date, even though its events are present in `audit/<date>.jsonl`.

## Suggested fix

Make the breadcrumb identify the process, not just the PID — e.g. write `<pid>:<process start time>` (or the session's own start timestamp) and require both to match on read. That makes recycled PIDs fail closed and lets the GC collect them.

A simpler stopgap: reject a breadcrumb whose mtime predates the reading MCP's own start time.

### LINKED ISSUE #467: Platform detection can misattribute a session to another installed agent (e.g. Copilot reported as Claude Code)
## Summary

When more than one supported agent is installed, Preflight can attribute a session to the wrong platform. In my case a GitHub Copilot session is reported as `claude-code` in the New Relic platform breakdown, even though the Preflight MCP server was explicitly registered with `MCP_CLIENT=copilot`.

## Environment

- Preflight v1.16.0
- Host: VS Code with GitHub Copilot, on a machine that also has Claude Code installed
- Preflight registered as an MCP server in VS Code's `mcp.json` with `env: { "MCP_CLIENT": "copilot" }`

## What happens

The session shows up under `platform: claude-code` in the dashboard instead of `copilot`, despite the explicit `MCP_CLIENT=copilot` in the server registration.

## Steps to reproduce

1. Have both Claude Code and VS Code + Copilot installed.
2. Launch VS Code from an environment where Claude Code has exported its ambient variables (notably `CLAUDE_CODE_VERSION`) into the process environment - these are inherited by every child process VS Code spawns, including the Preflight MCP server.
3. Register Preflight as an MCP server for Copilot with `MCP_CLIENT=copilot`.
4. Run a Copilot session and observe the platform attribution: it reports as `claude-code`.

## Root cause

Platform detection is first-match-wins by registration order, and `ClaudeCodeAdapter` is registered before `CopilotAdapter`:

- Detection loop: [`platform-registry.ts#L33-L41`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/platforms/platform-registry.ts#L33-L41)
- Registration order: [`platform-registry.ts#L65-L69`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/platforms/platform-registry.ts#L65-L69)

`ClaudeCodeAdapter.isSupported()` returns true on the presence of an ambient env var, treating it as equivalent to an explicit client selection:

- [`claude-code-adapter.ts#L60-L65`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/platforms/claude-code-adapter.ts#L60-L65)

Because `CLAUDE_CODE_VERSION` is inherited into the server's environment and `ClaudeCodeAdapter` is checked first, it matches and wins before the explicitly-configured `CopilotAdapter` (`MCP_CLIENT=copilot`) is ever evaluated:

- [`copilot-adapter.ts#L190-L191`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/platforms/copilot-adapter.ts#L190-L191)

The underlying issue is that an ambient, inherited environment variable (which can leak in from a sibling process that merely launched the host) is given the same weight as an explicit, per-server configuration signal - and registration order, rather than signal strength, decides the winner.

## Impact

Any user who has two supported agents installed can get silently misattributed sessions, with cost and activity landing under the wrong platform in the dashboards. It is not specific to Copilot/Claude Code - it is a general precedence problem that affects any pair where one agent exports ambient environment variables that another adapter also keys on. The failure is silent: nothing surfaces that the detected platform disagreed with the explicit configuration.

## Notes

I have a view on how I'd approach a fix, but I'm deliberately not prescribing one here - there are a few angles (precedence between explicit vs ambient signals, richer identity available at the protocol layer, detection timing, etc.) and I'd rather align on the approach with maintainers before anyone writes code. Happy to open a PR once there's a direction.

chrisdhaan:
## Summary
- Fixes #464: seed SessionTracker's toolCallCount/bashCommandCount on resume
- Fixes #465: scope tokenDedup/subagentDedup rings per-session/per-agent instead of one shared global ring
- Fixes #466: reject PPID breadcrumbs older than the resolving process's own start time (and refresh the breadcrumb's mtime on content-unchanged writes, so an actively-hooked session never falsely goes stale after a restart)
- Fixes #467: explicit MCP_CLIENT/NEW_RELIC_AI_PLATFORM now wins over an ambient signal in platform detection

Bumps version to 1.16.2 with a matching CHANGELOG entry.

## Test plan
- [x] `npm run build && npm test && npm run lint` all pass in this repo
- [x] New unit tests added per fix (session-tracker, session-resolver, platform-registry, event-processor, collector-script)

### DIFF
--- a/src/hooks/event-processor.ts
+++ b/src/hooks/event-processor.ts
@@ -133,6 +133,75 @@ const DEFAULT_POLL_INTERVAL_MS = 100;
 const DEFAULT_ORPHAN_TIMEOUT_MS = 60_000;
 const DEFAULT_MAX_PENDING = 2_000;
 
+/**
+ * Fixed-capacity FIFO dedup set for one session/agent's own event stream.
+ * Extracted out of a single global ring (see DedupRingRegistry) so one
+ * scope's dedup keys can never evict another's.
+ */
+class DedupRing {
+  private readonly seen = new Set<string>();
+  private readonly order: string[] = [];
+
+  constructor(private readonly capacity: number) {}
+
+  /** Returns true if `key` was already seen (and does nothing further); otherwise records it and returns false. */
+  hasAndAdd(key: string): boolean {
+    if (this.seen.has(key)) return true;
+    this.seen.add(key);
+    this.order.push(key);
+    if (this.order.length > this.capacity) {
+      const evicted = this.order.shift();
+      if (evicted !== undefined) this.seen.delete(evicted);
+    }
+    return false;
+  }
+}
+
+/**
+ * LRU-bounded registry of per-scope DedupRings, keyed by session/agent id —
+ * mirrors ContextTrackerRegistry's eviction pattern (src/metrics/context-tracker.ts).
+ * A scope that goes quiet for a while eventually gets evicted entirely, so a
+ * burst of short-lived scopes can't grow this registry without bound while a
+ * long-lived scope's own ring keeps its full per-scope capacity regardless of
+ * how many OTHER scopes are active.
+ *
+ * Exported (unlike DedupRing) so tests can exercise the maxScopes eviction
+ * path directly with a small value — mirroring how ContextTrackerRegistry's
+ * own `{ maxSessions }` option is tested — rather than only through
+ * HookEventProcessor's fixed production values.
+ */
+export class DedupRingRegistry {
+  private readonly rings = new Map<string, DedupRing>();
+
+  constructor(
+    private readonly maxScopes: number,
+    private readonly perScopeCapacity: number,
+  ) {}
+
+  /** Returns true if `dedupKey` was already seen within `scopeKey`'s own ring. */
+  hasAndAdd(scopeKey: string, dedupKey: string): boolean {
+    let ring = this.rings.get(scopeKey);
+    if (ring) {
+      // Move to end for LRU ordering.
+      this.rings.delete(scopeKey);
+      this.rings.set(scopeKey, ring);
+    } else {
+      if (this.rings.size >= this.maxScopes) {
+        const oldest = this.rings.keys().next().value;
+        if (oldest !== undefined) this.rings.delete(oldest);
+      }
+      ring = new DedupRing(this.perScopeCapacity);
+      this.rings.set(scopeKey, ring);
+    }
+    return ring.hasAndAdd(dedupKey);
+  }
+
+  /** Number of scopes currently tracked — test seam for eviction assertions. */
+  get scopeCount(): number {
+    return this.rings.size;
+  }
+}
+
 export class HookEventProcessor {
   private store: LocalStore;
   private readonly pollIntervalMs: number;
@@ -147,12 +216,13 @@ export class HookEventProcessor {
   private readonly onWorkflowRun: ((event: WorkflowRunEvent) => void) | null;
   private readonly platformAdapter: PlatformAdapter;
   /**
-   * Dedup ring of `(agentId, messageId)` for recent subagent turns. Cursor
-   * recovery may re-read a line after a crash mid-write; double-counting would
-   * be silent, so we drop on the seen-set (capped at 4096 to bound memory).
+   * Per-agent dedup rings for recent subagent turns (scoped by agentId, one
+   * DedupRing per agent, LRU-bounded across agents). Cursor recovery may
+   * re-read a line after a crash mid-write; double-counting would be silent,
+   * so we drop on the seen-set. Was previously one global ring shared across
+   * every agent — see DedupRingRegistry's doc comment for why that's wrong.
    */
-  private readonly subagentDedup = new Set<string>();
-  private readonly subagentDedupOrder: string[] = [];
+  private readonly subagentDedupRegistry = new DedupRingRegistry(50, 4096);
   /**
    * Dedup ring of `(workflowRunId, timestamp)` for recent workflow_run
    * events, mirroring subagentDedup above. This path only matters for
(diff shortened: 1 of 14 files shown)


src/hooks/event-processor.ts
```diff
@@ -133,6 +133,75 @@ const DEFAULT_POLL_INTERVAL_MS = 100;
 const DEFAULT_ORPHAN_TIMEOUT_MS = 60_000;
 const DEFAULT_MAX_PENDING = 2_000;
 
+/**
+ * Fixed-capacity FIFO dedup set for one session/agent's own event stream.
+ * Extracted out of a single global ring (see DedupRingRegistry) so one
+ * scope's dedup keys can never evict another's.
+ */
+class DedupRing {
```
corinnekunze, line 141: nit: I prefer to have classes broken out into separate files instead of all in one big file

corinnekunze (approved):

src/hooks/event-processor.ts
chrisdhaan: Good call — filed #472 to track splitting `DedupRing`/`DedupRingRegistry` out into their own file. Left as-is here since it's a fresh addition and the split is a low-risk mechanical follow-up rather than something that should block this fix.

Issue #460, opened 2026-08-18
RetryDetector flags normal work as thrashing: tools with no extractInputMeta case always score similarity 1.0 (PowerShell), and the dedupe key never dedupes


tswdsawada:
## Summary

`RetryDetector` reports thrashing for ordinary, successful, non-repeated work. In a session with **zero failed tool calls**, it emitted **17 thrashing alerts** and drove the dashboard's Compute Waste panel to `needs attention` with ~37,862 "wasted" tokens.

The root cause is that similarity is computed over the serialized `ToolCallRecord`, and for any tool that has no `case` in `extractInputMeta()` the record carries **no input-derived fields at all** — so every call of that tool serializes to the same string and scores similarity `1.0` against every other call. Three calls of such a tool inside a 5-call window are unconditionally flagged.

Three smaller defects compound it: the dedupe key never dedupes, the waste estimate counts the whole cost of every call in the group, and the `needs_attention` threshold is 2,000 tokens.

## Environment

- preflight **1.15.4**, Windows 11, Node.js 22, `mode: local`
- Claude Code, `PowerShell` and `Bash` tools

## Observed

`nr_observe_get_retry_alerts` from one session (abridged — 17 alerts total, all `Bash` or `PowerShell`):

| tool | occurrences | similarity | tokensWastedEstimate |
|---|---|---|---|
| PowerShell | 3 | **1.0** | 2,953 |
| PowerShell | 3 | **1.0** | 2,953 |
| PowerShell | 4 | **1.0** | 2,790 |
| PowerShell | 4 | **1.0** | 2,790 |
| Bash | 5 | 0.8024 | 2,667 |
| Bash | 4 | 0.8007 | 2,800 |
| Bash | 4 | 0.8007 | 2,800 |
| Bash | 3 | 0.8016 | 732 |

Meanwhile `nr_observe_get_session_stats` for the same session reports:

```json
{ "tool_calls": 25, "success_rate": 1, "failed_calls": 0 }
```

Nothing failed and nothing was retried. The PowerShell calls scoring `1.0` were, in order, a process listing, an HTTP probe of the local dashboard, and a log aggregation — no textual resemblance whatsoever.

`nr_observe_get_compute_waste`:

```json
{ "total_tokens_wasted": 29854, "retry_tokens_wasted": 29854, "anti_pattern_tokens_wasted": 0, "breakdown": [], "status": "needs_attention" }
```

## Root cause

[...]

PR #474, opened 2026-08-21
Fix #462, #470, #473: Copilot SDK health signal, MCP PR tracking, retry false positives (v1.16.3)

### LINKED ISSUE #462: Copilot SDK usage extension: no health-flag/dashboard signal when it silently fails
Follow-up from the review of PR #437's Copilot SDK adapter + `assistant.usage` extension work.

`copilot-sdk-extension/extension.mjs`'s `joinSession()`/write failures now get written to stderr (fixed in a follow-up commit on #437), which the CLI captures into `~/.copilot/logs/extensions/`. That's better than fully silent, but it's still not visible anywhere a user would normally look.

The sibling VS Code integration (`CopilotUsageWatcher`) has a real health flag for the equivalent problem — `copilotDebugLoggingDisabled`, surfaced via `getHealth().debugLoggingLikelyDisabled` and shown in the dashboard/API (see `src/index.ts` and `src/dashboard/routes/api-handler.ts`). The Copilot SDK extension has no equivalent: if the extension was never copied, `--experimental` isn't set, `joinSession()` is version-incompatible with the installed CLI, or the storage dir isn't writable, a session will just show $0.00 token-exact cost with no indication of *why* — same failure mode already fixed once for the VS Code platform, now open again for this one.

**Suggested fix:** add a `copilotSdkExtensionMissing`-style health signal (or similar), surfaced the same way `copilotDebugLoggingDisabled` is, so a broken/missing extension setup is distinguishable from "extension genuinely not installed by choice" in the dashboard/API rather than only in CLI logs.

Files likely involved: `src/hooks/copilot-sdk-usage-mapper.ts`, `src/index.ts`, `src/dashboard/routes/api-handler.ts`, `copilot-sdk-extension/extension.mjs`.

### LINKED ISSUE #470: PR tracking only counts `gh` CLI commands; PRs opened via the GitHub MCP tools are captured but never counted
## Summary

The Git Efficiency "PRs Opened" metric only registers pull requests created through the `gh` CLI (`gh pr create` etc.). When an agent opens a PR via the GitHub MCP server's tools (`create_pull_request`, `update_pull_request`, ...) - now a very common path - the tool call **is captured and stored by Preflight**, but the PR metric ignores it and reports zero. The signal is present in the pipeline and simply not consumed.

This is not host-specific: it affects any platform whose agent uses the GitHub MCP rather than shelling out to `gh`, including Claude Code and Copilot.

## Evidence (from a live New Relic account)

Over a 7-day window on one account, PR creation happened through **both** paths, and only the CLI path was counted:

| Tool call | Event | Count | Counted by PR metric? |
|---|---|---|---|
| `create_pull_request` (GitHub MCP) | `AiToolCall` | 42 | No |
| `update_pull_request` (GitHub MCP) | `AiToolCall` | 31 | No |
| `gh pr create` (shell) | `AiToolCall` (Bash/run_in_terminal) | 27 | Yes |

So 42 of 69 PR-creation events (~61%) were silently dropped from the metric purely because they came through the MCP tool instead of the CLI. The `create_pull_request` calls are stored as first-class `AiToolCall` events with `tool = "create_pull_request"` - Preflight already has them.

## Root cause

PR detection lives entirely in `GitEfficiencyTracker.recordToolCall()` and keys **only** off the shell command string:

- The record is counted, then the method returns early for any tool call that has no `command` field - which is every structured MCP/native tool call: [`git-efficiency-tracker.ts#L376`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/metrics/git-efficiency-tracker.ts#L376) then [`#L398`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/metrics/git-efficiency-tracker.ts#L398)
- PR events are only produced by regex-matching `gh pr ...` in the surviving command string: [`#L410-L413`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/metrics/git-efficiency-tracker.ts#L410-L413), [`processGhCommand` #L441](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/metrics/git-efficiency-tracker.ts#L441), regexes at [`#L74-L80`](https://github.com/newrelic-experimental/preflight/blob/a66f8e9035a51dc38b53b12f3eb21813e2382160/src/metrics/git-efficiency-tracker.ts#L74-L80)

The tracker never inspects `record.toolName`, so a captured `create_pull_request` / `update_pull_request` MCP call (or a host's native PR tool) contributes nothing. MCP tool calls are captured elsewhere in the pipeline (the collector's `beforemcpexecution`/`aftermcpexecution` handling and the proxy's `AiMcpToolCall` path), so the data reaches Preflight - it just isn't wired into this metric.

Note the asymmetry with commits: commits have a second capture path (`git log` hydration) that backfills activity made outside tool calls, so they show up regardless of how they were made. PRs have no equivalent fallback and a single CLI-only detection path, so anything not run through `gh` is lost.

## Impact

"PRs Opened" (and the derived PR KPIs) systematically undercount for any developer or team whose agents use the GitHub MCP to manage pull requests - which is an increasingly common default. The undercount is silent and can be large (61% in the sample above). Because the underlying tool calls are already captured, the metric understates activity that Preflight can already see.

## Notes

As with #467, I have a view on how I'd fix this but I'm not prescribing an approach here - there are a few angles (keying PR events off recognized tool names as well as `gh` strings, a hydration-style fallback, etc.) and I'd rather align with maintainers first. Happy to open a PR once there's a direction. I can also share the exact NRQL used for the evidence table if useful.

### LINKED ISSUE #473: RetryDetector still flags distinct, successful tool calls as thrashing: serializeInput() compares a mostly-constant metadata envelope, and inputHash is compared with Levenshtein
Follow-up to #460 (closed by #461, shipped in v1.15.6). **#461 did fix the two defects #460 named** — tools with no `extractInputMeta` case no longer score a constant 1.0, and the dedupe key now actually dedupes. Thank you.

The false positive itself, however, is still present in **v1.16.2**, with a different root cause. Filing separately rather than reopening #460, since that issue's title no longer describes the defect.

### Reproduction (v1.16.2, native Windows 11, Claude Code)

I issued 9 PowerShell calls and 25 Bash calls in one session. Every call had different content, every call succeeded, and none was a retry of another. Verbatim `nr_observe_get_retry_alerts` output from that session:

```json
{
  "alerts": [
    { "toolName": "Bash",       "occurrences": 3, "windowSize": 5, "similarity": 0.8038174772900417, "tokensWastedEstimate": 2015 },
    { "toolName": "PowerShell", "occurrences": 3, "windowSize": 5, "similarity": 0.8370439505596713, "tokensWastedEstimate": 85 },
    { "toolName": "PowerShell", "occurrences": 4, "windowSize": 5, "similarity": 0.827619943647968,  "tokensWastedEstimate": 133 },
    { "toolName": "PowerShell", "occurrences": 3, "windowSize": 5, "similarity": 0.8116071428571429, "tokensWastedEstimate": 88 }
  ],
  "totalTokensWasted": 2321,
  "totalAlertsEmitted": 4
}
```

The PowerShell calls it grouped were, in order:

```
Get-Date -Format o
$PSVersionTable.PSVersion.ToString()
(Get-ChildItem -Path $PWD -File | Measure-Object).Count
[System.Environment]::OSVersion.VersionString
Get-Process -Name node -ErrorAction SilentlyContinue | Measure-Object | Select-Object -ExpandProperty Count
[math]::Round((Get-PSDrive C).Free / 1GB, 2)
git -C $PWD rev-parse --short HEAD
$env:USERNAME + " / " + $env:COMPUTERNAME
```

The flagged Bash group was 4 calls with 4 different `inputHash` values (`a64f32dc6d81b50b`, `34a9463a4a9d9a8e`, `55046d9c9818816d`, `11939a52350a351b`), all successful, reading different files. Replaying the session's real hook records through the shipped `RetryDetector` confirms `allFailed === false` in every case, so these fire purely on the similarity path.

`tokensWastedEstimate: 2015` for the Bash group is reported as waste for work that had no duplication in it at all.

### Root cause

`serializeInput()` strips per-call metadata, then Levenshtein-compares whatever is left. Here is an actual serialized record from the run above:

```json
{"toolName":"PowerShell","inputHash":"a201daf703bdbc63","cwd":"C:\Users\<user>\preflight_test","transcriptPath":"C:\Users\<user>\.claude\projects\C--Users-<user>-preflight-test\d5be26f9-….jsonl","permissionMode":"auto","command":"pwsh -File \"$env:USERPROFILE\.claude\scripts\preflight-sync.ps1\"","isTestCommand":false,"isBuildCommand":false,"isLintCommand":false,"bashCategory":"shell-other","bashLeading":"pwsh","bashDestructive":false,"bashNetwork":false,"commandDescription":"Run preflight sync script","commandTimeout":600000}
```

Almost none of that varies between two genuinely different calls:

1. **Constant environment fields are never stripped.** `handlePostEvent()` in `event-processor.ts` spreads `cwd`, `transcriptPath` and `permissionMode` into the record, and `serializeInput()` keeps all three. `transcriptPath` alone is a ~110-character string that is byte-identical for every call in a session. These are pure similarity padding.
2. **The classifier fields are near-constant too.** `isTestCommand` / `isBuildCommand` / `isLintCommand` / `bashCategory` / `bashDestructive` / `bashNetwork` are identical for any two ordinary shell commands.
3. **`inputHash` is compared with Levenshtein.** A hash carries information only as equality or inequality. Two unrelated 16-hex-char hashes still share roughly 1/16 of their characters by chance, so this contributes a *positive* similarity floor to calls that have nothing in common.

The variable content — the command text and its description — is a minority of the string. The result is that unrelated calls land in a narrow band just around the 0.8 threshold: measured 0.80–0.84 here, 0.79–0.85 in my earlier v1.15.6 measurements. The detector is not distinguishing repeats from non-repeats; it is measuring how much boilerplate two records share.

Raising `DEFAULT_SIMILARITY_THRESHOLD` would not fix this. The distribution is compressed near the threshold by construction, so any cutoff is arbitrary and genuine retries would be lost along with the noise.

### Suggested fixes

1. **Compare the actual input, not the record envelope.** Run the similarity over the raw tool input (the `command` / `query` / `file_path` string) instead of the serialized record. That makes the score mean what its name says.
2. **Use `inputHash` as an equality signal.** Exact hash matches are exactly what a genuine retry produces. Counting `inputHash` collisions within the window is both cheaper than Levenshtein and strictly more accurate — a real repeat hashes identically, and unrelated calls never collide. If this replaces the fuzzy path entirely, items 1 and 3 become moot.
3. **If fuzzy matching is kept, strip the constants** — at minimum `cwd`, `transcriptPath`, `permissionMode`, and the boolean classifier fields — and remove `inputHash` from the fuzzy comparison.

### Secondary (also from #460, still open)

`OUTPUT_PARSERS` in `hooks/tool-parsers.js` has entries for `Bash`, `Edit`, `Grep`, `Agent` only. `PowerShell` is absent, so `exitCode` is never recorded for PowerShell calls and the `allFailed` branch cannot fire for them on Windows. Adding `PowerShell: <the Bash output parser>` would cover this, matching the `PowerShell: parseBash` input-side entry #461 already added.

### Environment

- Preflight v1.16.2 (also reproduced on v1.16.0 and v1.15.6)
- Windows 11 native, Node.js v22, Claude Code, `mode: local`

chrisdhaan:
## Summary
- Adds a `copilotSdkExtensionMissing` observability-health signal for a GitHub Copilot SDK usage-capture extension that was never installed at its documented path — mirrors the existing signal for the VS Code Copilot Chat integration's equivalent gap.
- Counts pull requests created or edited through the GitHub MCP server's tools (`create_pull_request`, `update_pull_request`), not just the `gh` CLI, in `nr_observe_get_git_efficiency`'s PR metrics.
- Fixes `nr_observe_get_retry_alerts` flagging genuinely distinct, successful tool calls as thrashing by excluding near-constant session metadata (working directory, transcript path, permission mode, and classifier fields) from the similarity comparison, and scoping the detection window by session as well as tool name.

Fixes #462, fixes #470, fixes #473.

## Test plan
- [x] New unit tests added per fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.16.3] - 2026-08-21
+
+### Added
+
+- **A GitHub Copilot SDK usage-capture extension that was never installed at its documented path is now reported via the observability-health API** instead of failing silently to a log file no one normally checks. Mirrors the existing signal for the VS Code Copilot Chat integration's equivalent gap.
+
+### Fixed
+
+- **`nr_observe_get_git_efficiency`'s PR metrics no longer undercount pull requests created or edited through the GitHub MCP server's tools** (`create_pull_request`, `update_pull_request`) rather than the `gh` CLI. These tool calls were already captured by Preflight but were previously ignored by the PR metric entirely.
+- **`nr_observe_get_retry_alerts` no longer flags genuinely distinct, successful tool calls as thrashing when they share the same working directory, transcript, and permission mode.** Session-constant metadata fields were previously compared for similarity alongside the actual command content, pulling unrelated calls into a false-positive band just above the detection threshold.
+
 ## [1.16.2] - 2026-08-20
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.16.2",
+  "version": "1.16.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.16.2",
+      "version": "1.16.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.16.2",
+  "version": "1.16.3",
   "type": "module",
   "license": "Apache-2.0",
   "private": false,
--- a/src/dashboard/routes/api-handler.ts
+++ b/src/dashboard/routes/api-handler.ts
@@ -318,6 +318,15 @@ export interface ObservabilityHealthSnapshot {
    * Optional/absent for non-Copilot deployments and older snapshot producers.
    */
   readonly copilotDebugLoggingDisabled?: boolean;
+  /**
+   * True when the active platform is the GitHub Copilot SDK/CLI runtime and
+   * its optional token-exact-cost extension isn't at its documented install
+   * path (~/.copilot/extensions/preflight/extension.mjs) — the extension was
+   * likely never copied. Doesn't catch every failure cause (see
+   * copilot-sdk-extension-health.ts's doc comment). Optional/absent for
+   * non-copilot-sdk deployments and older snapshot producers.
+   */
+  readonly copilotSdkExtensionMissing?: boolean;
 }
 
 export interface ApiHandlerDeps {
--- a/src/hooks/copilot-sdk-extension-health.test.ts
+++ b/src/hooks/copilot-sdk-extension-health.test.ts
@@ -1,0 +1,46 @@
+import { existsSync } from 'node:fs';
+import { join } from 'node:path';
+import {
+  copilotSdkExtensionInstallPath,
+  isCopilotSdkExtensionMissing,
+} from './copilot-sdk-extension-health.js';
+
+jest.mock('node:fs', () => {
+  const real = jest.requireActual<typeof import('node:fs')>('node:fs');
+  return { ...real, existsSync: jest.fn() };
+});
+
+const mockExistsSync = existsSync as jest.Mock;
+
+afterEach(() => mockExistsSync.mockReset());
+
+describe('isCopilotSdkExtensionMissing()', () => {
+  it('is false for a non-copilot-sdk platform regardless of file presence', () => {
+    mockExistsSync.mockReturnValue(false);
+    expect(isCopilotSdkExtensionMissing('claude-code')).toBe(false);
+    expect(mockExistsSync).not.toHaveBeenCalled();
+  });
+
+  it('is true for copilot-sdk when the extension file is absent', () => {
+    mockExistsSync.mockReturnValue(false);
+    expect(isCopilotSdkExtensionMissing('copilot-sdk')).toBe(true);
+  });
+
+  it('is false for copilot-sdk when the extension file is present', () => {
+    mockExistsSync.mockReturnValue(true);
+    expect(isCopilotSdkExtensionMissing('copilot-sdk')).toBe(false);
+  });
+
+  it('checks the documented install path', () => {
+    mockExistsSync.mockReturnValue(true);
+    isCopilotSdkExtensionMissing('copilot-sdk');
+    expect(mockExistsSync).toHaveBeenCalledWith(copilotSdkExtensionInstallPath());
+  });
+});
+
+describe('copilotSdkExtensionInstallPath()', () => {
+  it('resolves to the documented .copilot/extensions/preflight/extension.mjs path', () => {
+    const path = copilotSdkExtensionInstallPath();
+    expect(path.endsWith(join('.copilot', 'extensions', 'preflight', 'extension.mjs'))).toBe(true);
+  });
+});
--- a/src/hooks/copilot-sdk-extension-health.ts
+++ b/src/hooks/copilot-sdk-extension-health.ts
@@ -1,0 +1,30 @@
+/**
+ * Health check for the bundled GitHub Copilot SDK usage-capture extension
+ * (`copilot-sdk-extension/extension.mjs`) — detects the one failure cause
+ * that's checkable from the MCP server process without a live-session
+ * signal: the extension was never copied to its documented install path (see
+ * `getHookInstallInstructions()` in `../platforms/copilot-sdk-adapter.js`).
+ * Deliberately does NOT catch the issue's other named causes (`--experimental`
+ * not set, a version-incompatible `joinSession()`, an unwritable storage
+ * dir) — none of those are filesystem-observable from this process. Same
+ * partial-coverage tradeoff already accepted by `CopilotUsageWatcher`'s
+ * `debugLoggingLikelyDisabled` (see that type's doc comment).
+ */
+
+import { existsSync } from 'node:fs';
+import { homedir } from 'node:os';
+import { join } from 'node:path';
+
+export function copilotSdkExtensionInstallPath(): string {
+  return join(homedir(), '.copilot', 'extensions', 'preflight', 'extension.mjs');
+}
+
+/**
+ * True only when `platformName` is `'copilot-sdk'` and the extension isn't
+ * at its documented install path. Always false for every other platform —
+ * the extension is copilot-sdk-only, so its absence there is meaningless.
+ */
+export function isCopilotSdkExtensionMissing(platformName: string): boolean {
+  if (platformName !== 'copilot-sdk') return false;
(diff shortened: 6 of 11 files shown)


corinnekunze (approved):

tswdsawada:
Thanks for turning this around so quickly — the strip list is the right call, and I can confirm it resolves the case I reported.

Two things worth adding before this lands. Neither is a Windows-specific ask: I checked, and the defect and the fix are both entirely platform-neutral.

## 1. The false positive is not Windows-specific

I framed #473 around a Windows session, which may have made it look narrower than it is. `RetryDetector` has no platform branching at all (the only `process.platform` in `src/metrics/` is the WSL path fixup in `transcript-reasoning.ts`), `inputHash` is `sha256(JSON.stringify(input)).slice(0,16)` from the shared collector on every OS and every adapter, and `INPUT_PARSERS` maps `PowerShell: parseBash`, so the two shells produce the same record shape.

Feeding six genuinely distinct, successful commands through the published v1.16.2 `RetryDetector`, once with Linux-shaped records (`Bash`, `/home/...` cwd and transcript path) and once with Windows-shaped ones (`PowerShell`, `C:\...`):

| record shape | v1.16.2 as published | with this PR's strip list |
|---|---|---|
| Linux / `Bash` | fires, similarity **0.8991 – 0.9100** | no alert |
| Windows / `PowerShell` | fires, similarity **0.9011 – 0.9118** | no alert |

Same defect, same magnitude, both platforms — and the strip list clears both. (Method: records hand-shaped to match each platform's hook envelope and run against the published v1.16.2 `dist`; the second column emulates this PR by omitting the newly-excluded fields rather than by building the branch. Happy to redo it against the actual branch if that is more useful.)

## 2. One case the strip list doesn't reach

After the exclusions, a tool with no `extractInputMeta()` case — any third-party MCP tool, so again OS-independent — serializes to just `{"toolName": ..., "inputHash": ...}`. The tool name dominates that string and only 16 hex characters vary, so distinct calls still land above the threshold. Six calls to the same MCP tool with genuinely different arguments:

```
pairwise similarity of distinct hashes: 0.857, 0.867, 0.857, 0.857, 0.857   (threshold 0.8)
-> fires at n=3 (0.8619), n=4 (0.8603), n=5 (0.8595), n=5 (0.8571)
```

So the near-constant envelope isn't the only thing that can carry a group over 0.8; a long tool name does it on its own.

## Suggestion

I understand and agree with the reasoning for keeping `inputHash` on the record — dropping it would reopen the `{toolName}`-only regression from #460. My suggestion doesn't require dropping it. It's about how it is *compared*: a hash is a two-valued signal (equal / not equal), and edit distance over it is meaningless in either direction — it dilutes a genuine repeat and manufactures similarity between unrelated calls, as above.

Keeping the field but counting it by equality gets both properties, with no new platform-specific code:

- if both records carry `inputHash`, its contribution to the pair's similarity is `1` when the hashes are equal and `0` when they differ, instead of a Levenshtein term;
- Levenshtein still applies to the remaining semantic fields (`command`, `commandDescription`, `filePath`, ...), so near-identical-but-not-identical retries are still caught.

On the same data as above, that yields no alert for the six distinct MCP calls, and still alerts on a genuine retry (three calls with identical arguments → identical hashes). It also makes the metadata-less-tool path strictly correct rather than incidentally safe: those records reduce to an exact-match check, which is exactly the semantics you want there.

PR #480, opened 2026-08-26
Fix self-referential comment wording

chrisdhaan:
## Summary

Comment referenced "where a per-connection counter caused" without saying what it was contrasted against — leftover from an earlier hygiene pass that stripped self-referential "fixes the issue" framing but left the sentence dangling. Rewrote to state the invariant directly. Matches the equivalent fix on the private GHE mirror.

## Test plan

- [x] `npm run build && npm test && npm run lint && npm run format:check`


chrisdhaan:
Folding this fix into the staging PR instead of keeping it standalone.

Issue #463, opened 2026-08-19
Verify Copilot SDK extension's sessionId matches the CLI hook's session_id


chrisdhaan:
Follow-up from the review of PR #437's Copilot SDK adapter + `assistant.usage` extension work.

The extension (`copilot-sdk-extension/extension.mjs`) keys its token buffer filename on `CopilotSession.sessionId` (from `joinSession()`), while tool-call events arrive separately via the CLI's `PreToolUse`/`PostToolUse` hooks carrying their own `session_id` field (see `src/platforms/copilot-sdk-adapter.ts`, `src/hooks/collector-script.ts`). Nothing in GitHub's public docs states these two identifiers are guaranteed to be the same value for a given session.

If they ever diverge, token-usage lines would land in a `buffer-<id>.jsonl` distinct from the one tool-call events use, and per the health-flag gap in a companion issue, that failure would currently be completely silent — cost just never attaches to the session, with no error anywhere.

**Suggested fix:** empirically confirm (via a live Copilot CLI session with both the hooks and the extension installed) that `session.sessionId` from `joinSession()` and the hook payload's `session_id` are identical for the same session. If confirmed, add a one-line note to that effect in `docs/ADAPTERS.md`'s Copilot SDK section. If they can diverge, that's a real bug needing a fix (e.g. reconciling the two id spaces) rather than just a doc note.

Files likely involved: `copilot-sdk-extension/extension.mjs`, `src/hooks/copilot-sdk-usage-mapper.ts`, `src/platforms/copilot-sdk-adapter.ts`, `docs/ADAPTERS.md`.

PR #483, opened 2026-08-27
Fix Copilot connection to New Relic

### LINKED ISSUE #538: Copilot (and other unscoped) sessions drained via --local can silently never reach New Relic, or arrive misattributed
## Problem

Sessions from platforms that don't run a dedicated owning `--stdio` engine — e.g. a Copilot chat with no per-session engine process — get drained by `--local`'s unscoped drain path (`LocalStore`/`LocalSessionAggregator`). That path has three separate correctness bugs that together mean this kind of session's activity can silently never reach New Relic, or arrives there misattributed:

1. **`--local` never sends to NR even when credentials are configured.** `--local` mode unconditionally forces `NR_AI_MODE=local` at startup (`src/index.ts`), so `NrIngestManager` is never constructed at all in this mode — regardless of whether `mode: "both"`/`"cloud"` and valid credentials are present in config. Any session only visible via `--local`'s drain is captured locally (visible in the dashboard) but never forwarded to New Relic.

2. **Drained sessions are misattributed to the wrong `session_id`.** `NrIngestManager` unconditionally overwrites every `AiToolCall`/`AiMcpToolCall` event's `session_id` (and the corresponding `ai.tool.*` metric dimension) with the ingesting process's own `sessionTraceId`, discarding the record's real `sessionId`. A session drained by `--local` (or by a proxy client's own connection) becomes unqueryable under its own ID in NR — its data exists, but filed under the draining process's synthetic session instead.

3. **Every hook-based tool call is tagged `platform: "claude-code"` regardless of actual origin.** `normalizeToolCall()` is implemented by every platform adapter (Copilot, Cursor, Windsurf, etc.) but is never actually invoked from the hook-write pipeline (`src/hooks/collector-script.ts`) — only from the generic-MCP self-report path. So hook-sourced events from any non-Claude-Code platform still get labeled as Claude Code data once they reach NR.

## Impact

For any org running Copilot (or another hooked platform) through `--local` without a dedicated owning engine per session, that activity can be invisible in New Relic entirely, or visible but mislabeled as a different session/platform than it actually is — with no error surfaced anywhere.

## Suggested fix

- Only force `NR_AI_MODE=local` as a fallback when cloud credentials are genuinely absent from config, not unconditionally, and start the harvest scheduler whenever real credentials are present.
- Prefer a `ToolCallRecord`'s own `sessionId` over the ingesting process's `sessionTraceId` when building `AiToolCall`/proxy events and `ai.tool.*` metrics; fall back to `sessionTraceId` only when the record has no `sessionId` of its own.
- Stamp the true originating platform at hook-write time in `collector-script.ts` (using that invocation's own environment, e.g. `MCP_CLIENT`), and thread it through every `ToolCallRecord` construction site in `event-processor.ts` instead of leaving it defaulted to Claude Code.

Files likely involved: `src/index.ts`, `src/transport/nr-ingest.ts`, `src/hooks/collector-script.ts`, `src/hooks/event-processor.ts`, `src/storage/types.ts`.

jcountsNR:
Closes #538

**Summary**

- --local dashboard mode unconditionally forced NR_AI_MODE=local at startup, so NrIngestManager was never constructed — any session captured only through the dashboard's unscoped drain (e.g. a Copilot chat with no dedicated owning --stdio engine) was visible locally but silently never sent to New Relic, even with valid mode: "both" + real credentials configured. Now it only forces local mode as a fallback when credentials are genuinely absent, and starts the harvest scheduler when they're present. Safe from duplicate sends — LocalStore.drainAllBuffers() already skips any buffer with a live owning heartbeat (hasLiveOwner()).
- Every AiToolCall/AiMcpToolCall event and ai.tool.* metric had its session_id unconditionally overwritten with the ingesting process's own sessionTraceId, discarding the record's real sessionId. This meant a session drained by --local (or a proxy client's own connection) became unqueryable under its own ID — its data existed in NR, just filed under the dashboard's synthetic local-<timestamp> session. Now prefers the record's own sessionId, falling back to sessionTraceId only when absent.
- Every hook-based tool call (Claude Code, Copilot, Cursor, etc.) was always tagged platform: 'claude-code' regardless of true origin — normalizeToolCall() is implemented by every platform adapter but was never actually invoked from the hook pipeline (only from the generic-mcp self-report path). Fixed by stamping the true platform at hook-write time in collector-script.ts (using that invocation's own environment, e.g. MCP_CLIENT), then threading it through every ToolCallRecord construction site in event-processor.ts.

**Files changed:** src/index.ts, src/transport/nr-ingest.ts, src/hooks/collector-script.ts, src/hooks/event-processor.ts, src/storage/types.ts, src/transport/nr-ingest.test.ts

**Known follow-up (not fixed here):** a long-lived Copilot CLI / VS Code Copilot Chat host still pins its MCP engine to whichever session existed at startup and never adopts later chats (pre-existing, documented in local-session-aggregator.ts). This PR makes those orphaned sessions actually reach NR, correctly attributed — it doesn't fix why they're orphaned in the first place.

**Test plan**

-  npm run build && npm test — 5821 passed, 3 skipped (pre-existing exclusions), no failures
-  npm run lint — 0 errors/warnings
-  npm run format:check — clean
-  Updated two existing tests in nr-ingest.test.ts that asserted the old (buggy) unconditional-sessionTraceId behavior; added one new test for the intentional fallback case
-  Verified live end-to-end against a real New Relic account: injected a synthetic Copilot-shaped hook event through --local and confirmed via NRQL it now arrives with platform: "copilot-sdk" and the correct real session_id (previously: never arrived at all)
-  User confirmed real Copilot activity now shows up correctly in New Relic


adjohn (commented):
Reviewed and runtime-tested this at head 2c29f88 (built dist, full suite, plus targeted collector/server experiments). Overall: two of the three fixes verify cleanly, one has a regression worth fixing before merge.

## What checks out

**`--local` cloud ingest.** `loadConfigOrDie` throws rather than exiting, so the new try/catch is live, and the fallback regex matches the real messages in `src/config.ts`. `isProvisional` is always false for `--local`, so the ingest manager is actually constructed and the new `nrIngest?.start()` is not a no-op. Runtime-verified both paths with a sandboxed HOME: no credentials falls back cleanly to local dashboard mode; with credentials present and `mode: both` it logs "Harvest scheduler started" and makes a genuine send attempt (403 on a 14-metric batch against a deliberately invalid key). The no-duplicate-sends claim also holds — `drainAllBuffers` skips buffers with a live owner heartbeat.

**session_id preference.** Correct, and the updated tests cover both directions. Full suite: 5862 passed, 3 skipped, 0 failed; lint clean. One non-blocking caveat for dashboard consumers: `AiToolCall`/`AiMcpToolCall` and `ai.tool.*` metrics now key to the record's real sessionId while session-level events (`AiCodingTask`, `AiContextSnapshot`) still key to the ingesting process's sessionTraceId. A `--local` drain therefore produces tool events under real session ids with no matching session-level rows. Strictly better than before, but NRQL joins should be aware.

## Regression: Claude Code hooks get tagged `generic-mcp`

The platform stamping in `collector-script.ts` works for the Copilot case (verified: with `MCP_CLIENT=copilot-sdk` the buffer record gets `platform: "copilot-sdk"`, and the base build provably writes no platform field). But I ran the built collector under a genuine Claude Code hook environment and the record comes out as `platform: "generic-mcp"`.

Cause: `ClaudeCodeAdapter.isSupported()` checks `CLAUDE_CODE` / `CLAUDE_CODE_VERSION` / `MCP_CLIENT=claude-code`, but Claude Code's actual hook environment sets `CLAUDECODE=1`, `CLAUDE_CODE_SESSION_ID`, and `CLAUDE_CODE_ENTRYPOINT` — none of the checked vars. `GenericMcpAdapter.isSupported()` returns true unconditionally, so the catch-all wins detection. Before this PR the field was absent and `nr-ingest.ts` defaulted it to `claude-code`, which was correct. After this PR, every Claude Code hook event is mis-attributed as `generic-mcp` in NR.

There's also a hot-path cost: the platforms import adds ~40ms per hook invocation (~50-60ms base vs ~90ms at PR head, five-run comparison), and hooks fire pre+post per tool call on every platform.

## Suggested change

`CopilotSdkAdapter` detection is explicit-env only (`MCP_CLIENT` / `NEW_RELIC_AI_PLATFORM`), so the registry buys nothing here over reading those vars directly. Replacing the registry call with:

```ts
const explicit = process.env.MCP_CLIENT ?? process.env.NEW_RELIC_AI_PLATFORM;
if (explicit) event.platform = explicit;
```

keeps the verified Copilot fix, removes the import cost, and leaves Claude Code hooks on the correct downstream default. If ambient detection is wanted later, `ClaudeCodeAdapter.isSupported()` should first learn the real env vars (`CLAUDECODE` / `CLAUDE_CODE_SESSION_ID` / `CLAUDE_CODE_ENTRYPOINT`).

chrisdhaan:
## Contributor PR triage

**Sense check:** passes — real bugs, in scope for the project. Minor nit: this PR's branch is named `main` on the fork rather than a feature branch per `CONTRIBUTING.md`'s convention (cosmetic, not a blocker).

**Tracked issue:** none existed for this specific set of fixes — opened #538 to track it (checked #463 first, which is a related but different Copilot session-id bug).

**Build/test/lint:** all clean — `npm run build` succeeded, `npm test` → 5862 passed / 3 skipped / 0 failed, `npm run lint` → 0 errors/warnings. (Branch was based on `v1.17.0`; verified against `main` after merging it in.)

**Review findings:**
- All three described fixes are real and correctly implemented — traced each against the actual functions changed (`toolCallToNrEvent`, `NrIngestManager.ingestToolCall`, `loadConfigOrDie`'s thrown error strings, `PlatformRegistry.getActive()`).
- `nit`: the `proxyToolCallToNrEvent` call site's `sessionTraceId: record.sessionId ?? this.sessionTraceId` is redundant — that function already prefers `record.sessionId` internally. Harmless.
- `suggestion`: `dashboardEnabled` (`src/index.ts`) is `config.mode === 'local' || config.mode === 'both'`. With this fix, launching with `--local` while `mode: 'cloud'` is explicitly configured now correctly sends to New Relic, but silently starts **no dashboard server** — `--local` no longer unconditionally guarantees the dashboard it used to force. Worth handling explicitly.
- Tests: the two modified assertions match the corrected behavior (not weakened), and a genuine new fallback-case test was added.
- No `src/shared/` edits.

**Version + CHANGELOG:** missing — will need a patch bump and a `Fixed` entry once merged.

🤖 Generated with [Claude Code](https://claude.ai/code)

chrisdhaan:
Confirmed and fixed — thanks for the runtime testing, this was a real one. Pushed `ac3eb50`: `collector-script.ts` now reads `MCP_CLIENT`/`NEW_RELIC_AI_PLATFORM` directly instead of going through `PlatformRegistry`, so it no longer depends on `ClaudeCodeAdapter.isSupported()` matching the real hook env (it doesn't — filed #539 for that separately, since it affects other ambient-detection call sites too). This also drops the registry-import cost you measured. Added tests in `collector-script.test.ts` covering both the explicit-platform and no-override cases.

Full suite (6032 passed) and lint clean after the change. Left your session-level-event caveat (`AiCodingTask`/`AiContextSnapshot` still keying to `sessionTraceId`) as a documented note rather than fixing here, to keep this PR scoped — happy to open a follow-up issue for it if you'd like.

chrisdhaan (approved):

PR #498, opened 2026-08-27
Fix #496: Fail closed on licenseKey without explicit mode; default local

### LINKED ISSUE #496: Config-level `mode` default is 'cloud', conflicting with the project's local-first positioning
## Problem
The README leads with "Local First — Offline by default" and "sends your data nowhere by default," and the setup wizard defaults to local mode. But the config loader's fallback is `'cloud'`: `src/config.ts:607` resolves `mode` to `(env) ?? (file.mode) ?? 'cloud'`.

#359 fixed one consequence (the `install --mode` help-text default not being enforced), but the loader-level default remains: any path that produces a config with `licenseKey`/`accountId` but no explicit `mode` — a hand-written `config.json`, CI/scripted setups, older configs predating the `mode` key — silently exports telemetry to New Relic. That's the opposite of the documented product default, and it's the kind of surprise PRIVACY.md's pre-cloud checklist exists to prevent.

## Proposed fix
Either flip the loader default to `'local'` (breaking for configs that relied on implicit cloud — worth a `configVersion` bump note), or require an explicit `mode` whenever a license key is present (fail closed with a clear doctor/startup message). Add a `preflight doctor` line surfacing the resolved mode and where it came from.

adjohn:
Fixes #496.

## Summary

- The config loader resolved `mode` as `env ?? file ?? 'cloud'`, so any config carrying a `licenseKey` but no explicit `mode` silently exported telemetry — the opposite of the README's "Local First — Offline by default" positioning.
- Resolution is now: explicit `NR_AI_MODE` → explicit file `mode` → **fail closed** if a licenseKey is present without either (startup error naming the exact fix) → `'local'` otherwise. Credential-less configs now run local instead of erroring for a missing licenseKey.
- **Breaking, deliberately loud:** configs with credentials but no `mode` key now error at startup with a self-describing message (`set "mode": "cloud" … or remove the credentials`) instead of silently exporting. The silent alternative — flipping the default — would have silently stopped existing implicit-cloud deployments' telemetry; an explicit error seemed safer in both directions.
- The setup wizard's `buildConfig()` always writes an explicit `mode` key now; it was the one writer producing exactly these ambiguous configs (`...(inputs.mode ? { mode } : {})`).
- `preflight doctor` gains a `Telemetry mode` check reporting the resolved mode and its source (`env NR_AI_MODE` / `config file` / `default (local)`). It calls the real `loadMcpConfig` so it cannot drift from the loader, and it surfaces the fail-closed error as a normal `fail` diagnostic instead of a stack trace.
- `example.config.js` default-mode comments updated. README/ADVANCED already described local-first and needed no change.
- Known related quirk left untouched (pre-existing, narrower): doctor's `NR reachable` check keeps its own internal `mode ?? 'cloud'` approximation for its skip heuristic, so a failing-mode config still attempts that connectivity check. Happy to align it in a follow-up.
- A middle path exists if the hard startup error feels too abrupt for existing implicit-cloud installs: resolve to `cloud` with a loud deprecation warning for one release, then fail closed. I went straight to fail-closed because a warning in stderr logs is easy to miss precisely in the unattended setups this bug bites, but the two-step variant is a small change if you prefer it.
- Test-suite note: `config.test.ts` gained `NR_AI_MODE='cloud'` in the shared `beforeEach` so ~100 pre-existing fixtures that set credentials without a mode keep exercising their original intent; the mode-resolution tests themselves explicitly delete or override it. Reviewers should know most of that suite now runs under an explicit env mode.

## Test plan

- [x] RED first: `defaults to 'local'` and both fail-closed throw tests observed failing against the old resolution (resolved `'cloud'`)
- [x] `config.test.ts` 220/220, `setup-wizard.test.ts` 108/108, `diagnostics.test.ts` 59/59, `cli.test.ts` 119/119
- [x] Doctor verified end-to-end against the built `dist/` with three real configs (fail-closed, explicit cloud, no-credentials local)
- [x] `npm run build && npm test` — 175 suites / 5871 tests pass; `npm run lint` and `prettier --check` clean


chrisdhaan:
## Triage summary

Sense-check: passes — this fixes a real conflict between the config loader's `mode ?? 'cloud'` default and the README's "Local First — Offline by default" positioning. Tracked to #496 (self-filed, verified legitimate). No duplicate PRs found.

**Verification (local, before my commit):**
- `npm run build` — clean
- `npm test` — 175/175 suites, 5871 tests pass
- `npm run lint` — clean

**Review notes:**
- Mode-resolution order (`env → file → fail-closed if licenseKey present → 'local'`) is correct and well covered by tests, including the escape hatches (explicit mode + licenseKey, `NR_AI_MODE=local` overriding fail-closed).
- `checkTelemetryMode`'s source attribution (env/file/default) reads the raw parsed config mode rather than the Zod-validated one, so it stays accurate even when unrelated fields fail schema validation — verified in code.
- `loadMcpConfig` has no filesystem side effects, so calling it from a diagnostic is safe.
- The doctor's separate `NR reachable` check still uses its own `mode ?? 'cloud'` approximation for its skip heuristic (pre-existing, unrelated to this PR) — already disclosed in the PR description as a known follow-up.

**Branch is 8 commits behind `origin/main`** (confirmed the underlying bug is still present on main, so this isn't a no-op) — will need a rebase before merge; not doing that here since it'd require a force-push to your fork.

## Pushed to this branch (additive commit, maintainer-can-modify)

Since this fix is a deliberate breaking change (documented in the PR description — configs that silently defaulted to `cloud` now hard-fail at startup), I bumped the version as a **minor**, not a patch, and added the CHANGELOG entry:

```
## [1.19.0] - 2026-08-31

### Fixed

- **Configs with New Relic credentials but no explicit `mode` now fail to start with a clear error instead of silently sending telemetry.** The config loader previously defaulted to `cloud` whenever a license key was present without an explicit `mode`, contradicting the project's local-first, offline-by-default default; it now requires an explicit `mode` in that case, and defaults to `local` when no credentials are configured. `preflight doctor` also reports the resolved telemetry mode and where it came from.
```

I also applied one small suggestion: `checkTelemetryMode`'s fail case can surface config errors unrelated to mode (invalid `accountId`, a literal `"null"` licenseKey) via the real `loadMcpConfig` call it makes — those weren't previously distinguished from the mode-specific fail-closed error, so the check's detail text now makes clear when it's not actually a mode problem. No behavior change for the fail-closed case itself (still `no explicit mode` verbatim), and all 59 diagnostics tests still pass.

Note: the version bump/changelog were applied on top of the current (stale) branch base, so this will likely need to be resolved as part of the rebase onto main rather than merged as-is.

chrisdhaan:
Rebased this onto current `main` as #541 rather than force-pushing here — this branch was 8 commits behind and had conflicts in `package.json`/`package-lock.json`/`CHANGELOG.md` once the version bump landed, and rewriting your branch's history felt like the wrong way to resolve that.

Your two commits are cherry-picked over unchanged, with your original authorship preserved. #541 also fixes a real interaction bug that only showed up after rebasing: the shared `beforeEach` mode override collided with the "homelab config fields" tests that landed on `main` after this branch forked (details in #541's description).

Happy to hand this back to you if you'd rather drive it from here — just let me know. Otherwise #541 is ready for review.

chrisdhaan:
Resolved the conflicts directly here instead, via a merge commit (`9eea921`) bringing in current `main` — no rewrite of your commits, just a merge on top. Also carried the version bump (1.19.0), CHANGELOG entry, and the diagnostics-check clarity fix from my earlier review comment, plus one thing that only surfaced after merging: the `beforeEach` NR_AI_MODE override needed to be scoped away from the "homelab config fields" tests (added to `main` after this branch forked), since they rely on their config file's `mode: 'local'" and env otherwise overrides it. Build/test/lint all pass. Ready for review.

chrisdhaan (approved):

PR #502, opened 2026-08-28
Feat #485: Classify rejections, denials, interrupts in tool telemetry

### LINKED ISSUE #485: User tool rejections export as error_type='timeout' — no edit-acceptance signal
## Problem
When the user rejects a tool call in the permission prompt, the `PreToolUse` hook has already fired but no `PostToolUse` follows. `HookEventProcessor.sweepOrphans()` flushes the orphan after 60s as `success:false, errorType:'timeout'` (`src/hooks/event-processor.ts:557-587`), so in NRDB **a deliberate user rejection is indistinguishable from a genuinely hung tool**.

Related losses in the same path:
- The collector captures `PostToolUseFailure`'s `is_interrupt` (`src/hooks/collector-script.ts:716-724`, typed at `src/storage/types.ts:35`), but `handlePostEvent()` never copies it onto the `ToolCallRecord`, so interrupt-vs-failure is dropped before export.
- `preflight install` registers only `PreToolUse`/`PostToolUse` (`src/install/install-helper.ts:88-90`) — the collector's `posttoolusefailure` branch exists but nothing ever subscribes to the event.

## Why it matters
Claude Code's OTel exports `claude_code.code_edit_tool.decision` (accept/reject, by language) — a headline "acceptance rate" ROI metric. Preflight has the raw ingredients and throws them away.

## Proposed fix
- Wire Claude Code's `PermissionRequest` / `PermissionDenied` hook events in the collector and installer.
- Introduce distinct `errorType` values (`rejected`, `interrupted`) vs. `timeout`; propagate `isInterrupt`.
- Derive an acceptance-rate metric for Edit/Write tools, with language from the already-captured `filePath` extension.

Known ceiling: OTel's decision-provenance values (`config`/`hook`/`user_permanent`/`user_temporary`/…) are not replicable from hooks; accept/reject is.

adjohn:
Fixes #485.

## Summary

- A user's tool rejection, an auto-mode policy denial, a mid-run interrupt, and a genuinely hung tool all exported as `error_type: 'timeout'`. They now carry distinct values: `rejected`, `denied`, `interrupted`, `timeout`.
- **One premise of the issue needed correcting against the hooks reference** (code.claude.com/docs/en/hooks.md): Claude Code fires no hook event for an interactive user rejection in default/plan modes. `PermissionDenied` covers only auto-mode policy denials. So rejection is inferred structurally: a pending call marked by a `PermissionRequest` event that never receives a post sweeps as `rejected`, while a bare orphan stays `timeout`. Permission-requested entries wait a longer TTL (300s vs 60s) since prompts legitimately dwell; an approval arriving later than that becomes the existing post-without-pre record shape.
- `PostToolUseFailure`'s already-captured `is_interrupt` now reaches the record: interrupted calls export `error_type: 'interrupted'` in both paired and orphaned-post paths.
- The installer registers `PermissionRequest`/`PermissionDenied` alongside the existing two, driven by a new single `HOOK_SUBCOMMANDS` table from which the uninstaller's recognition regex and the doctor's checks are derived — a hook event can no longer be registered without removal and diagnostics recognizing it by construction. Install, uninstall, headless install, doctor, and the two user-facing instruction surfaces all follow the shared vocabulary.
- Acceptance rate is deliberately NOT shipped as a pre-baked metric: with `error_type` on `AiToolCall`, it is a query-time ratio, e.g. `SELECT percentage(count(*), WHERE error_type != 'rejected') FROM AiToolCall WHERE tool IN ('Edit','Write') FACET developer` — the same rationale `CostTracker.emitMetrics` documents for not shipping derivable gauges.
- Review point: `areHooksInstalled` now requires all four hooks, so doctor reports `fail` (not `warn`) for a pre-upgrade install until `preflight install` runs again, and headless install upgrades instead of reporting already-installed. Consistent with the existing absent-hook severity, but note the fleet effect: every existing install reports `fail` after upgrading until re-installed. A one-line severity tweak if you prefer gentler.
- Three honest limitations worth knowing:
  - `rejected` reads "a permission prompt appeared and no completion ever arrived." A post lost for any other reason after a prompt (collector crash, buffer loss) carries the same signature and would also classify as `rejected`; a lost post with no prompt stays `timeout`, as today. Acceptance-rate consumers inherit that asymmetry.
  - `interrupted` activates on `PostToolUseFailure.is_interrupt`, a field this repo's collector already captured before this PR but which the current hooks reference does not document on that payload. If the field never arrives, the value is simply never emitted — nothing breaks, but the vocabulary entry may be latent until the payload carries it.
  - The collector now registers on `PermissionRequest`, a hook that CAN return permission decisions. Preflight's hook stays an observer by construction: it writes to its buffer, emits no decision JSON, and exits 0, so the permission flow is unaffected. Worth stating since it's the first decision-capable hook the installer touches.

## Test plan

- [x] RED→GREEN per behavior with fake timers: request→denied, request→sweep-as-`rejected` (survives 61s, flushes at 301s), bare orphan still `timeout` at 60s, late approval becomes post-without-pre, `stop()` flush classifies by phase, interrupt in paired and orphan paths, collector parse branches incl. missing-`tool_use_id` drop, installer/uninstaller symmetry, doctor pre-upgrade fail
- [x] Targeted suites 737/737; full `npm run build && npm test` — 5885 tests pass; lint 0/0; prettier clean


chrisdhaan (commented):
Nice work here — the fix is well-scoped, the phase-based pending-map redesign is clean, and the test coverage for the timing edge cases (bare orphan vs. permission-requested orphan TTLs, late-approval-after-sweep, stop() flush classification) is thorough. Build/test/lint all pass locally: 5999/5999 tests, 0 lint errors.

A few things worth addressing before merge:

**Stale branch / real merge conflicts.** This branch is 8 commits behind `main` (currently at 1.21.0), and a test merge produces real conflicts in `src/hooks/event-processor.ts` (3 hunks) and `docs/METRICS_TABLE.md` (1 hunk) — not just line-number noise. `main` independently added a `platform` field to the synthesized timeout/orphan records since this branch forked; this PR refactors that same code into a shared `emitUnpairedPreRecord()` helper, so a careless conflict resolution would silently drop `platform` from every future `timeout`/`rejected` record.

**Denied-record path drops some context fields.** `handlePermissionDeniedEvent`'s synthesized record only carries `inputSizeBytes`/`inputHash` from the pending pre-event, while the existing paired-post path (`handlePostEvent`) also carries `cwd`, `transcriptPath`, and `permissionMode`. `permissionMode` in particular seems directly relevant to a denial record — it's literally the setting that produced the denial — and it's already sitting on the pre-event, unused.

**Missing test for the "approved normally" path.** Every new test exercises `PermissionRequest` followed by rejection/denial/TTL-sweep, but none covers `PermissionRequest` → approved within the TTL → normal `PostToolUse`. That's likely the highest-volume real path for anyone running default (non-auto) permission mode, since most gated calls get approved, not rejected. By inspection `handlePostEvent` ignores `phase` so this should already work, but it's currently untested.

I have `maintainerCanModify` on this PR, so I'm going to merge current `main` into this branch (a merge commit, not a rebase — keeps your commit history intact) and push a couple of follow-up commits addressing the two findings above. You'll see them show up on this branch shortly.

Also confirmed and linked: this fixes #485.

chrisdhaan (approved):

Issue #491, opened 2026-08-27
OTel companion mode: prevent double-counting when Claude Code's built-in OTel is also enabled


adjohn:
## Problem
An org that enables both Claude Code's built-in OTel export (→ New Relic OTLP) and Preflight cloud mode ingests four signal families **twice**: cost (`claude_code.cost.usage` vs `ai.cost.session_total_usd` / `AiCodingTask.estimated_cost_usd`), tokens, lines of code, and session time. A blended "org AI spend" dashboard that sums across both roughly doubles the truth, and nothing on either side prevents it today.

The join key exists: for Claude Code sessions, Preflight's `session_id` and OTel's `session.id` are the same UUID — so joint deployments are joinable and de-duplicable, but only if dashboards know to do it.

## Proposed fix
- A companion-mode knob (e.g. `costSource: 'otel'` / `companionMode: true`) that, for `platform = 'claude-code'`, suppresses — or tags with e.g. `cost_authority: 'external'` for dashboard-side exclusion — the overlapping cost/token/LOC/duration signals while keeping everything Preflight-exclusive (tasks, efficiency, anti-patterns, audit, context, MCP proxy, per-repo attribution).
- Document canonical-source guidance for blended deployments (which pipeline owns which signal).
- Stretch: use the overlap as a feature — a reconciliation check between Preflight-computed and OTel-reported cost per session (the `token_reconciliation_delta` self-check on `AiWorkflowRun` is precedent).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #503, opened 2026-08-28
Feat #492: Decode and enrich protobuf OTLP payloads in the receiver

### LINKED ISSUE #492: OTLP receiver: decode and enrich protobuf payloads (enable ingesting Claude Code's own OTel)
## Problem
The inbound OTLP receiver enriches JSON-encoded payloads only; `application/x-protobuf` bodies are forwarded unmodified (`src/proxy/otlp-receiver.ts:359-360`). #146 fixed the docs to disclose this; this issue proposes actually implementing protobuf support — because it unlocks a significant architecture: **Preflight as the local OTel collector for AI coding tools.**

Claude Code exports OTel over gRPC/protobuf by default. With protobuf decode, an org could point `OTEL_EXPORTER_OTLP_ENDPOINT` at Preflight's receiver, and Preflight could:
- enrich Claude Code's own telemetry with what only Preflight knows (repo/`project_id`, task boundaries, session context),
- reconcile it against hook-observed data (and suppress its own overlapping cost/token signals — see the companion-mode issue),
- forward one unified, de-duplicated stream to New Relic.

That collapses the "two competing pipelines" problem into one pipeline with two capture paths, while keeping Claude Code's supported telemetry as the stable input for cost/outcome signals.

## Scope
- Protobuf (and ideally gRPC) decode for OTLP logs + metrics in the receiver path; re-encode or JSON-forward after enrichment.
- Recognize `claude_code.*` metrics/events and map session.id ↔ Preflight `session_id`.

adjohn:
Fixes #492.

## Summary

- The inbound OTLP receiver enriched JSON payloads only; `application/x-protobuf` bodies (what Claude Code and most OTel SDKs send by default) forwarded opaque and never gained session/repo attributes. The receiver now decodes protobuf export requests, runs the SAME `injectResourceAttributes` enrichment the JSON path uses, and re-encodes — for all three signals, routed by the OTLP/HTTP path (`/v1/logs|metrics|traces`).
- New runtime dependency `protobufjs@7.6.6` (server-side only — verified absent from the built web bundle; `check:bundle-size` still passes at 231.8 KB / 400 KB). Pinned to the widely-deployed 7.x line; 8.8.0 exists if you prefer the newer major.
- The OTLP schema ships as a vendored pbjs JSON descriptor (`src/proxy/otlp-descriptor.ts`) generated from `open-telemetry/opentelemetry-proto @ dfd0b0e`; the header carries the exact regeneration recipe (including deleting the gRPC `*Service` stubs, which this HTTP transcoder never uses and which don't type-check against protobufjs's `INamespace`).
- Fail-open everywhere: unknown `/v1/` path, malformed bytes, or any decode error forwards the original bytes unmodified — enrichment failure never drops telemetry, matching the JSON path.
- **Known limitation, disclosed in code and `docs/ADVANCED.md`:** re-encoding through a fixed descriptor drops fields a newer OTLP schema might add (the JSON path doesn't have this property). The docs carry the descriptor vintage and regeneration pointer.
- Out of scope, deliberately: gRPC transport (the receiver is plain HTTP; `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` covers Claude Code) and `claude_code.*` metric recognition / session-id mapping, which belong to companion mode (#491). No CHANGELOG entry — version pick (1.17.1 vs 1.18.0) left as a release call.
- Security posture, since this adds a parser to an inbound network path: the decode sits behind the receiver's existing guards (auth, rate limiting, body-size cap, SSRF-checked forwarding), every parse failure falls open to byte passthrough rather than an error path, and protobufjs decode allocates proportionally to the received (already size-capped) body. I know the receiver's v1 comment deliberately deferred this over the dependency question — if the project prefers no `protobufjs` at runtime, the alternative is a hand-rolled wire parser, which I'd argue is the worse review surface; happy to discuss on the issue instead if this should have gone there first.
- The descriptor's regeneration currently includes one manual step (deleting the three unused gRPC `*Service` stubs, documented in the header). A `scripts/regenerate-otlp-descriptor.ts` that automates fetch→pbjs→strip→write is a natural follow-up if you want it mechanically refreshable.

## Test plan

- [x] RED first: per-signal HTTP round-trips against the unchanged receiver showed forwarded protobuf carrying no injected attributes; GREEN after (60/60 in the receiver suite)
- [x] Fidelity proofs: 16-byte traceId with high-bit bytes, `startTimeUnixNano` > 2^53, `asInt` 2^53+1 (off-by-one under any float trip), and a no-op enrichment test asserting decode→re-encode equals the input byte for byte
- [x] Independent fixture: a log request serialized by the installed `@opentelemetry/otlp-transformer` `ProtobufLogsSerializer` (hand-rolled writer, shares no code with protobufjs) decodes and enriches correctly through our descriptor
- [x] Dist-level round-trip against compiled `dist/proxy/*` output
- [x] `npm run build && npm test` — 5869 tests pass; lint 0/0; prettier clean; existing JSON-path and passthrough tests unmodified and green


chrisdhaan:
## Review

**Verdict: solid, tested change — good to merge once you're happy with it.**

### Build / test / lint
- `npm run build` — passes
- `npm test` — 5869/5869 pass (matches the PR description exactly), 3 skipped (pre-existing, unrelated)
- `protobufjs` confirmed absent from the web bundle (`grep -r protobufjs src/web/` — no hits) and `check:bundle-size` passes at 232.1 KB gzipped

### Code review
- Decode/re-encode sits behind the existing body-size cap, auth check, and content-type allowlist, so it doesn't add a new unbounded-allocation surface — the 10 MB cap applies before any protobuf parsing happens.
- Fail-open behavior on decode error matches the JSON path exactly (forwards original bytes rather than dropping telemetry).
- `injectResourceAttributes` is reused unchanged between the JSON and protobuf paths, since `decodeOtlpRequest` normalizes to the same plain-object shape — no logic duplication.
- Fidelity test coverage is genuinely thorough: high-bit trace/span IDs, `>2^53` nanosecond timestamps, an independent-encoder fixture via `@opentelemetry/otlp-transformer` (not just protobufjs round-tripping against itself), and a byte-for-byte no-op re-encode test.
- No completeness gaps against #492 — covers all three signals (traces/metrics/logs), a superset of what the issue asked for. Deferring `claude_code.*` recognition to the companion-mode issue (#491, confirmed open and real) and gRPC transport is a reasonable scope boundary, not scope-dodging.
- No `src/shared/` edits.

### One suggestion — addressed
The PR flagged the descriptor's regeneration as a manual step and suggested a `scripts/regenerate-otlp-descriptor.ts` as a natural follow-up. I've pushed that automation on top of your branch (fetch → pbjs → strip → write), verified to produce byte-for-byte identical output to your hand-generated descriptor at the same pinned commit (only the header comment differs, since it now points at the script). Also merged current `main` into the branch to catch it up (a normal merge commit, no rebase, so your original commit is untouched) and bumped the version/CHANGELOG:

- **v1.23.0**
- CHANGELOG: *The inbound OTLP receiver now enriches protobuf-encoded payloads, not just JSON — `application/x-protobuf` bodies (the default for most OTel SDKs, including Claude Code's own) are decoded, tagged with session/repo context, and re-encoded, matching what the JSON path already did. A payload using OTLP schema fields newer than the receiver's vendored descriptor loses those unrecognized fields on re-encode; a payload that fails to decode is forwarded unmodified rather than dropped.*

Full build/test/lint re-verified green after both pushes.

chrisdhaan (approved):

PR #516, opened 2026-08-31
Homebrew tap distribution support

chrisdhaan:
## Summary
- Adds an installable Homebrew formula for `@newrelic/preflight` (macOS): `brew tap newrelic-experimental/preflight && brew install preflight`
- `scripts/update-homebrew.sh <version>` regenerates the formula (url + sha256) from the published npm tarball for each release
- `docs/maintaining-homebrew-tap.md` covers one-time tap repo setup and the per-release update process
- README documents the new install option alongside npm
- Stacked on the still-open `chore/v1.18.3-sync` (#515)

## Test plan
- [x] `npm run build`
- [x] `npm test` (177/178 suites, 6012 tests passing)
- [x] `npm run lint`
- [x] `scripts/verify-public-diff.sh` passed with no leak-pattern matches

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Preflight is now installable via Homebrew on macOS** (`brew tap newrelic-experimental/preflight && brew install preflight`), alongside the existing npm install path. The formula tracks the npm package; a `scripts/update-homebrew.sh` helper regenerates it for each new release, documented in `docs/maintaining-homebrew-tap.md`.
+
 ## [1.50.1] - 2026-09-10
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -116,6 +116,13 @@ Restart your AI tool — hooks and the MCP server load at session start. Every t
 
 **Cursor** — [Add to Cursor](https://cursor.com/en/install-mcp?name=newrelic-preflight&config=eyJjb21tYW5kIjoibnB4IiwiYXJncyI6WyIteSIsIkBuZXdyZWxpYy9wcmVmbGlnaHQiLCItLXN0ZGlvIl19) installs the MCP server via a one-click deeplink; run `preflight setup` afterwards for hook capture.
 
+**Homebrew (macOS)**:
+
+```bash
+brew tap newrelic-experimental/preflight
+brew install preflight
+```
+
 ---
 
 ## Works With
--- a/docs/maintaining-homebrew-tap.md
+++ b/docs/maintaining-homebrew-tap.md
@@ -1,0 +1,123 @@
+# Maintaining the Homebrew Tap
+
+The Homebrew tap for `@newrelic/preflight` lives at `newrelic-experimental/homebrew-preflight` on GitHub.com. This doc covers one-time tap setup and the per-release update process.
+
+## One-time tap setup
+
+These steps create the tap repo. Run them once when setting up the tap for the first time.
+
+**Prerequisites:** `gh` CLI authenticated to `github.com` (not GHE). If `gh auth status` shows only GHE, run `GH_HOST=github.com gh auth login` first.
+
+````bash
+# 1. Create the tap repo on github.com
+GH_HOST=github.com gh repo create newrelic-experimental/homebrew-preflight \
+  --public \
+  --description "Homebrew tap for @newrelic/preflight — AI coding observability"
+
+# 2. Clone it
+git clone https://github.com/newrelic-experimental/homebrew-preflight.git
+cd homebrew-preflight
+
+# 3. Create the Formula directory
+mkdir -p Formula
+
+# 4. Copy the current formula from this repo (adjust path as needed)
+cp /path/to/preflight/homebrew/Formula/preflight.rb Formula/
+
+# 5. Create a minimal README
+cat > README.md << 'EOF'
+# homebrew-preflight
+
+Homebrew tap for [@newrelic/preflight](https://github.com/newrelic-experimental/preflight) — AI coding observability for Claude Code and other AI coding tools.
+
+## Install
+
+```bash
+brew tap newrelic-experimental/preflight
+brew install preflight
+````
+
+EOF
+
+# 6. Commit and push
+
+git add -A
+git commit -m "preflight 1.1.0"
+git push origin main
+
+````
+
+**Verify the tap works:**
+
+```bash
+brew tap newrelic-experimental/preflight && brew install preflight && preflight --version
+```
+
+Expected: Homebrew installs `preflight` and `preflight --version` prints `1.1.0`.
+
+---
+
+## Per-release update (every new version)
+
+Run these steps each time a new version is published to npm.
+
+**1. In the `preflight` repo, update the formula:**
+
+```bash
+# Replace X.Y.Z with the new version
+scripts/update-homebrew.sh X.Y.Z
+```
+
+Review the change:
+
+```bash
+cat homebrew/Formula/preflight.rb
+```
+
+Confirm `url` and `sha256` match the new version.
+
+**2. Commit in the `preflight` repo:**
+
+```bash
+git add homebrew/Formula/preflight.rb
+git commit -m "Chore: update Homebrew formula for vX.Y.Z"
+```
+
+**3. Copy the updated formula to the tap repo:**
+
+```bash
+cp homebrew/Formula/preflight.rb /path/to/homebrew-preflight/Formula/preflight.rb
+```
+
+**4. Commit and push in the tap repo:**
+
+```bash
+cd /path/to/homebrew-preflight
+git add Formula/preflight.rb
+git commit -m "preflight X.Y.Z"
+git push origin main
+```
+
+**5. Verify:**
+
+```bash
+brew update
+brew upgrade preflight
+preflight --version
+```
+
+Expected: `preflight --version` prints `X.Y.Z`.
+
+---
+
+## Troubleshooting
+
+**`brew install` fails with sha256 mismatch**
+The sha256 in the formula doesn't match the downloaded tarball. Re-run `scripts/update-homebrew.sh <version>` — it fetches and hashes the tarball fresh.
+
+**`brew tap newrelic-experimental/preflight` returns 404**
+The tap repo doesn't exist yet or is private. Check `GH_HOST=github.com gh repo view newrelic-experimental/homebrew-preflight`.
+
+**`preflight --version` output doesn't match formula version**
+The formula version in the `url` field and the version field in `package.json` must match. They should always match when `update-homebrew.sh` is used.
+````
--- a/homebrew/Formula/preflight.rb
+++ b/homebrew/Formula/preflight.rb
@@ -1,0 +1,18 @@
+class Preflight < Formula
+  desc "AI coding observability for Claude Code and other AI coding tools"
+  homepage "https://github.com/newrelic-experimental/preflight"
+  url "https://registry.npmjs.org/@newrelic/preflight/-/preflight-1.50.1.tgz"
+  sha256 "fec1aa9d8e23f406d9f01121d4db27e4ac580147b58e47e071005faeeb1ccf4e"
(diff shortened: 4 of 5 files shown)


adjohn:
## Automated review at 6b437f1
**Verdict:** needs changes
**CI:** verify green at 6b437f1
**Scope:** no linked issue. Adds a Homebrew install path, a formula regenerator, a release-workflow step that pushes the formula to `newrelic-experimental/homebrew-preflight`, and docs. The tap repo exists with default branch `main`, so the `--base main` in the new step is right. The checked-in `sha256` at `homebrew/Formula/preflight.rb:5` matches the published `@newrelic/preflight@1.57.0` tarball; verified by hashing it.

### Findings

1. **[bug] The documented token scope cannot open the PR the step opens.** `docs/maintaining-homebrew-tap.md:60` describes `HOMEBREW_TAP_TOKEN` as "a fine-grained PAT scoped to only that repo, `Contents: Read and write`". `.github/workflows/release.yml:165` runs `gh pr create` and `:169` runs `gh pr merge --auto`, both of which need `Pull requests: Read and write` on the tap repo. `gh pr merge --auto` additionally needs auto-merge enabled in the tap repo's settings. Scenario: a maintainer provisions the PAT exactly as documented, runs the Release workflow, npm publish succeeds, then `gh pr create` returns 403. The step has no `continue-on-error` and no `if:` guard, so the Release job goes red after a successful publish, the MCP registry submission step below it never runs, and the tap is never updated. Fix: document `Contents: Read and write` plus `Pull requests: Read and write`, and either confirm auto-merge is enabled on the tap or drop the `--auto` line.

2. **[bug] Re-running the Release workflow for the same version fails at the push.** `.github/workflows/release.yml:159` sets `BRANCH="bump-${VERSION}"` and `:163` pushes it with no force and no existence check. Scenario: the workflow runs for 1.58.0 and opens tap PR `bump-1.58.0`. A maintainer has not approved it yet, so it has not merged. The workflow is re-run, which is a realistic path given the npm-propagation flakiness the comment at `:113` to `:120` documents for the step right above. The guard at `:153` compares the copied formula against the tap's `main`, which still lacks the bump, so the step proceeds. `git checkout -b bump-1.58.0` succeeds in the fresh shallow clone, but the new commit's parent is the tap's `main` while the remote branch's head is the previous run's commit off the same parent, so `git push origin bump-1.58.0` is rejected as a non-fast-forward. Step fails, release red, again after publish. Fix: `git push --force origin "$BRANCH"` (the branch is bot-owned and single-purpose) and let `gh pr create` fall through when a PR for that head already exists.

3. **[bug] The up-to-date guard silently no-ops when the formula is untracked in the tap.** `.github/workflows/release.yml:153` runs `git diff --quiet` before any `git add`, so it only sees changes to files git already tracks. Scenario: the tap repo has a `Formula/` directory but no `Formula/preflight.rb` yet, which is the state after one-time setup steps 1 to 3 in `docs/maintaining-homebrew-tap.md` if step 4 is skipped. `cp` at `:150` creates an untracked file, `git diff --quiet` reports clean, the step prints "Tap formula already up to date" and exits 0. The release is green and the tap was never touched, with nothing in the log to say so. Fix: `git add Formula/preflight.rb` first and test with `git diff --cached --quiet`, or test with `git status --porcelain`.

4. **[convention] The doc's code fences are unbalanced, so most of it renders as one code block.** `docs/maintaining-homebrew-tap.md:9` opens the setup block with four backticks. The nested heredoc block opens at `:33` with three backticks and is closed at `:36` with four, which closes the outer fence instead. Lines 38 to 45 then render as prose, with `# 6. Commit and push` becoming an H1 heading and the `git add`/`git commit`/`git push` lines as loose paragraphs. The four backticks at `:46` open a fresh fence that nothing closes until `:101`, so "Verify the tap works", the entire "Per-release update" section, and all of "Troubleshooting" render inside a code block. The heredoc is also wrong as shell: a copy-paste run writes a README whose fence is four backticks. Fix: close the nested block at `:36` with three backticks and drop the stray four-backtick line at `:46`.

5. **[convention] New user-facing functionality with no version bump, and an `[Unreleased]` heading this changelog does not use.** `CHANGELOG.md:8` adds `## [Unreleased]`; no other entry in the file uses that heading, and `main` has none. `package.json`, `package-lock.json`, `server.json`, `plugin/.claude-plugin/plugin.json`, and `kiro-power/plugin.json` are untouched. This matters more than usual here, because `.github/workflows/release.yml:146` reads the version straight out of `package.json` to build the formula URL, so the release that ships this feature is the one that decides which tarball the formula points at. Fix: bump to a real minor across all six files and retitle the section to match. Note that #722 already claims 1.58.0, so whichever of the two lands second needs 1.59.0.

### Notes (non-blocking)
- `.github/workflows/release.yml:149` embeds the PAT in the clone URL, so it lands in `$TAP_DIR/.git/config` and in any git error text. The runner is ephemeral and Actions masks secret values in logs, so the exposure is small, but `git -c http.extraheader=...` or a `GIT_ASKPASS` helper keeps it out of the on-disk config.
- `docs/maintaining-homebrew-tap.md:43` and `:54` still use `1.1.0` as the worked example while the repo is at 1.57.0. A reader following the one-time setup verbatim commits a message for a version that never existed.
- `scripts/update-homebrew.sh` has no test and no dry-run, and the formula it rewrites is regenerated in an ephemeral checkout at release time, so a regression in the `perl -i -pe` substitution surfaces only as a broken release. A `--check` mode that re-reads the formula and asserts the url and sha256 it just wrote would make the failure local.
- Principles applied: Make Operations Idempotent drove findings 2 and 3. A release step runs amid retries and partial state, so re-running it for the same version has to converge on the same tap PR rather than fail or silently skip. Prove It Works drove the scope line, where the committed `sha256` was checked by hashing the real published tarball rather than taken on faith, and the tap repo's existence and default branch were checked rather than assumed. Boundary Discipline drove finding 1. The PAT's permission set is the boundary contract between this workflow and the tap repo, and the doc that records it has to match what the code calls.

_Automated review; a maintainer makes the final call. Push a new commit for a fresh review._

adjohn:
## Automated review at 0b91c79

**Verdict:** needs changes
**CI:** verify green at 0b91c79
**Scope:** no linked issue. Adds a Homebrew install path, a formula regenerator, a release-workflow step that opens a PR against `newrelic-experimental/homebrew-preflight`, and a maintainer doc. The version bump is complete. `package.json:3`, `package-lock.json:3` and `:8`, `server.json:9` and `:14`, `plugin/.claude-plugin/plugin.json:4`, and `kiro-power/plugin.json:5` all move to 1.59.0 together, and the `Verify manifest versions are in sync` gate at `.github/workflows/release.yml:50` will pass. The formula's `test do` block works. `bin.install_symlink` plus npm's own `libexec/bin` symlink resolves through `realpathSync` in `src/version.ts:9`, so `dirname(resolved)/../package.json` lands on the installed package and `preflight --version` prints the real version. `HOMEBREW_TAP_TOKEN` is provisioned on this repo.

Since the previous review at 6b437f1, all five findings are resolved. The PAT scope doc now names both `Contents` and `Pull requests` (`docs/maintaining-homebrew-tap.md:59`). The push is `--force` with a deterministic branch and an existing-PR guard (`release.yml:176` to `:184`). The up-to-date check now stages first and tests `git diff --cached --quiet` (`release.yml:163`). The doc's code fences are balanced. The version bump landed at 1.59.0, clearing the 1.58.0 claim. Two earlier non-blocking notes are also resolved. The PAT no longer rides in the clone URL, and the doc's worked examples use `X.Y.Z`. The remaining findings below are new or carried over, and all four are small.

### Findings

1. **[bug] The tap clone has no `Formula/` directory guard, so `cp` can red the release after publish.** `.github/workflows/release.yml:154`. GitHub Actions runs `run:` blocks under `bash -e`, so `cp homebrew/Formula/preflight.rb "$TAP_DIR/Formula/preflight.rb"` into a nonexistent directory aborts the step. Scenario: the tap repo exists but `Formula/` does not, which is the partially-complete setup state the step's own comment at `:161` reasons about when it says the formula may not exist in the tap yet. The tag, the GitHub release, and the npm publish have all already happened by then. The step fails, and `Sync server.json version` at `:196` and `Publish to MCP Registry` at `:207` never run, so the release reaches npm but not the MCP registry. Fix: `mkdir -p "$TAP_DIR/Formula"` before the `cp`. I could not read the tap repo from this environment, so I cannot say whether the directory is present today. Independently, consider moving this whole step below `Publish to MCP Registry`. Nothing in the registry submission depends on the tap, and the tap is the less critical of the two.

2. **[bug] The documented manual-update path pushes to a branch the same doc says is gated.** `docs/maintaining-homebrew-tap.md:85` ends the "To update manually" recipe with `git push origin main`. `docs/maintaining-homebrew-tap.md:59`, twenty-six lines above, states that the tap's org-wide ruleset requires one approving review before a PR can merge into `main`. Scenario: the tap drifts, a maintainer follows the manual recipe verbatim, and the push is rejected by the ruleset with no fallback in the doc. `:44` in the one-time setup has the same `git push origin main`. Fix: make the manual recipe a branch plus a PR, matching what the workflow does, or say explicitly which of the two pushes the ruleset permits and why.

3. **[bug] A substitution that stops matching is indistinguishable from "nothing to do".** `scripts/update-homebrew.sh:32` to `:35` runs `perl -i -pe` with two `s|||` patterns and never checks that either one matched, then prints `Updated:` at `:38` regardless. `.github/workflows/release.yml:164` then reads a byte-identical formula as already current and exits 0. Scenario: someone reformats the formula, say to `url("...")` or to move the tarball into a `resource` block. The next release runs the script, which silently rewrites nothing, and the workflow prints `Tap formula already up to date for 1.60.0, skipping`. The release is green, the tap still serves 1.59.0, and no log line anywhere says the bump was lost. The new staging guard made the workflow side correct, which is exactly what leaves the script side as the only place this can go quiet. Fix: after the `perl` call, assert the result. `grep -q "preflight-${VERSION}.tgz" "$FORMULA_PATH"` and `grep -q "$SHA256" "$FORMULA_PATH"`, exit 1 otherwise.

4. **[convention] The checked-in formula pins 1.57.0 while the repo is at 1.59.0, and the doc claims they always agree.** `homebrew/Formula/preflight.rb:4` points at `preflight-1.57.0.tgz` with the matching 1.57.0 sha256, while `package.json:3` is 1.59.0. `docs/maintaining-homebrew-tap.md:99` asserts that the url version and the `package.json` version "must match" and "should always match when `update-homebrew.sh` is used". Scenario: a maintainer runs one-time setup, step 4 at `:23` copies this file into the tap verbatim, and the tap ships a 1.59.0-era repo's formula that installs 1.57.0. The `test do` block at `homebrew/Formula/preflight.rb:16` still passes, because `version` is inferred from the url. The design is sound, since the workflow regenerates the url at release time, but the checked-in file is a placeholder and nothing says so. Fix: say in the formula and at `:99` that the committed url and sha256 are a placeholder regenerated at release, and that only the tap's copy is authoritative.

### Notes (non-blocking)

- `.github/workflows/release.yml:140` sets `HOMEBREW_TAP_TOKEN` in the step env and nothing reads it. Every authenticated call goes through `gh`, which reads `GH_TOKEN` at `:141`, and `scripts/update-homebrew.sh` touches neither. Delete the line.
- `.github/workflows/release.yml:185` ends `gh pr merge --auto --squash` with `|| true`, which swallows the one outcome a maintainer needs to know about. If auto-merge is disabled on the tap or the PAT is short a scope, the release is green and the tap PR sits unmerged forever. `|| echo "auto-merge not armed, approve and merge the tap PR by hand"` keeps the step non-fatal and still says so.
- `scripts/update-homebrew.sh:6` uses 1.2.0 as the worked example and `:13` uses 1.1.0, seven lines apart in the same file.
- Principles applied. Make Operations Idempotent drove findings 1 and 3. A release step runs amid retries and partial external state, so every path through it has to either converge or fail loudly, and finding 3 is the one path left that does neither. Prove It Works drove the scope paragraph, where the formula's `test do` block was traced through `src/version.ts`'s `realpathSync` chain to confirm `preflight --version` resolves under a double symlink rather than assumed to work, and the six version fields were read individually against the gate at `release.yml:50`. Boundary Discipline drove findings 2 and 4. The doc is the contract between this repo and the tap repo, and a contract that contradicts itself twenty-six lines apart, or asserts an invariant the committed tree violates, misleads the one reader who most needs it. Laziness Protocol drove the first note, since a dead env entry on a secret-bearing step is the kind of line a later reader reasons about for nothing.

_Automated review; a maintainer makes the final call. Push a new commit for a fresh review._

PR #529, opened 2026-08-31
Fix #521: Read native duration_ms instead of pre/post wall-clock delta

### LINKED ISSUE #521: Tool latency includes permission-prompt wait time — read Claude Code's native duration_ms instead of the pre/post hook wall-clock delta
## Problem

Every tool-call duration Preflight reports — `LatencyTracker`, `nr_observe_get_latency_percentiles`, per-tool cost/latency breakdowns — is computed in `src/hooks/event-processor.ts` as:

```
durationMs: Math.max(0, event.timestamp - preEvent.timestamp),
```

That's wall-clock time between when this collector received `PreToolUse` and when it received `PostToolUse`. It necessarily includes:
- however long the user took to approve a permission prompt for that tool call,
- execution time of any *other* `PreToolUse` hooks configured for the same tool (ours included).

Claude Code's `PostToolUse`/`PostToolUseFailure` hook payloads now carry a `duration_ms` field defined explicitly as "tool execution time, **excluding** permission prompts and PreToolUse hooks" — the exact number our own pre/post delta contaminates. Any session with interactive permission prompts (i.e. most non-`--dangerously-skip-permissions` sessions) has inflated p95/p99 tool latency today, and the inflation is invisible — it looks like real tool slowness.

## Proposed fix

- Read `data.duration_ms` off the `PostToolUse`/`PostToolUseFailure` payload in `src/hooks/collector-script.ts` and pass it through as the authoritative duration.
- Fall back to the current pre/post delta only when `duration_ms` is absent (older Claude Code versions, or other platforms that don't send it).
- Worth flagging on the dashboard/tool-selection-scorer side too: `ToolSelectionScorer` and `nr_observe_get_cost_per_tool` both key off the same `durationMs` field and inherit the same contamination.

adjohn:
## Summary
- Reads Claude Code's own `duration_ms` field off `PostToolUse`/`PostToolUseFailure` hook payloads (tool execution time, excluding permission-prompt wait and PreToolUse hook overhead) and uses it as `ToolCallRecord.durationMs` when present, instead of this collector's pre/post wall-clock delta. Falls back to the wall-clock delta on older Claude Code versions / other platforms that don't send the field.
- Adds `permissionWaitMs` to `ToolCallRecord` (wall-clock delta minus native duration, clamped to 0; `null` when there's no native duration to decompose against) so the human/hook-overhead time isn't just discarded — it's a real, separate signal, intended as a local-dashboard friction metric rather than something folded into cross-session latency percentiles or cost-per-tool numbers.
- Applies the same native-duration preference to the orphaned-post case (no matching pre-event), which previously always recorded `durationMs: null` even when the post event itself carried real duration data.

## Test plan
- [x] `npx jest -- src/hooks/collector-script.test.ts src/hooks/event-processor.test.ts` — 221 passed, including new cases for: capturing/validating `duration_ms`, preferring native duration over wall-clock delta, computing `permissionWaitMs`, falling back when the field is absent, clamping on clock skew, and the orphaned-post native-duration case
- [x] `npm test` (full suite, also re-run by the pre-push hook) — 5983 passed, 3 pre-existing skips

Fixes #521


chrisdhaan:
Thanks for this — confirmed the bug and the fix. Ran a full local verification and code review:

**Build/test/lint:** clean. `npm run build`, `npm test` (5983 passed, 3 pre-existing skips — matches your test plan exactly), `npm run lint` — all pass.

**Logic:** checked the `hasNativeDuration` gating, the `permissionWaitMs` clamp on clock skew, and the orphaned-post case — all correct, and the new tests exercise every branch (matched pair, fallback, clamp, orphan-with-native-duration).

Two things worth a look before merge:

1. **Suggestion — cross-platform scope of `duration_ms`.** The `posttooluse`/`posttoolusefailure` branches in `collector-script.ts` are shared by every platform using the uniform hook envelope (Kiro, Amazon Q, Droid, Codex, VS Code Copilot), but the new `duration_ms` handling cites only Claude Code's own hooks docs for the "excludes permission-prompt wait + PreToolUse hook time" semantics. I checked Factory Droid's hooks reference directly — its `PostToolUse` payload has no `duration_ms` field at all, so the fallback path handles it correctly there. I couldn't verify Kiro/Amazon Q/Codex the same way. Not a proven issue today, but if any of those platforms send `duration_ms` with different semantics down the line, this would silently mislabel their overhead as tool time. Given `collector-script.ts` intentionally has no platform-adapter import, the practical fix is probably just a comment flagging the assumption rather than a code branch.

2. **Issue — branch is behind main.** This forked before v1.18.3; main is now several releases ahead and one of those PRs (#502, permission rejection/denial classification) touches the same `else if` chain in `collector-script.ts`, producing one small textual merge conflict (adjacent branches, resolves by keeping both). Needs a sync before this can land cleanly.

I'll push a merge commit to bring the branch up to date plus a couple of small follow-up commits (version bump/changelog, and a comment addressing #1) so this is ready to merge. Nothing else needed from you — flagging here for visibility.

chrisdhaan (approved):

PR #530, opened 2026-08-31
Fix #522: Capture native agent_id/agent_type on every hook event

### LINKED ISSUE #522: Hook events now carry agent_id/agent_type natively — stop inferring subagent identity from transcript filenames
## Problem

Preflight's only source of subagent identity today is filename parsing: `src/hooks/subagent-watcher.ts` derives `agentId` by slicing it out of the transcript filename it's watching —

```
const agentId = name.slice('agent-'.length, -'.jsonl'.length);
```

— and that `agentId` is what eventually reaches `CostTracker.recordTokenUsage(usage, model, { agentId })` for the subagent/parent cost split, and `WorkflowRunTracker`'s per-run attribution. It works, but it's a filesystem-convention dependency: it breaks silently if Claude Code ever changes the transcript file naming scheme, and it can't distinguish `agent_type` at all (subagent workflow analytics currently has no per-type breakdown — just an opaque id).

Claude Code hook events (`PreToolUse`/`PostToolUse`/etc.) now include `agent_id` (for subagents) and `agent_type` (for subagents and sessions started with `--agent`) directly in the payload. That's a first-party field on every hook event, not something to reconstruct from a filename.

## Proposed fix

- Read `data.agent_id`/`data.agent_type` in `src/hooks/collector-script.ts` and thread them onto `ToolCallRecord` the same way `subagent_type` already is (see the existing `obj.subagent_type` handling nearby).
- Prefer the hook-native `agent_id` over the filename-derived one where both exist; keep the filename fallback for older Claude Code versions or platforms without the field.
- Use `agent_type` to add a per-type dimension to subagent cost/workflow metrics (e.g. "how much did `general-purpose` subagents cost vs `Explore` subagents this session") — that breakdown doesn't exist today because the type was never available.

adjohn:
## Summary
- Claude Code now sends `agent_id`/`agent_type` on every hook event, identifying which subagent (if any) made a given tool call. Threads these through `collector-script.ts` (common-fields attachment, same site as `sessionId`/`cwd`/`transcriptPath`) and `event-processor.ts`'s pairing logic onto `ToolCallRecord.agentId`/`agentType`.
- One correction versus how I originally scoped issue #522: individual tool calls had **no** subagent attribution at all before this fix — this isn't replacing a fragile filename-based mechanism, it's adding a capability that didn't exist. `SubagentWatcher`'s filename-derived `agentId` (parsed from `agent-<id>.jsonl`) is a separate, still-necessary pipeline: it extracts subagent **token usage** from transcript files, a different signal from per-tool-call attribution, and this PR doesn't touch it.
- Paired records prefer the pre-event's `agentId`/`agentType`, falling back to the post-event's when the pre-event has none. Orphaned-post records (no matching pre-event) still get `agentId`/`agentType` from the post event itself.

## Test plan
- [x] `npx jest -- src/hooks/collector-script.test.ts src/hooks/event-processor.test.ts` — 220 passed, including new cases for: capturing `agent_id`/`agent_type` on both Pre/PostToolUse, omission for parent-session calls, pre-event preference with post-event fallback, and the orphaned-post case
- [x] `npm test` (full suite, also re-run by the pre-push hook) — 5982 passed, 3 pre-existing skips

Fixes #522


chrisdhaan:
Thanks for this — verified locally:

**Build/test/lint:** ✅ Clean. `npx tsc -b .`, `npm test` (5982 passed, 3 pre-existing skips), `npm run lint` (0 errors/warnings) — matches the test plan above.

**Findings:**

1. **Issue:** `agentId`/`agentType` reach the `AiToolCall` NR event in camelCase, not the `agent_id`/`agent_type` snake_case used by every other attribute on that event (`tool_use_id`, `duration_ms`, `error_type`, `session_id`, etc.). `toolCallToNrEvent()` in `src/transport/nr-ingest.ts` explicitly snake_cases known fields via `STANDARD_KEYS`, then passes unknown fields through under their original key — `agentId`/`agentType` aren't in that set, so they fall through unchanged. The same file already has an explicit `agent_id: event.agentId` mapping for the separate `SubagentTokenEvent`/`WorkflowRunEvent` path, so this breaks an established local convention. No test in `toolCallToNrEvent()`'s suite covers it. Pushed a follow-up commit adding the explicit mapping plus a test.

2. **Suggestion:** No test covered the pre-event/post-event carrying *different* `agentId` values, verifying the pre-event wins. Added that test.

3. **Nit:** Branch was several releases behind `main`. Merged `main` in (merge commit, not a rebase) to bring it current — all conflicts were trivial sibling optional-field additions next to #529's `platform`/`nativeDurationMs` work, nothing logical.

4. Pushed a version bump + CHANGELOG entry, since this reads as new capability (per your own PR description — no prior per-tool-call subagent attribution existed) rather than a fix to broken behavior.

Nice catch keeping this scoped to the first bullet of #522 rather than trying to land all three at once — worth deciding separately whether the remaining two (preferring native `agent_id` over the filename-derived one, and per-`agent_type` cost/workflow breakdowns) should stay open as follow-up work on #522.

chrisdhaan (approved):

PR #532, opened 2026-08-31
Fix #523: Wire Claude Code's PostModelSwitch hook into ModelUsageTracker

### LINKED ISSUE #523: PreModelSwitch/PostModelSwitch hooks would give ModelUsageTracker visibility into fallbacks and mid-session model changes it can't see today
## Problem

`ModelUsageTracker` (`src/metrics/model-usage-tracker.ts`) only ever learns "which model" from whatever model string arrives with a token-usage report (`recordUsage(model, ...)`, called from `src/index.ts` wherever tokens are self-reported or parsed from a transcript). It has no independent signal for *when or why* the model in use actually changed mid-session.

That matters more than it used to, because Claude Code now silently substitutes models under a few conditions the changelog documents directly:
- automatic retry on a fallback model after certain non-retryable API errors,
- fast mode toggling model behavior/pricing tier,
- a user running `/model` mid-session.

Right now all of these show up in Preflight, if at all, as just another `recordUsage()` call with a different model string mixed into the same session totals — there's no event marking "a switch happened here, for this reason." A session that silently fell back from Opus to Sonnet mid-turn because of a transient API error looks identical, in Preflight's data, to a user who deliberately switched models.

Claude Code added `PreModelSwitch`/`PostModelSwitch` hook events specifically to let hooks "block, confirm, or annotate a model switch" — which means the event payload should carry the from/to model and (for automatic switches) a reason.

## Proposed fix

- Install `PostModelSwitch` (informational only — no need to block anything) and dispatch it in `src/hooks/collector-script.ts`.
- Feed it into `ModelUsageTracker` as a discrete switch event, not just another usage record, so a dashboard can show "N model switches this session, M of them automatic fallbacks" instead of inferring it from a model column changing between rows.
- Consider surfacing automatic-fallback switches as a lightweight signal alongside `ApiFailureTracker`'s data — a fallback switch is usually a symptom of the same underlying API trouble `StopFailure` reports on.

adjohn:
## Summary

`ModelUsageTracker` previously had no independent signal for model switches — it only learns "which model" from whatever string arrives with a token-usage report (`recordUsage()`), with no notion of *when or why* the model changed mid-session.

This wires **only** `PostModelSwitch` — not `PreModelSwitch`, which exists to block/confirm a switch (decision control Preflight has no reason to use), and every field it needs is also present on `PostModelSwitch`'s own input:
- **Installer** (`install-helper.ts`): registers `PostModelSwitch` as its own settings.json hooks key, same pattern as `StopFailure`/`InstructionsLoaded` — `HookEntries`, `generateHookEntries()`, the merge/remove loops, `NR_HOOK_RE`, and the zod schema all extended. Deliberately **not** added to `areHooksInstalled()`, following the existing precedent for `StopFailure`.
- **Collector** (`collector-script.ts`): new `postmodelswitch` dispatch branch capturing `from_model`/`to_model`/`requested_model`/`source` into a new `model_switch` buffer mode. Model IDs are a closed-ish identifier vocabulary, not free-text content, so — unlike `StopFailure`'s `error_details` — these aren't gated behind `recordContent`. Deliberately **not** capturing the cache-cost-estimate fields (`context_tokens`, `prompt_cache_warm`, `cache_ttl`, `estimated_cache_write_usd`, `pricing`) — that's a separate, richer cache-cost-forecast feature, out of scope here.
- **Event processor** (`event-processor.ts`): new `ModelSwitchHookEvent`/`ModelSwitchFrame` types and an `onModelSwitch` callback, mirroring `onApiFailure`'s shape exactly.
- **Tracker** (`model-usage-tracker.ts`): new `recordModelSwitch()` with a bounded FIFO event log (100 entries, `ApiFailureTracker`'s event-list pattern), plus `switchCount`/`automaticSwitchCount` (switches with `source === 'auto'`) on `ModelUsageMetrics`. `combineBreakdowns()` returns empty switch fields — switch history is this process's own live state, not part of any persisted per-model breakdown, so there's nothing to combine across sessions.
- **Wiring** (`index.ts`): one new `onModelSwitch` callback calling `modelUsageTracker.recordModelSwitch()`.
- `docs/ADAPTERS.md` updated.

## Correction to the original issue

I'd originally framed this partly around detecting *"Claude Code now retries a turn once on the fallback model when the API rejects an unexpected non-retryable error"* (from the changelog) as a silent-fallback case this hook would expose. Checked against code.claude.com/docs/en/hooks.md directly and that's not quite right: Claude Code does **not** fire `PostModelSwitch` for a single-turn fallback-model-chain substitution that leaves the session's nominal model unchanged — only for a **persistent** automatic change (`source: 'auto'`, e.g. a sustained fallback or `opusplan` entering/leaving plan mode) or the model restored on resume (`source: 'resume'`). The specific one-turn transient case from the changelog stays invisible to any hook; this PR is honest about that in the doc comments rather than overclaiming.

## Test plan
- [x] `npx jest -- src/hooks/collector-script.test.ts src/hooks/event-processor.test.ts src/install/install-helper.test.ts src/metrics/model-usage-tracker.test.ts` — 338 passed, including new coverage for: hook generation/merge/idempotency/removal, collector field capture (including the "no gating on recordContent" case), event-processor routing (including the no-callback and throwing-callback cases), and the tracker's switch-log bounding/reset/automatic-count behavior
- [x] `npm test` (full suite, also re-run by the pre-push hook) — 5995 passed, 3 pre-existing skips

Fixes #523


chrisdhaan (commented):
Reviewed and verified locally — looks good, with two small doc gaps to close before merge.

**Build/test/lint:** all clean, matching the numbers in the PR description.
- `npm run build` — clean
- `npm test` — 5995 passed, 3 pre-existing skips

**Design:** matches the existing `StopFailure` hook precedent exactly — its own settings.json hooks key, full merge/remove/idempotency coverage in `install-helper.ts`, and clearly-stated scope cuts (no `PreModelSwitch`, no cache-cost-estimate fields). The correction note in the PR description about what `PostModelSwitch` does and doesn't cover (no visibility into the one-turn fallback-chain case) is appreciated — that kind of honesty about a hook's actual limits is exactly what this tracker needs to stay trustworthy.

Test coverage is thorough: FIFO bounding/eviction, defaults, `reset()`, `combineBreakdowns()` zeroing the switch fields (since switch history is live per-process state, not part of a persisted breakdown), hook idempotency, and both the no-callback and throwing-callback paths in the event processor.

**Two documentation gaps** (functionally harmless — the tool already returns the new data — but worth closing so the feature is actually discoverable):
- `docs/COMMANDS_TABLE.md`'s sample response for `nr_observe_get_model_usage` doesn't show the new `switchCount` / `automaticSwitchCount` / `recentSwitches` fields.
- The tool's own MCP `description` string in `analytics-tools.ts` doesn't mention model switches, so nothing signals that this tool now covers that.

I'll bring the branch up to date with `main` via a merge commit and add the version bump, CHANGELOG entry, and the two doc fixes above as follow-up commits on this branch so it's ready to merge.

chrisdhaan (approved):

PR #533, opened 2026-08-31
Fix #527: Wire SessionStart resume-cost fields into a new SessionResumeTracker

### LINKED ISSUE #527: SessionStart resume hooks now report session staleness and estimated re-cache cost — Preflight doesn't consume SessionStart at all
## Problem

Preflight never installs or dispatches a `SessionStart` hook for its own purposes today — confirmed on current `main` (`src/hooks/collector-script.ts` has no `sessionstart` branch). The closest thing that exists is `GitEfficiencyTracker` *recommending the user* add their own `SessionStart` hook for `git fetch` (`src/metrics/git-efficiency-tracker.ts`, the "Field guide: suggest SessionStart hook" section) — Preflight knows the hook exists, it just never listens to one itself.

Claude Code's `SessionStart` resume payload now includes the session's **staleness** and an **estimated re-cache cost** — a prediction, made before the resume happens, of how much it will cost to rebuild the prompt cache for a stale/resumed session. Two things fall out of not reading this:

- `CostForecastTracker` (`src/metrics/cost-forecast.ts`) has no concept of resume-driven cost at all — a session that resumes after being stale for a while will show a cache-hit-rate dip and a cost bump with no explanation attached, when Claude Code itself already knows in advance that this was coming and roughly how much.
- There's no independent check between what Claude Code *estimated* the re-cache would cost and what `CostTracker` actually observed once the tokens came in — a natural, cheap reconciliation signal (same shape as the `token_reconciliation_delta` self-check `AiWorkflowRun` already does elsewhere).

## Proposed fix

- Install and dispatch `SessionStart` in `src/hooks/collector-script.ts`, at minimum for the resume case.
- Feed the reported staleness/estimated-re-cache-cost into `CostForecastTracker` as a one-time forward-looking adjustment right after a resume, and optionally log the delta against what `CostTracker` actually measures once real usage comes in.

adjohn:
## Summary

Preflight never installed or read a `SessionStart` hook at all before this — confirmed against current `main` (no `sessionstart` branch in `collector-script.ts`, no `SessionStart` key anywhere in the installer). A session resumed after being stale showed up as an unexplained cost bump and cache-hit dip, with nothing attached even though Claude Code's own `SessionStart` hook already reports the staleness (`seconds_since_last_response`) and its own estimate of what re-warming the prompt cache will cost (`estimated_cache_write_usd`), for `source: 'resume'`/`'fork'` sessions with a prior response (v2.1.251+, per code.claude.com/docs/en/hooks.md).

- **Installer** (`install-helper.ts`): registers `SessionStart` as its own settings.json hooks key, same pattern as `StopFailure`/`InstructionsLoaded`/`PostModelSwitch` — appends to (never replaces) any existing `SessionStart` hook a user already has, notably the git-fetch hook `GitEfficiencyTracker` itself already recommends adding. Deliberately **not** added to `areHooksInstalled()`, following the existing precedent.
- **Collector** (`collector-script.ts`): new `sessionstart` dispatch branch capturing `source`/`seconds_since_last_response`/`context_tokens`/`prompt_cache_likely_expired`/`estimated_cache_write_usd` into a new `session_start` buffer mode. `SessionStart` fires on **every** session (startup/resume/clear/compact/fork); only `resume`/`fork` carries the four resume-cost fields, so every other case is a fast no-op. None of it is gated behind `recordContent` — `source` is a closed enum, the rest are numbers/booleans.
- **Event processor** (`event-processor.ts`): new `SessionStartHookEvent`/`SessionStartFrame` types and an `onSessionStart` callback, mirroring `onApiFailure`'s shape exactly.
- **New tracker** (`session-resume-tracker.ts`): `SessionResumeTracker` — `resumeCount`, `totalEstimatedCacheWriteUsd` (summed across every resume), and the last resume event. `recordResume()` is only ever called when all four resume-cost fields are present (checked in `index.ts`'s `onSessionStart` callback).
- **Surfacing**: `nr_observe_get_cost_forecast` gains an optional `resumeContext` field carrying the tracker's data alongside the existing forecast. `buildCostForecastFromInputs` (the forecast math itself) is completely untouched — `resumeContext` is appended as a sibling field in the handler, so it can't perturb the well-tested rate/confidence calculations.
- `docs/ADAPTERS.md` and `docs/COMMANDS_TABLE.md` updated.

## Scope cut

Matching this repo's own precedent for partial integration (see `ApiFailureTracker.recordRequest()`'s dormant-forever note in `index.ts`): `NrIngestManager` (NR cloud-egress harvest) and the local dashboard's own REST API (`DashboardServer`'s `api.costForecast` closure) do **not** get this data in this PR. Both are real, separate follow-on integration points into subsystems this PR doesn't otherwise touch — this PR covers the capture side completely and the one MCP tool the original issue named.

## Test plan
- [x] `npx jest -- src/hooks/collector-script.test.ts src/hooks/event-processor.test.ts src/install/install-helper.test.ts src/metrics/session-resume-tracker.test.ts src/tools/cost-tools.test.ts` — all passing, including new coverage for: hook generation/merge/idempotency/removal (including "foreign SessionStart hook preserved"), collector field capture (present vs. absent resume fields, no recordContent gating), event-processor routing, the tracker's accumulation/reset behavior, and `handleGetCostForecast`'s `resumeContext` field (absent/empty/populated, and that it never changes the forecast fields)
- [x] Fixed two pre-existing tests (`collector-script.test.ts`) that used `SessionStart` specifically as their example of "an unhandled hook event name" — updated to `PreCompact`, which is still genuinely unhandled
- [x] `npm test` (full suite, also re-run by the pre-push hook) — 5997 passed, 3 pre-existing skips

Fixes #527


chrisdhaan (commented):
## Review

**Scope check:** matches the exact pattern of the three most recently merged PRs (#529–532), which wired `StopFailure`/`InstructionsLoaded`/`PostModelSwitch` into dedicated trackers. Confirmed against #527, which this fixes.

**Build/test/lint — all clean locally:**
- `npx tsc -b .` — clean
- `npm test` — 5997 passed, 3 pre-existing skips (matches the PR description exactly)
- `npm run lint` — clean

**Code review:**
- Traced `sessionResumeTracker` end-to-end: `index.ts`'s `onSessionStart` callback → `SessionResumeTracker.recordResume()` (correctly gated on all four resume-cost fields being present) → `ToolRegistrationOptions` → `registerCostTools()` → `resumeContext` on `nr_observe_get_cost_forecast`.
- Verified `resumeContext` is purely additive — a test confirms `buildCostForecastFromInputs`'s output is byte-identical with and without a resume recorded.
- The collector's new `sessionstart` branch omits `tool`/`sessionId` inside the branch itself, but that matches the existing `stopfailure`/`api_failure` branch precedent exactly — both get those fields attached generically afterward. Not a new inconsistency.
- Correctly does not gate the four resume-cost fields behind `recordContent` — they're a closed enum plus numbers/booleans, no free text.
- No edits to `src/shared/`. No bugs found.

**One thing to resolve before merge:** this branch is currently 17 commits behind `main` (`main` is at 1.29.0; this branch still shows 1.18.2), and GitHub reports it as conflicting. No independent `SessionStart` handling has landed on `main` in the meantime, so this isn't a no-op — it just needs updating. I'll bring it up to date with a merge commit rather than a rebase.

chrisdhaan (approved):

Issue #524, opened 2026-08-31
ClaudeMdTracker/InstructionDriftTracker infer instruction-file load timing from Edit/Write calls — InstructionsLoaded hook reports the real moment


adjohn:
## Problem

`ClaudeMdTracker` and `InstructionDriftTracker` both currently learn about CLAUDE.md / `.claude/` instruction-file changes exclusively from `Write`/`Edit` tool calls (`src/metrics/claudemd-tracker.ts`: "Within a session: Write/Edit tool calls targeting CLAUDE.md or .claude/") plus cross-session SHA-256 hash diffing. That's a proxy for "the file changed," not "the file was (re)loaded into this session's context" — the two aren't the same moment, and for `InstructionDriftTracker`'s whole purpose (correlating instruction-file changes with subsequent behavior) the load moment is the one that actually matters.

Claude Code added an `InstructionsLoaded` hook event that fires exactly when CLAUDE.md or `.claude/rules/*.md` files are loaded into context — including on session start and on reload, independent of whether an edit happened in *this* session at all (e.g. a file someone else edited between sessions, or a `SessionStart` hook that returns `reloadSkills: true` mid-session).

## Proposed fix

- Install and dispatch `InstructionsLoaded` in `src/hooks/collector-script.ts`.
- Use its firing as the authoritative "instruction context changed here" timestamp for `InstructionDriftTracker`'s before/after windowing, instead of (or as a cross-check against) the Write/Edit-call heuristic.
- This also closes a real gap the current heuristic can't cover at all: a CLAUDE.md edit made *outside* the tracked session (another session, another tool, a teammate) still changes what gets loaded into context here — `InstructionsLoaded` sees that; Edit/Write-call tracking never can.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #535, opened 2026-08-31
Fix #520: wire UserPromptSubmit/Stop hooks as turn/task boundary signals

### LINKED ISSUE #520: Wire UserPromptSubmit/Stop hooks — TurnTracker/TaskDetector run on heuristics, LatencyDecompositionTracker stays fully dormant because of it
## Problem

Preflight's Claude Code integration installs and reads `PreToolUse`/`PostToolUse`/`PostToolUseFailure`/`StopFailure` hooks, but never `UserPromptSubmit` or `Stop` — both long-standing Claude Code hook events that mark the actual start and end of a turn. This isn't a guess; `src/index.ts` says so explicitly in its own comment where `LatencyDecompositionTracker` is left permanently unset:

> "there is no turn-boundary hook (UserPromptSubmit/Stop) wired either, so even a coarse tool-execution-vs-everything-else split isn't currently derivable... Kept dormant; would need new hook wiring..."

Three trackers pay for this gap today:
- `TurnTracker` (`src/metrics/turn-tracker.ts`) groups tool calls into a "turn" using a **2-second idle-gap heuristic** (`DEFAULT_GAP_THRESHOLD_MS`), not a real turn boundary — two genuinely separate turns that happen to follow each other quickly get merged, and one slow turn with a mid-turn pause can get split.
- `TaskDetector` (`src/metrics/task-detector.ts`) infers "task complete" from a **30-second idle-gap heuristic** (`DEFAULT_IDLE_TIMEOUT_MS`) plus `AskUserQuestion`/`TaskUpdate` signals — again, a proxy, not the real boundary.
- `LatencyDecompositionTracker` is never instantiated at all (`const latencyDecompositionTracker: LatencyDecompositionTracker | undefined = undefined`) — it's a complete, tested tracker with no producer, entirely because of this missing hook.

## Why it matters now

Claude Code's `Stop`/`SubagentStop` hook input has also picked up `background_tasks` and `session_crons` fields since these trackers were last touched — wiring the hook doesn't just fix turn boundaries, it's also the only way to see those two fields at all, which would help `WorkflowRunTracker`'s background-task visibility.

## Proposed fix

- Add `UserPromptSubmit` and `Stop` to the installed hook set (alongside the existing `PreToolUse`/`PostToolUse`/`StopFailure` entries) — see `src/install/install-helper.ts`'s hook-type list.
- Dispatch `usersubmitprompt`/`stop` event names in `src/hooks/collector-script.ts` next to the existing `stopfailure` branch.
- Feed real turn/task boundaries to `TurnTracker`/`TaskDetector` from the `Stop` event instead of (or as a tiebreaker alongside) the idle-gap heuristics.
- Instantiate `LatencyDecompositionTracker` for real, using `Stop` timestamp minus the sum of tool `duration_ms` (see the companion issue on reading `duration_ms` natively) as the coarse "everything that wasn't a tool call" bucket.

adjohn:
## Summary

Closes #520.

`TurnTracker`'s idle-gap heuristic (default 2s gap) can merge two genuinely separate turns that happen to follow each other quickly, and it understates a turn's true duration by missing the time Claude spends generating its final response after the last tool call. `TaskDetector`'s idle-based task boundary has the same kind of lag. Claude Code's native `UserPromptSubmit` and `Stop` hooks give a precise, corroborating signal for both boundaries — this is additive/fallback, not a replacement for the existing heuristics, since `Stop` structurally never fires on a user interrupt (a conversation can end without it).

- `storage/types.ts` — new `UserPromptSubmitHookEvent`/`StopHookEvent` (just `mode` + optional `sessionId`; no prompt/response content captured — neither hook's free text is needed for a boundary marker, and capturing it would add a new sensitive-content surface gated behind `recordContent`)
- `collector-script.ts` — dispatches the two new `hook_event_name` values
- `event-processor.ts` — new `BoundaryFrame` + `onUserPromptSubmit`/`onStop` callback options, routed through a shared `handleBoundaryEvent` helper (mirrors the existing `onApiFailure` pattern)
- `turn-tracker.ts` — `TurnTracker.finalizeTurnAt(stopTimestampMs)`: closes the currently-open turn immediately on `Stop`, only ever *extending* (never shrinking) the recorded end time, so a late/skewed `Stop` timestamp can't make a turn look shorter than its own last tool call
- `task-detector.ts` — `TaskDetector.startTaskIfNone(timestamp)`: opens a new task on `UserPromptSubmit` if none is already active. Task *closing* on `Stop` reuses the existing `markBoundary()` method (already wired to the `AskUserQuestion` boundary signal) — no new closing logic needed.
- `index.ts` — wires both callbacks into `turnTracker`/`taskDetector`
- `install-helper.ts` — registers both hooks: `generateHookEntries()`, `NR_HOOK_RE`, the settings zod schema, and both merge/remove hook-type loops
- `docs/ADAPTERS.md` — documents the new hooks

**Explicitly out of scope:** `LatencyDecompositionTracker` still requires genuinely-measured LLM-API latency (`TurnTimingReport.llmApiMs`), which neither of these hooks provides — left as its existing "not currently functional" scope cut (see `docs/COMMANDS_TABLE.md`), not fed misleading derived data.

## Test plan

- [x] `npm test` — full suite green (6000 passed, 3 pre-existing skips)
- [x] New/updated tests: `collector-script.test.ts`, `event-processor.test.ts`, `install-helper.test.ts`, `turn-tracker.test.ts`, `task-detector.test.ts`

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


chrisdhaan (commented):
## Triage summary

**Sense-check:** passes — squarely in scope, matches tracked issue #520 exactly, no duplicate PRs found.

**Build/test/lint on this branch:** all clean — `npm run build`, `npm test` (6000 passed, 3 pre-existing skips), `npm run lint` (0 errors/warnings).

### Issue — branch is stale relative to `main` and will not merge cleanly

This branch is 19 commits behind `origin/main`. Several PRs merged since it was forked (#521 native `duration_ms`, #522 `agent_id`/`agent_type`, #523 `PostModelSwitch`, #524 `InstructionsLoaded`, #527 `SessionStart`) touch the exact same surfaces this PR extends — the `HookEvent` discriminated union in `storage/types.ts`, the `processHook()` dispatch chain in `collector-script.ts`, `HookEventProcessor` in `event-processor.ts`, the hook-wiring block in `index.ts`, and the hook-type list in `install-helper.ts`. A test-merge against `origin/main` produces real content conflicts in all 5 of those files. I'll bring the branch up to date with a merge commit (not a rebase, to avoid rewriting your existing commits) as a follow-up.

### Suggestion — `startTaskIfNone` may create zero-tool-call "tasks"

`onUserPromptSubmit` unconditionally calls `taskDetector.startTaskIfNone(frame.timestamp)`. Previously, `TaskDetector` only ever started a task from a tool call (`recordToolCall`) — a task with zero tool calls never existed. Now a pure conversational turn with no tool calls (e.g. a quick Q&A) starts a task on `UserPromptSubmit` and gets closed via `markBoundary()` on `Stop`, producing a completed task with `toolCallCount: 0`. That will inflate `totalTasksCompleted` and dilute `averageToolCallsPerTask`/`averageTaskDurationMs`. Per the module's own definition ("the user gives an instruction, Claude executes a series of tool calls, and eventually responds"), a zero-tool-call turn probably shouldn't count as a completed task. I'll push a small fix + test for this alongside the rebase.

Otherwise this is clean — redaction discipline is correct and tested (`prompt`/`last_assistant_message` never captured even with `recordContent=true`), the boundary-extension logic in `TurnTracker.finalizeTurnAt`/`TaskDetector.startTaskIfNone`/`markBoundary` is all correctly no-op-safe, and `install-helper.ts` idempotency/foreign-hook-preservation is well tested. No `src/shared/` edits.

I'll push a merge commit to bring this up to date with `main`, plus follow-up commits for the task-count fix and the version/changelog bump. Thanks for the contribution!

chrisdhaan (approved):

PR #536, opened 2026-08-31
Fix #525: add costRateMultiplier and dataResidencyPremium config options

### LINKED ISSUE #525: Claude Code's modelPricing managed setting and data-residency premium mean Preflight's list-price cost can diverge from what an org is actually billed
## Problem

Related to #291, but a different root cause. #291 is about accounts on flat-rate/subscription plans where Preflight's per-token math shouldn't apply at all. This issue is about accounts that *are* billed per-token, but not at public list price:

- Claude Code added a `modelPricing` managed setting so an organization's **contracted per-model rates and discount multiplier** are used for `/cost`, the status line, and telemetry cost figures — instead of list price.
- Separately, cost estimates (`/cost`, status line, `--max-budget-usd`) now include a **1.1× US-only-inference premium for data-residency workspaces**.

Preflight computes cost entirely from its own vendored pricing table (`src/shared/pricing.ts` / `pricing-data.ts`, list price) plus the gap-fill overlay (`src/metrics/pricing-overlay.js`, post-#506). Neither mechanism has any way to apply an org's negotiated discount or the residency premium — so for any org using either setting, every dollar figure Preflight reports (session cost, budget-threshold alerts, cost-per-outcome, everything downstream of `CostTracker`) is systematically wrong in a *known, computable* direction, not just approximate.

## Proposed fix

Options, roughly in order of effort:
- Cheapest: let Preflight read the same `modelPricing` value out of the user's/org's Claude Code `settings.json` (it's already a documented managed setting) and feed it through the existing pricing-overlay mechanism as a per-org rate override — the overlay's "fill gaps, don't silently override the vendored table" guard (`applyGapFilledOverlay` in `src/metrics/pricing-overlay.js`) would need a deliberate opt-in path here, since this is an intentional override, not a gap-fill.
- Simpler stopgap: a config flag for the discount multiplier and/or a `dataResidency: true` flag that applies the 1.1× premium, so at least the *shape* of the correction is available even without auto-discovering the org's contracted rate.
- Document the gap clearly wherever cost figures are surfaced (dashboard, MCP tool descriptions) so users on negotiated contracts know the number is a list-price estimate, similar to whatever #291 lands on for subscription plans.

adjohn:
## Summary

Closes #525.

`CostTracker` computes every cost figure from its own vendored public list-price table (`src/shared/pricing.ts`), with no way to reflect an org's contracted discount or Claude Code's own 1.1x US-only-inference premium for data-residency workspaces. For any org on a negotiated rate, every dollar figure downstream — session cost, budget-threshold alerts, cost-per-outcome, the dashboard — was systematically wrong in a known, computable direction.

The issue proposed three options in increasing effort: auto-read Claude Code's `modelPricing` managed setting, a simpler config-flag stopgap, or just documenting the gap. I looked into the first option in detail (`docs/settings-reference#modelpricing`, `docs/managed-settings`) and it's genuinely not the "cheapest" option once you dig in: `modelPricing` is **Managed-only** scope, delivered through one of four mechanisms (server-managed settings from claude.ai, MDM/OS policy, a `managed-settings.json` file, or a policy helper program), and only the single highest-priority source actually present is the one in effect ("first-wins" by default). Only one of those four is a static file at a documented path. Reading that file unconditionally would risk silently applying a stale or overridden rate whenever a different mechanism is the one Claude Code actually selected — a worse outcome than list price, because it looks authoritative while still being wrong. So this PR implements the second and third options instead.

## Changes

- `config.ts` — new `costRateMultiplier` (0 < x ≤ 1, env `NEW_RELIC_AI_COST_RATE_MULTIPLIER`) and `dataResidencyPremium` (bool, env `NEW_RELIC_AI_DATA_RESIDENCY_PREMIUM`) fields, resolved with the same env > config-file > default precedence and validation pattern already used for the budget fields
- `index.ts` — combines both into a single factor passed to `CostTracker`'s constructor
- `cost-tracker.ts` — `CostTracker` now accepts an optional `rateMultiplier`, applied once at the single funnel point (`accumulateTokens()`) where `calculateCost()`'s result is consumed. Every downstream reader — `ModelUsageTracker`, `BudgetTracker`'s threshold alerts, day/task/workflow-run cost buckets — inherits the correction automatically, since they all read from the same scaled `CostBreakdown`. New `costRateMultiplierApplied` field on `CostMetrics` surfaces whether a correction is in effect (1 = none).
- `cost-tools.ts` — `nr_observe_get_cost_breakdown` reports the applied factor as `rate_multiplier_applied`
- `docs/ADVANCED.md` — new "Cost / Pricing Corrections" section documenting both new fields, plus the existing (previously undocumented) `customPricingFile` escape hatch for entering exact per-model contracted rates, and why auto-discovery of Claude Code's managed `modelPricing` is out of scope
- `docs/COMMANDS_TABLE.md` — documents the new `rate_multiplier_applied` field

**Explicitly out of scope**, left for a future issue if wanted:
- Auto-reading Claude Code's managed `modelPricing` setting (see rationale above)
- Per-response detection of data-residency billing — Claude Code's own `/usage` figure knows which specific API responses were billed at the residency rate; no hook or event exposes that to Preflight, so `dataResidencyPremium` is an explicit always-on opt-in flag, not automatic detection

## Test plan

- [x] `npm test` — full suite green (6027 passed, 3 pre-existing skips, 1 pre-existing todo)
- [x] New/updated tests: `cost-tracker.test.ts` (rateMultiplier scaling, including `savingsFromCacheUsd`), `config.test.ts` (env/file precedence, range validation), `cost-tools.test.ts` (`rate_multiplier_applied` in the tool response), plus the pre-existing `CostMetrics` literal fixtures in `session-store.test.ts` updated for the new required field

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


chrisdhaan:
## Triage review

**Sense check:** passes — closes a real, systematic cost-computation gap in `CostTracker` for orgs on negotiated rates or data-residency workspaces. No duplicate PRs found.

**Tracked issue:** #525, verified — body matches this PR's stated scope and tradeoff analysis exactly.

**Build/test/lint (on the branch as submitted):**
- `npm run build` — clean
- `npm test` — 6027 passed, 3 pre-existing skips, 1 pre-existing todo (matches the PR description's own numbers)

**Code review:**
- Correction is applied at the right funnel point (`accumulateTokens()` scales `calculateCost()`'s output) without touching the vendored `src/shared/pricing.ts`.
- Only one `CostTracker` construction site in production code (`src/index.ts`), so the fix is wired consistently — no other call site left uncorrected.
- Config validation (`validateFileRatio`, env > file > default precedence) matches the existing pattern used for `retainSessionsDays`/budget fields, including the same "invalid env var silently falls back to the file value" behavior already present elsewhere in `config.ts` — not a new inconsistency.
- Checked `SessionResumeTracker`'s `estimatedCacheWriteUsd` (added on `main` after this branch was forked) — it's Claude Code's own self-reported figure relayed verbatim, not computed via `CostTracker`, so it's correctly untouched by this PR.
- No issues or nits found.

**Branch staleness:** this branch was ~17 commits behind `main` (forked around v1.18.3; `main` is now at v1.31.0). I did a local trial merge to check impact: exactly one conflict, a trivial doc-insertion collision in `docs/ADVANCED.md` between this PR's new "Cost / Pricing Corrections" section and an unrelated "Companion Mode" section added independently — both insert content at the same point, resolved by keeping both. After that, build/test/lint all still pass clean on the merged result, so the fix itself needed no changes for current `main`.

**Version bump / CHANGELOG:** not yet present on `main`. Recommending a minor bump to `1.32.0` (new opt-in capability, matching this repo's `### Added` convention) with a customer-facing CHANGELOG line — happy to add that as a follow-up commit if useful.

Nice work tracking down why auto-reading Claude Code's managed `modelPricing` setting isn't the "cheapest" option here — that section in ADVANCED.md is a good explanation for future readers.

chrisdhaan (approved):

PR #542, opened 2026-09-01
Feat: GitHub Copilot desktop app support + platform-detection fix

### LINKED ISSUE #539: ClaudeCodeAdapter.isSupported() checks env vars that real Claude Code hook processes don't set
## Problem

`ClaudeCodeAdapter.isSupported()` (`src/platforms/claude-code-adapter.ts`) checks:

```ts
process.env.CLAUDE_CODE !== undefined ||
process.env.CLAUDE_CODE_VERSION !== undefined ||
process.env.MCP_CLIENT === 'claude-code'
```

None of these match the environment variables Claude Code's own CLI/hook processes actually set. A real Claude Code session sets `CLAUDECODE=1` (no underscore between CLAUDE and CODE), `CLAUDE_CODE_ENTRYPOINT`, and `CLAUDE_CODE_SESSION_ID` — verified directly against a live session's environment. `CLAUDE_CODE` and `CLAUDE_CODE_VERSION` are not present at all.

## Impact

Any code path that calls `PlatformRegistry.detect()`/`getActive()` to identify a genuine Claude Code process by ambient environment (rather than an explicit `MCP_CLIENT`/`NEW_RELIC_AI_PLATFORM` override) will fail to match `ClaudeCodeAdapter` and fall through to `GenericMcpAdapter`, the unconditional catch-all registered last. This was first surfaced during review of PR #483, which added a registry-based platform-detection call to the hook-write path in `collector-script.ts` — every Claude Code hook event came out tagged `platform: "generic-mcp"` instead of `claude-code`. That specific call site is being reworked to avoid the registry entirely, but the underlying adapter bug is separate and will resurface anywhere else ambient Claude Code detection is relied on.

## Suggested fix

Update `ClaudeCodeAdapter.isSupported()` to check the real environment variables, e.g.:

```ts
process.env.CLAUDECODE === '1' ||
process.env.CLAUDE_CODE_SESSION_ID !== undefined ||
process.env.CLAUDE_CODE_ENTRYPOINT !== undefined ||
process.env.MCP_CLIENT === 'claude-code'
```

Worth auditing whether `CLAUDE_CODE_VERSION` is ever actually set by any supported Claude Code surface, or should be dropped/replaced.

adjohn:
## Why

The GitHub Copilot desktop app is a Rust GUI over a warm pool of Copilot CLI stdio processes, verified live on macOS (app v1.1.14, commit 0d498e8). Its pooled CLIs fire the standard PascalCase hooks, so Preflight already captured every tool call — but those sessions carried zero economics (`tokensInput: 0`, `estimatedCostUsd: null`) and filed under `platform: "generic-mcp"`. The same mislabeling hit every Claude Code session on a dogfooding machine (#539): the adapter checks env vars Claude Code does not actually set. This PR makes the app a first-class platform and fixes the label for everyone.

## Scope

- `Fix #539`: `ClaudeCodeAdapter.isSupported()` now checks `CLAUDECODE` (the env var current Claude Code really sets, kept in one exported `CLAUDE_CODE_ENV_SIGNALS` const). `collector-script.ts` ambient-stamps `claude-code` on hook events in addition to the existing explicit `MCP_CLIENT` stamp; `HookEventProcessor` adopts non-generic stamps per session (`stampedAdapterBySession`) for tool-name mapping and exposes the last non-generic stamp as `activePlatform`; `mergeSummaries` refuses to let a later `generic-mcp` checkpoint clobber a real label; the dashboard health snapshot gains `platformDetectionFellBack`.
- New `CopilotAppAdapter` (`copilot-app`, `full-hooks`), registered before `copilot-sdk`, reusing the now-exported `COPILOT_SDK_TOOL_MAP` (the pooled CLI is the same runtime). Ambient detection keys on `~/.copilot/data.db` — created only by the app; `session-state/` is not a discriminator because the app's own pooled CLIs create it after the first session (verified live). `rename_session`/`rename_branch` are deliberately unmapped, originals preserved.
- New `CopilotAppUsageWatcher`: polls `data.db` read-only via `node:sqlite`, diffs the `sessions` table's cumulative token totals against a durable per-session cursor, and emits `mode: 'token'` buffer lines — the app's session id is byte-identical to Preflight's, so actions (hooks) and economics (DB) join with no correlation key. Deterministic messageIds dedup crash replays; shrinking totals re-baseline silently; any SQLite failure degrades to zero emissions. Gated by `NR_AI_ENABLE_COPILOT_APP_USAGE_WATCHER`, default on.
- `docs/ADAPTERS.md`: new `copilot-app` section; the `copilot-sdk` "desktop app untested" hedge replaced with the confirmed finding and the two CLI assumptions that fail on the app.

## Tradeoffs

- `total_nano_aiu` (GitHub's own billing meter) is deliberately not read yet — its semantics, and its overlap with `total_agent_merge_nano_aiu`, are unestablished. Noted in the watcher's module doc as the follow-up.
- The cache-inclusive `total_input_tokens` split is inherited from the verified VS Code pipeline, not independently verified against the app DB; if wrong it under-reports plain input, never double-bills.
- Reasoning-token deltas are cursor-tracked but not emitted (`TokenEvent` has no reasoning field).
- The watcher hand-mirrors the `CopilotUsageWatcher` template rather than extracting a shared base; all four existing watchers share that shape by copy, so the extraction is a separate refactor.

## Blast Radius

- Collector hot path gains three env-var lookups; no registry import (the <5ms budget holds).
- On existing installs, sessions that previously filed as `generic-mcp` start labeling as `claude-code`/`copilot-app` — platform comparisons and the weekly digest will show the shift, which is the point: `full-hooks` data stops being caveated as `self-reported`.
- The new watcher is default-on but no-ops on a single `existsSync` when no `data.db` exists.

## Verification

- `npm run build`, `npm test` (179 suites / 6,100 tests), `npm run lint` (0/0), `prettier --check` — all green on the assembled branch; the pre-push hook re-ran the suite.
- Live smoke against real artifacts: the compiled collector stamps `copilot-app` (explicit env), `claude-code` (ambient `CLAUDECODE`), and nothing (no signal); the compiled watcher, pointed at the live `~/.copilot/data.db`, emitted one token event per real app session whose numbers reconcile exactly with the DB rows (607,956 − 534,400 = 73,556 and 121,431 − 108,032 = 13,399 uncached input) and emitted nothing on a second poll.

https://claude.ai/code/session_01Ee6uLJ83sagHsxGf3Q4jVu


chrisdhaan:
## Triage review

**Sense-check: passes.** In scope — multi-platform adapter support is this project's core differentiator. No duplicate PR/issue found.

**Tracked issue:**
- The platform-detection fix references `Fix #539` — verified, matches exactly, and confirmed not yet fixed independently on `main`.
- The new `CopilotAppAdapter` + `CopilotAppUsageWatcher` had no tracking issue — created #560 for it.

**Build/test/lint (on the assembled branch):**
- `npm run build` — clean
- `npm test` — 179/180 suites, 6,103/6,107 tests passing (3 skipped, 1 todo — normal)

**Findings:**
1. **issue** — Branch is stale vs. `main` (16 commits / ~13 releases behind: `1.19.0` vs `1.32.1`). Needs a rebase/merge before merge.
2. **suggestion** — `platformDetectionFellBack`'s actual computation (`src/index.ts`) has no test coverage — only a hardcoded `false` is passed into a mock snapshot in `api-handler.test.ts`.
3. No `src/shared/` edits. `mergeSummaries`'s real-label-wins-over-generic-mcp logic, the per-session `stampedAdapterBySession` FIFO, and the usage watcher's shrink/re-baseline + dedup design all check out on inspection.
4. `docs/ADAPTERS.md` is thorough and honest about unconfirmed edges (Windows/Linux, cloud-sandbox sessions, concurrent app sessions).

**Recommended version bump:** `1.33.0` (minor — bundles a `Feat` with a `Fix`).

Great work overall — thanks for the thorough live verification and docs. I'll bring the branch up to date with `main` via a merge commit and push a couple of follow-up commits (version/changelog + a test for the coverage gap above) so this is ready to merge.

chrisdhaan (approved):

PR #543, opened 2026-09-01
Add repo_url attribution + recommended NR Scorecard rules

### LINKED ISSUE #558: Add repo_url attribution field + New Relic Scorecard rules for team dashboards
## Problem

Preflight already attributes every event/metric with \`team_id\`/\`project_id\`/\`org_id\`, where \`project_id\` carries a short \`org/repo\` form. There's no field carrying the full git remote URL, which limits how precisely telemetry can be attributed to a specific repo (e.g. distinguishing forks, or resolving ambiguous short names) without custom entity synthesis.

Separately, there's no documented set of New Relic Scorecard rule definitions for tracking AI-coding metrics (cost, anti-pattern rate, efficiency score, cost-per-file, security alerts) per team — something Scorecards can do today via NRQL against Preflight's existing custom events, without needing entity synthesis.

## Proposed fix

- Add a \`repo_url\` custom attribute to every NR event and metric, alongside the existing \`team_id\`/\`project_id\`/\`org_id\`. Auto-infer it from \`git remote get-url origin\`, credential-strip it via the existing \`redactSensitive()\`, and gate it behind the same opt-out as \`project_id\`.
- Add \`docs/SCORECARDS.md\` with ready-to-use Scorecard rule definitions (NRQL + rationale) for the metrics above.

## Open question

\`repo_url\` is more sensitive than \`project_id\` (it can reveal an internal git hostname). Worth deciding whether it needs its own dedicated opt-out, or whether coupling it to \`project_id\`'s existing toggle is the right default.

adjohn:
## Summary

Two small, independent additions for attributing Preflight's telemetry to New Relic Teams and repos, without needing custom entity synthesis:

- **`repo_url` attribution field**: a new custom attribute on every NR event and metric, alongside the existing `team_id`/`project_id`/`org_id`. Carries the full git remote URL (`project_id` already carried the shorter `org/repo` form). Auto-inferred from `git remote get-url origin`, credential-stripped via the existing `redactSensitive()`, and gated behind the same opt-out as `project_id`.
- **`docs/SCORECARDS.md`**: five ready-to-use New Relic Scorecard rule definitions (NRQL + rationale) for tracking AI-coding cost, anti-pattern rate, efficiency score, cost-per-file, and security alerts per team — no entity synthesis needed, since Scorecards can run arbitrary NRQL against Preflight's existing custom events.

## Notable fixes found during review (see `decisions.tsv` for the full trail)

- `repo_url` was wired through all 14 event-emitting call sites but initially missed the two metric-dimension builders (`teamDims`/`teamAttrs`) — caught by an independent cross-model review, not by build/test/lint (the field is optional everywhere, so nothing failed structurally). Fixed, with regression tests.
- `repoUrl` was missing from `ConfigFileSchema`'s zod shape — the value still worked via `.passthrough()`, but a user setting it in `config.json` would see a spurious "Unknown keys in config file (ignored)" warning.
- `getGitRemoteUrl()` didn't clear `GIT_DIR`/`GIT_WORK_TREE`, which git sets for hook subprocesses — under any hook/CI context (this repo's own `.husky/pre-push`, for one) that silently redirects the git-remote read to the wrong repo. Mirrors a fix already proven elsewhere in this codebase (`local-session-aggregator.ts`'s `GIT_OPTS`).
- Consolidated the ~13 duplicated `team_id`/`project_id`/`org_id` attach blocks in `nr-ingest.ts` into one shared, now-exported `attachTeamAttribution()` helper, also adopted by `audit-trail.ts`.

## Open question for a human, not resolved in code

`repo_url` ships on by default, gated only by the same opt-out as `project_id` (no dedicated toggle), and `preflight install`'s setup wizard doesn't prompt for it. `PRIVACY.md` itself calls this field "more sensitive" than `project_id` (it can reveal an internal git hostname). Worth a product decision on whether it needs its own opt-out before merge, or whether coupling it to `project_id`'s existing toggle is the right default.

## Test plan

- [x] `npm run build`
- [x] `npm test` (6050 passed, full suite, twice consecutively for flake-check)
- [x] `npm run lint` (0 errors, 0 warnings)
- [x] `npm run format:check`
- [x] Pushed via `git push`, exercising the repo's own `.husky/pre-push` hook end to end


chrisdhaan:
## Triage review

**Sense-check:** passes — in scope, no duplicate PRs/issues.

**Tracked issue:** none existed; created #558.

**Build/test/lint:** all green on this branch — `npm run build` clean, `npm test` 6050/6050 passed, `npm run lint` 0 errors/0 warnings.

### Findings

- **issue — branch is stale relative to `main`** (was 15 commits / 13 minor versions behind, several touching the same files this PR edits). Resolving via a merge commit rather than a rebase, to avoid rewriting this branch's history.
- **issue — `decisions.tsv`** is a verbatim internal AI-agent audit trail (delegate/session IDs, references to an unavailable "architect skill", etc.) with no relevance to the shipped product. Will be dropped in a follow-up commit.
- **nit — test name overstates its guarantee.** `config.test.ts`'s `"repoUrl is null when projectId is null, even if repoUrl is set explicitly"` doesn't actually exercise an explicit-null override — `file.projectId: null` falls through to `inferProjectId()` (pre-existing, unrelated to this PR) and only resolves to null because the test's `cwd` is a non-git tmpdir. Harmless, just a slightly misleading name — will be clarified.
- **Open product decision (not resolved here):** `repo_url` ships on by default, gated only by `project_id`'s existing opt-out, with no dedicated toggle and no setup-wizard prompt — despite `PRIVACY.md` itself calling this field more sensitive than `project_id` (can reveal an internal git hostname). Leaving this open for discussion before merge.

**Verified good:** independently confirmed the self-reported fixes in `decisions.tsv` against the actual diff rather than trusting the narrative — `repo_url` correctly reaches both `teamDims` and `teamAttrs` metric builders, `ConfigFileSchema` now includes `repoUrl`, `getGitRemoteUrl()` correctly clears `GIT_DIR`/`GIT_WORK_TREE`, and the credential-redaction regex does strip `user:token@host` from a git URL. `src/shared/` boundary respected. Feature completeness and test coverage are solid.

### Version bump + CHANGELOG

Target base is `main` at 1.32.0 → this should ship as **1.33.0** (Added). Will be handled in a follow-up commit along with the fixes above.

Thanks for the thorough writeup and self-review trail on this one!

chrisdhaan:
@adjohn — following up on the open question in the PR description about `repo_url`'s privacy posture.

Agreed it's a genuine design decision, not a bug, and worth resolving before merge rather than shipping as-is. Since `repo_url` can carry a full internal git hostname where `project_id` only ever carries `org/repo`, here are a few alternatives that keep most of the value while narrowing that specific exposure — curious which direction you'd lean:

1. **Default off, with its own opt-in.** Ship `repo_url` disabled unless a user explicitly turns it on, rather than riding along on `project_id`'s existing toggle. Most direct fix, and consistent with how this repo already treats anything riskier than baseline metadata (`recordContent`, `highSecurity`) as opt-in rather than opt-out.
2. **Host allowlisting.** Send the full URL verbatim only for known public hosts (`github.com`, `gitlab.com`, `bitbucket.org`, ...); for anything else, redact just the host segment (e.g. `https://[redacted-host]/org/repo`) and keep the path. This targets the actual stated risk — an internal git server's hostname — without losing org/repo distinguishing power for the common OSS case.
3. **Hash the host instead of stripping it.** If cross-dashboard host correlation matters (e.g. "same internal GHE instance as that other repo?"), a stable truncated hash of just the host preserves that joinability without exposing the literal hostname.
4. **Surface it in the setup wizard**, independent of whichever default is chosen. `projectId` already gets an explicit prompt during `preflight install`; `repoUrl` currently doesn't, so someone can end up sending it without ever having made an active choice. Even kept coupled to `project_id`'s opt-out, giving it its own wizard line (or just echoing the resolved value back) turns this into an informed decision instead of a silent default.

My inclination is 1 + 4 together — safest default, and a small change — but wanted your take before anyone touches the code, since you're the one who flagged the tradeoff in the first place. Happy to go a different way if one of the other options fits better with how you're thinking about this.

chrisdhaan:
Landed. Went with **1 (plain toggle) + 4 (wizard prompt)**, default-on, exactly as we converged on:

- `repoUrlEnabled` config field / `NEW_RELIC_AI_REPO_URL_ENABLED` env var — decoupled from `project_id`'s opt-out entirely, so setting `projectId: null` no longer silently disables `repo_url` too.
- `preflight install` now prompts for it explicitly, mirroring the existing `projectId` prompt, defaulting to on.
- Dropped the host-allowlist/hashing ideas per your point about self-hosted GitLab/GHE being indistinguishable from a public host by name alone — agreed that added complexity without covering the actual risk.
- `PRIVACY.md` and the CHANGELOG updated to describe the dedicated toggle instead of "no separate toggle, coupled to project_id."

Build/lint/tests all green (6311 passed). Ready to merge from my side.

chrisdhaan:
## Final code review

Ran a full adversarial review of everything in this PR (4 independent passes: core logic/config wiring, the setup-wizard + its test sequences, docs/CHANGELOG accuracy, and security/release hygiene). Found and fixed 4 real issues, all pushed:

1. **`retryAlertToNrEvent()` never got `repo_url`.** It was the one event-builder never migrated to the shared `attachTeamAttribution()` helper, so every `AiRetryAlert` event shipped without it — contradicting the "every NR event" claim. Fixed, threaded through `ingestRetryAlert()`, added a regression test.
2. **Credential redaction only covered the git-auto-inferred path.** An explicit `NEW_RELIC_AI_REPO_URL` env var or `repoUrl` config-file value skipped `redactSensitive()` entirely — so a URL pasted with embedded credentials would ship verbatim, contradicting `PRIVACY.md`'s unconditional "credentials are stripped" claim. Redaction now applies uniformly to all three sources; added tests for both override paths.
3. **The new `repoUrlEnabled` wizard prompt had inconsistent yes/no parsing.** Unlike every sibling `[Y/n]` prompt in the wizard, garbage input silently disabled the toggle even when the shown default was Y. Fixed to match the adaptive-default pattern used everywhere else; added 3 tests (decline, blank-preserves-existing-false, garbage-preserves-Y-default).
4. **`docs/METRICS_TABLE.md` still said `repo_url` "shares project_id's opt-out" in 9 places** — stale from before the dedicated toggle existed.

Build/lint/format/full suite (6317 passed, +6 new tests) all green; pre-push hook re-verified everything. Also re-confirmed real mergeability against the latest `main` (which picked up #566 since the last push) via a local `git merge-tree` dry run, not just GitHub's cached field — clean, no conflicts.

Nothing else outstanding on my end.

chrisdhaan (approved):

PR #545, opened 2026-09-01
Add Kiro Power support + fix session_id resolution behind wrapped launchers

### LINKED ISSUE #570: Kiro sessions: session_id fails to resolve behind wrapped launchers, misdetected as generic-mcp, tool-map casing wrong
### Summary

Several related bugs affect Kiro (and potentially other launchers that wrap the MCP server process):

1. **`session_id` never resolves when a launcher interposes a process above the MCP server.** `npx` runs `npm exec`, which runs the server — so the server's `process.ppid` no longer matches the breadcrumb the hook collector wrote at its own ppid. Resolution silently fails while the MCP connection still reports healthy.
2. **Kiro sessions record as `platform: "generic-mcp"` instead of `"kiro"`.** Kiro passes the MCP server 16 env vars, none of which identify it as Kiro, so `KiroAdapter.isSupported()` never fires and detection falls through to the generic fallback — leaving tool names unmapped.
3. **`KIRO_TOOL_MAP` uses wrong casing and field names.** The map assumes camelCase tool names (`readFile`, `strReplace`) and a `file_path` field, but Kiro actually sends snake_case tool names and puts the file path under `path` (edits use `oldStr`/`newStr`). This silently zeroes out `unique_files_read`, `unique_files_modified`, `bash_commands_run`, and `search_queries` while the raw tool-call count still looks correct.
4. **No `tools/list_changed` notification.** Clients cache `tools/list` at initialization. When `session_id` can't resolve synchronously, the server exposes only a pending/partial tool set and swaps in the full one shortly after — with nothing notifying the client. Kiro ends up listing only 3 tools for an entire session.

### Environment

- Platform: Amazon Kiro (Agent Plugins / Kiro Power format)
- Affects any wrapped launch (e.g. via `npx`) on any platform — not Kiro-specific for items 1 and 4.

RuchikaBakolia:
## Summary

- **Packages Preflight as a Kiro Power** (`kiro-power/`) in Kiro's current [Agent Plugins](https://agent-plugins.org) format — `plugin.json` + `mcp.json` + two Agent Skills (`setup`, `observability`). This format is required, not preferred: kiro.dev's submission page states *"Please create your power using the Agent Plugins format"* for their curated registry.
- **Fixes `session_id` never resolving when a launcher interposes a process above the MCP server.** `npx` doesn't run the server — it runs `npm exec`, which does. The server's `process.ppid` therefore stops matching the breadcrumb the hook collector writes at *its* ppid, and resolution silently fails while the MCP connection still reports healthy. New `src/hooks/process-ancestry.ts` walks a few levels of the server's own ancestry (`/proc` on Linux, one `ps` call on macOS/POSIX, no-op on Windows = today's behaviour). **Not Kiro-specific** — this affects any wrapped launch on any platform.
- **Fixes Kiro sessions recording as `platform: "generic-mcp"`.** Kiro passes a Power's MCP server 16 env vars and none identify it, so `KiroAdapter.isSupported()` never fired and detection fell through to the generic fallback — leaving tool names unmapped. Pinned `NEW_RELIC_AI_PLATFORM=kiro` in the Power's `mcp.json`, one of the adapter's four documented signals.
- **Corrects `KIRO_TOOL_MAP` casing and `tool_input` field names** against payloads captured from a live install. The map guessed camelCase (`readFile`, `strReplace`); Kiro sends snake_case. Kiro also sends the file under `path` (not `file_path`) and edits as `oldStr`/`newStr`. Together these silently reported **zero** for `unique_files_read`, `unique_files_modified`, `bash_commands_run` and `search_queries` while the raw tool-call count still looked correct.
- **Adds `tools/list_changed`.** Clients cache `tools/list` at initialization. When `session_id` can't resolve synchronously the server exposes only the pending tool set and swaps in the full one milliseconds later — with nothing notifying the client. Kiro listed 3 tools for an entire session while the rest were registered server-side. **Also not Kiro-specific.**

Kiro's meta tools (`kiro_powers`, `update_session_information`, `createHook`) are deliberately left unmapped — they're session/power management, not file or shell operations, and mapping them to a file verb would inflate file metrics with activity that never touched a file.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npm test` — 6108 passing. One pre-existing failure (`api-handler.test.ts`, DST-transition test) which I confirmed **also fails on unmodified `upstream/main`** (1 failed / 255 passed there too)
- [x] 35 new unit tests: `process-ancestry.test.ts` (23) + `session-resolver.test.ts` (12), covering the real Kiro tree shape, closest-match-wins, depth cap, cycle guards, stale-PID rejection, malformed `/proc`, `ps` absent/failing, and each platform branch
- [x] 6 new collector tests from verbatim captured Kiro payloads, including a regression test asserting a Grep search root is **not** promoted to `file_path`
- [x] 2 new server tests: `tools.listChanged` advertised, and notification actually delivered to a connected client over a real transport
- [x] **Real un-mocked process tree:** planted a breadcrumb at the real grandparent PID with the immediate parent bare — pre-fix lookup returns `null`, fixed resolver finds it at depth 1 in 154 ms; walked chain cross-checked independently against `ps`
- [x] **Regression on the normal path:** breadcrumb at the direct parent still resolves via the synchronous fast path in **1 ms**, no subprocess spawned
- [x] **Live in Kiro:** verified `session_id` resolves with the `npm exec` wrapper present (`Kiro Helper → npm exec → server`), `platform: "kiro"`, tool names normalized to `Read`/`Edit`/`Glob`/`Bash`, and the full tool list appearing after the `listChanged` fix


chrisdhaan (commented):
Thanks for this — really thorough work, especially the process-ancestry walk (cycle guards, platform degradation, all traced to a live install) and the `tools/list_changed` fix. Both are correctly identified as not Kiro-specific.

**Verification:**
- `npm run build && npm test && npm run lint` — all clean (6109 passed, 0 failed, 0 lint issues)
- Confirmed none of the four bugs (session_id resolution behind a wrapper, `generic-mcp` misdetection, tool-map casing, missing `tools/list_changed`) are already fixed on `main`
- Tracked under #570

**A few things I'm fixing directly rather than sending back for changes:**
- Branch was 18 commits behind `main` — merging `main` in to bring it current.
- `kiro-power/plugin.json`'s version (`1.16.4`) doesn't match the package version — this repo keeps plugin manifest versions in lockstep with `package.json`.
- A stray `.prettierignore` entry (`examples/*.html`) doesn't correspond to any file in the repo.
- Adding the version bump + CHANGELOG entry for release.

Nothing else needed — solid PR. Appreciate the depth of the testing writeup too.

chrisdhaan (approved):

PR #546, opened 2026-09-01
fix(web): tooltip item text hardcoded to black on dark charts

### LINKED ISSUE #559: Tooltip item text renders black on dark charts (Cost Per Outcome, Top Tools, Cost by Tool)
## Problem

Cost Per Outcome and Top Tools (History) and Cost by Tool (Today) color each bar per entry with `<Cell fill={...}>`, leaving the `<Bar>` element itself with no `fill`/`stroke` prop. Recharts' `DefaultTooltipContent` falls back to a hardcoded `#000` for the tooltip item row whenever the series' `color` is undefined.

The theme-aware ink token (`TOOLTIP_STYLE.color`) only reaches the tooltip label via `contentStyle`, not the item row — so on dark charts, the value line renders as near-black text on a dark tooltip background while the label stays readable.

## Affected panels

- `src/web/views/History.tsx` — Cost Per Outcome, Top Tools
- `src/web/views/Today.tsx` — Cost by Tool

## Proposed fix

Add a dedicated `itemStyle` (e.g. `TOOLTIP_ITEM_STYLE` / `CHART_TOOLTIP_ITEM_STYLE`) using the same ink token, and pass it to the affected `<Tooltip>` elements. Other `<Tooltip>` elements in these files are unaffected since their series already have an explicit `fill`/`stroke`.

adjohn:
## Why

Cost Per Outcome and Top Tools (History) and Cost by Tool (Today) color each bar per entry with `<Cell fill={...}>`. That leaves the `<Bar>` element itself with no `fill`/`stroke` prop. Recharts' `DefaultTooltipContent` falls back to a hardcoded `#000` for the tooltip item row whenever the series' `color` is undefined. `TOOLTIP_STYLE.color` (a theme-aware ink token) only reaches the tooltip label via `contentStyle`, not the item row, so the value line rendered as near-black text on the dark tooltip background while the label stayed readable.

## Scope

- `src/web/views/History.tsx`. Add `TOOLTIP_ITEM_STYLE` and pass `itemStyle={TOOLTIP_ITEM_STYLE}` on the Cost Per Outcome and Top Tools `<Tooltip>` elements.
- `src/web/views/Today.tsx`. Same fix for the Cost by Tool panel's `<Tooltip>`, via a new `CHART_TOOLTIP_ITEM_STYLE`.
- Left every other `<Tooltip>` in both files untouched. Their item text is already colored by an explicit `Bar`/`Area` `fill` or `stroke` (either a real theme color, in which case matching the series color in the tooltip is deliberate, or a gradient URL that the browser drops as an invalid `color` value and falls back to the inherited ink token) — neither case is affected by this bug.

## Blast Radius

Purely additive `itemStyle` props on three `<Tooltip>` elements plus two new style constants. No behavior change to any chart that wasn't showing black-on-black text.

## Verification

- Built the fixed dashboard and drove it with a headless Chromium instance, hovering the affected bars and reading `getComputedStyle` on the tooltip. In dark mode, item color went from `rgb(0, 0, 0)` to `rgb(255, 255, 255)` on both charts, matching the label and content color and matching the already-correct Daily Spend chart used as a control.
- `npx tsc -p tsconfig.web.json --noEmit`: same 5 pre-existing errors before and after this change (none introduced).
- `npx vitest run src/web/views/History.test.tsx src/web/views/Today.test.tsx`: 151 passed.
- Full `npm test` (husky pre-push hook): 6055 passed, 3 skipped, 1 todo, 0 failed.
- Did not add an automated regression test: reproducing this requires a real browser resolving CSS custom properties plus seeded historical session data to populate these two charts, which is integration-heavy relative to the fix size. Verified live instead and captured the before/after computed-style values above.


chrisdhaan (commented):
## Triage review

**Sense-check:** passes — real bug, small precise fix, in scope, doesn't touch `src/shared/`.

**Tracked issue:** none existed for this. Filed #559 from this PR's description.

**Build/test/lint (run locally against this branch):**
- `npm run build` — passes
- `npm test` — 6055 passed, 3 skipped, 1 todo, 0 failed (matches the numbers reported in the PR description)

**Review:**
Verified the root-cause claim directly in the code — `Cost Per Outcome` and `Top Tools` (`History.tsx`) and `Cost by Tool` (`Today.tsx`) each color bars per-entry via `<Cell fill=...>`, leaving the `<Bar>` itself with no `fill`/`stroke`, so Recharts' tooltip item row falls back to black. Checked every other `<Tooltip>` in both files and confirmed each of those sits on a `Bar`/`Area` with an explicit `fill`/`stroke` (a direct color or a gradient `url(#...)`) — so this PR's 3-tooltip scope is complete, and nothing else needed the same fix. No bugs found, no missing coverage beyond what's already explained in the PR description (the manual before/after computed-style verification is a reasonable substitute for a fix this size).

Only note: the branch is currently 12 commits behind `main`. Bringing it up to date via a merge commit now, followed by a separate commit with the version bump + CHANGELOG entry for this fix.

chrisdhaan:
Tracked in #559.

chrisdhaan (approved):

PR #557, opened 2026-09-02
Fix: GitHub Copilot tool-call capture and automate its install

### LINKED ISSUE #586: GitHub Copilot hooks file uses Claude Code's nested JSON shape, so Copilot's hooks-runner never executes any hook
### Summary

The Copilot hooks file generated by the installer (`generateCopilotHooksFile()` in `src/install/cli.ts`) uses Claude Code's nested settings.json hook shape:

```json
"PreToolUse": [{ "matcher": "", "hooks": [{ "type": "command", "command": "preflight-collector pre-tool" }] }]
```

GitHub Copilot's actual hooks schema is flat — each array entry is the command hook object directly:

```json
"PreToolUse": [{ "type": "command", "command": "preflight-collector pre-tool" }]
```

Both shapes are valid JSON, so nothing crashes or errors — Copilot's hooks-runner just silently never invokes the commands. This breaks tool-call count, tool selection, latency, audit, and session tracking for every Copilot session. Cost tracking kept working because it goes through the separate usage-watcher path, which masked the problem.

### Related

There's also a platform-attribution bug: a Copilot session drained by an unrelated already-running preflight process gets tagged with that draining process's own platform (e.g. `generic-mcp`) instead of the session's real platform, because platform was being stamped via the adapter registry instead of directly from `MCP_CLIENT`/`NEW_RELIC_AI_PLATFORM` env vars at hook-write time.

### Repro

1. Configure the Copilot adapter per `docs/ADAPTERS.md` and generate the hooks file via the installer.
2. Trigger tool calls in Copilot CLI or VS Code Copilot Chat.
3. Observe: no `PreToolUse`/`PostToolUse` hook activity reaches Preflight's buffer, despite the hooks file appearing well-formed.

### Suggested fix

Emit the flat `{type, command}` shape Copilot's hooks-runner actually expects, and stamp platform directly from env vars at hook-write time rather than via the adapter registry.

jcountsNR:
Fixes #586

## Summary

- Fixes a critical bug where the Copilot hooks file this installer generates used the wrong JSON shape (Claude Code's nested `{matcher, hooks: [...]}` instead of Copilot's flat `{type, command}`), which made Copilot's hooks-runner silently never execute any hook — breaking tool-call count, tool selection, latency, audit, and session tracking for every Copilot session, while cost tracking kept working via the separate usage-watcher path (masking the problem).
- Fixes a related platform-attribution bug: a Copilot session drained by an unrelated already-running preflight process was tagged with that *draining* process's own platform (e.g. `generic-mcp`) instead of the session's real platform.
- Fixes the underlying regression from the previous "Fix for copilot connection to NR" commit: platform is now stamped from `MCP_CLIENT`/`NEW_RELIC_AI_PLATFORM` env vars directly at hook-write time (the hooks-runner doesn't inherit env set on the MCP server registration), instead of via the adapter registry, which mis-tagged Claude Code events as `generic-mcp`.
- Automates the entire GitHub Copilot setup end-to-end: `preflight install --copilot` (and a new prompt in `preflight setup`) now configures the Copilot CLI hooks + MCP registration, VS Code Copilot Chat's `mcp.json`/`settings.json` (including enabling token-exact cost logging), and VS Code's hook double-capture fix — plus the corresponding `--copilot` uninstall path. Previously all of this was manual, undocumented-in-practice, and had the JSON-shape bug above.
- Corrects the setup instructions in both Copilot adapters and `docs/ADAPTERS.md` to match the real (flat) hooks schema and the new automation.

## Root cause (the big one)

`generateCopilotHooksFile()` was generating Claude Code's own settings.json hook shape instead of GitHub Copilot's actual schema. Both shapes are valid JSON, so nothing crashed or errored — Copilot's hooks-runner just quietly never invoked the commands. Verified live against a real New Relic account: after the fix, a simulated hook cycle correctly produced `platform: "copilot"`, `toolCallCount`, `toolSelectionMetrics`, and a `timeline` entry with real `durationMs`.

## Test plan

- [x] `npm run build && npm test` — 5938 passed, 0 failed
- [x] Live verification: ran `preflight install --copilot` against a real machine with both Copilot CLI and VS Code Copilot Chat, confirmed corrected hooks file, `copilot mcp list`, VS Code `mcp.json`/`settings.json`; triggered real tool calls in both hosts and confirmed `platform: "copilot"`, tool call count, tool selection, latency, and audit entries all populate correctly; confirmed via New Relic query
- [x] Ran `preflight uninstall --copilot` and `preflight uninstall` and confirmed full, clean reversion (backups taken beforehand) without disturbing unrelated Claude Code/VS Code config
- [x] Full reset-and-reinstall test via `preflight setup` from a genuinely blank `~/.newrelic-preflight/` and stripped hook/MCP config, confirming the wizard's new Copilot prompt works end-to-end


chrisdhaan:
## Review

Thanks for this — the root cause writeup and live verification notes are excellent. Ran the full local verification suite against this branch:

- `npm run build` — succeeds
- `npm test` — **6276 passed, 0 failed** (3 skipped, 1 todo)

### Verified

- The hooks JSON-shape bug is real and the fix is correct: `generateCopilotHooksFile()` now emits Copilot's flat `{type, command}` shape instead of Claude Code's nested `{matcher, hooks:[...]}`. Confirmed this is the kind of bug that fails silently (valid JSON either way) rather than erroring.
- The platform-attribution fix in `LocalSessionAggregator` (preferring the record's own `platform` over the draining process's ambient platform) is correct and threads properly through the existing `preEvent.platform ?? event.platform` logic in `event-processor.ts`.
- No edits to `src/shared/` (vendored snapshot boundary respected).
- Security: `execFileSync` is used with argv arrays throughout for `copilot mcp add/remove` — never a shell string — so license keys and other values can't be shell-injected. No issues found here.

### Minor notes

- The PR description's third bullet ("platform is now stamped from env vars directly... instead of via the adapter registry") describes logic that was already in place before this PR (landed via #483) — this PR only updates the comment wording in `collector-script.ts`. Not a bug, just a slightly overstated description.
- The one-line change in `nr-ingest.ts` (`sessionTraceId: record.sessionId ?? this.sessionTraceId` at the `proxyToolCallToNrEvent` call site) is a no-op: `proxyToolCallToNrEvent()` already applies `record.sessionId ?? attrs.sessionTraceId` internally, so this doesn't change behavior either way. Harmless, but worth knowing it's redundant.

### Tracked issue

No open issue covered this specific root cause, so I filed and linked #586.

Tracked separately: this branch is currently well behind `main` (mergeable state is conflicting) and will need updated version numbers + a CHANGELOG entry once it's brought current — happy to help land that.

chrisdhaan:
## Update: merged main + follow-up fixes

Merged `main` into this branch (merge commit, not a rebase, to preserve your commit history) and pushed. Highlights:

- Resolved two real conflicts in `src/hooks/collector-script.ts` (kept upstream's `detectStampPlatform()`, which supersedes this branch's narrower version with the same fix) and `src/install/setup-wizard.test.ts` (combined the two independently-added wizard prompts — `repoUrlEnabled` from main, `installCopilotHooks` from this PR — into one consistent answer order).
- While verifying the merge, found and fixed a real bug the merge exposed: the new "setupWizard Copilot hooks install step" test block never picked up the extra `repoUrlEnabled` prompt slot (no conflict marker, since that block didn't exist on either side of the 3-way merge base), so `dashboardPort` was silently receiving the string `'n'` and calling the real `process.exit(1)` — crashing the whole Jest worker rather than failing one test. Fixed the 8 affected answer arrays.
- Removed a test this PR added that's now both redundant with and contradicted by upstream's own (more thorough) platform-stamping tests in the same file.
- Simplified a redundant, no-op fallback at the `proxyToolCallToNrEvent` call site in `nr-ingest.ts` (that function already applies the same fallback internally).
- Bumped all five version files to 1.40.0 and added a CHANGELOG entry.

`npm run build && npm test && npm run lint && npm run format:check` all pass clean (6481 tests, 0 failures). Left as separate commits so each change is easy to review independently.

chrisdhaan (approved):

PR #567, opened 2026-09-03
Docs: rebuild the site landing page, add What's New and an agent install prompt

adjohn:
Fixes #582

## Summary

- **Landing page rebuilt.** The old page was a Starlight splash with no link to any docs page. The new `site/src/pages/index.astro` is a custom page in the style of ai-sdk.dev: hero with the install command, a "Copy prompt" button for coding agents, a tabbed showcase of real MCP tool responses, the platform list from ADAPTERS.md, build-time counts of tools, metrics, dashboards, and platforms, feature cards linked into the docs, the dashboard gif, and a footer listing every page.
- **Copy-prompt install path.** The agent prompt points at `site/public/start.md`, a guide written for a coding agent: check Node, `npm install -g`, `preflight install --mode local`, `preflight doctor`, then hand back to the person for the restart.
- **What's New page.** `docs/WHATS_NEW.md` batches a run of releases into one plain-language edition, leading with what a reader has to do after upgrading. It is the first sidebar entry, the landing page pill links to it, and the README links to it. `npm run whats-new:draft` lists changelog entries newer than the last edition. The per-version CHANGELOG stays on GitHub.
- **One source for landing content and base path.** `site/src/landing.ts` holds the typed tables the page iterates. `BASE` is defined there and imported by `astro.config.mjs` and the verify script.
- **Self-hosted Geist and Geist Mono** via a shared `fonts.css` used by both the landing page and the Starlight docs. No third-party requests.
- **Theme carries across.** The landing toggle writes Starlight's `starlight-theme` key so a choice made on either half applies to the other.
- **Showcase reads as a conversation.** Each tab shows the question and the answer as the assistant would give it, figures in bold, with a small line naming the Preflight tool behind it. The announcement pill reads "What's new in v<version>" from package.json.
- **For teams section.** Org-level ROI: cost per outcome with waste ratio and ROI estimate, efficiency across developers, the five prebuilt alert conditions, and platform comparison, ending in a copyable connect command. The hero and meta description gain one sentence about rolling up across the org.
- **Platform logos and links.** Every platform pill carries its official mark, sourced and licensed per entry in `site/src/logos.ts`, and links to that platform's heading on the Adapters page. Single-color marks are inlined and take `currentColor`.
[...]

PR #568, opened 2026-09-03
feat(metrics): track Skill calls per skill name in cost attribution

### LINKED ISSUE #544: Track per-Skill usage (name, tokens, duration) instead of collapsing into a generic "Skill" tool bucket
## Summary

Every invocation of the `Skill` tool is recorded as a generic `Skill` tool call. The specific skill name is dropped before it reaches storage, so no cost, token, or duration breakdown can distinguish one skill from another — everything collapses into a single `"Skill"` bucket.

The skill name is already present in the hook payload today (`tool_input.skill` on every `Skill` invocation). This is a parsing gap in Preflight, not a gap in what Claude Code's (or Kiro's, or Cursor's) hooks expose.

## Where this happens

- `src/hooks/collector-script.ts`'s `extractInputMeta()` has cases for `Read`, `Write`, `Edit`, `Bash`, `Grep`, `Glob`, `Agent`, `TaskCreate`, `TaskUpdate`. There is no case for `Skill`, so the switch falls through and the field is dropped.
- `src/hooks/tool-parsers.ts`'s `INPUT_PARSERS` dispatch table has the same gap on the post-event side.
- `ToolCallRecord` (`src/storage/types.ts`) has no `skillName` field to carry it even if extracted.
- `TurnCostAttributor` (`src/metrics/turn-cost-attributor.ts`) keys its `costByToolType` map by raw tool name only, so all `Skill` calls across every session accumulate into one `"Skill"` entry.

Duration is not the problem. It's already computed per tool call from the pre/post hook timestamp delta and stored on every `ToolCallRecord`. Only the skill-name dimension is missing.

## Proposed fix

1. Add a `Skill` case to `extractInputMeta()` and to the `INPUT_PARSERS` table, extracting `tool_input.skill` (and optionally a truncated/hashed form of `tool_input.args` to avoid unbounded cardinality from arbitrary arguments).
2. Add an optional `skillName` field to `ToolCallRecord`.
3. Extend `TurnCostAttributor` to key by `(toolName, skillName)` when `toolName === 'Skill'`, instead of `toolName` alone.
4. Expose the breakdown via a new `nr_observe_get_cost_per_skill` MCP tool, or a `skill` facet added to the existing `nr_observe_get_cost_per_tool`.

## Why it matters

Exec-facing ROI reporting needs a breakdown like "Skill: `/code-review`, 40 calls, 1.2M tokens, 3.4h" rather than one aggregate "Skill" row. This is a real request from teams evaluating whether to keep using Preflight-managed skills versus writing their own instrumentation. The gap is small and self-contained: the data already exists at the hook boundary, it just needs to be threaded through the same pattern every other tool already uses.

adjohn:
## Why

Closes #544. Every `Skill` call reached Preflight as a bare `Skill` tool call because `extractInputMeta()` had no case for it, so the skill name in `tool_input.skill` never left the hook. Teams evaluating skills need "`/code-review`: 40 calls, 1.2M tokens, 3.4h", and the data was already at the boundary. The skill name now travels the same path every other tool-specific field takes, and the cost attributor keeps one row per skill.

Cost cannot be answered from New Relic alone. A tool call's cost is only known when the turn's token event arrives, after the `AiToolCall` event is serialized, and `getCostForToolCall()` has no production caller. So calls and duration per skill are available in NRQL from this release, and cost and tokens per skill come from the in-process attributor through the MCP tool.

## Scope

- `src/hooks/collector-script.ts`: `extractInputMeta()` gains `case 'Skill'`. It stores `skill` (capped at 128 chars) and `argsLength`. The argument text is never written.
- `src/hooks/tool-parsers.ts`: `parseSkill` in `INPUT_PARSERS` maps `skill` to `skillName` and `argsLength` (or a raw `args` string) to `skillArgsLength`.
- `src/storage/types.ts`: `ToolCallRecord` declares `skillName?` and `skillArgsLength?`.
- `src/transport/nr-ingest.ts`: `skillName` joins `REDACT_FIELD_KEYS`. Both fields already reach `AiToolCall` through the generic spread.
- `src/metrics/turn-cost-attributor.ts`: the per-tool `costByToolType` map becomes one attribution table keyed by a bucket identity derived once per record through `VARIANT_FIELD_BY_TOOL = { Skill: 'skillName' }`. Buckets count calls and duration at `recordToolCall` and cost, tokens, and attributed calls at `recordTokenEvent`. `getMetrics()` folds that table into the unchanged `costByToolType` and the new `costBySkill`.
- `src/tools/session-stats.ts`: description of `nr_observe_get_cost_per_tool` only. The handler is unchanged.
- `docs/COMMANDS_TABLE.md`, `CHANGELOG.md`, and version 1.34.0 in `package.json`, `package-lock.json`, `server.json`, and the plugin manifest.

Out of scope: splitting `Agent` by `subagentType` (one line in the same table, left for its own change), persisting `costBySkill` into session summaries for weekly rollups, and a dashboard card for skills.

## Tradeoffs

- `costBySkill.*.callCount` counts every call while `costByToolType.*.callCount` still counts attributed calls, as it always has. The skill row carries `attributedCallCount` beside it so the gap is visible instead of silently deflating cost.
- Tokens per skill are an even split of the turn's token event. That is the same approximation the shipped cost figure already makes.
- `args` survives only as a length. Truncation keeps the front of the string where flags and tokens live, and a hash is unique per invocation, so neither groups usefully.

## Blast radius

`costByToolType` is byte-identical: entries only for tools with attributed calls, same fields, same values. The dashboard's Cost by Tool card, the `/api/cost-per-tool` route, and the documented response are unaffected, and the existing tests for them pass unchanged. New attributes on `AiToolCall` appear only on `Skill` calls. Platforms whose adapters map a skill tool to `Skill` without forwarding a `skill` string (opencode, Kilo, Zed, Continue) keep counting under `costByToolType.Skill` and get no per-skill row.

## Verification

- `npx tsc -b .` and `npm run lint` pass with no output. The full jest suite passes: 183 suites, 6319 tests (upstream main baseline was 6303). A rebase with an exec step ran the build plus the attributor and parser tests at each of the three code commits, and each passed on its own.
- An end-to-end probe (`probe-544.mjs`, kept out of the repo) pushed real `Skill` PreToolUse/PostToolUse payloads copied from a Claude Code transcript through `processHook`, `HookEventProcessor.processEvents`, `TurnCostAttributor`, and `toolCallToNrEvent`. It asserted `skillName` and `skillArgsLength` on every record and event, no `args` text anywhere, one `costBySkill` row per skill with the expected duration and even-split tokens, `costByToolType` keys unchanged, the `Skill` total equal to the sum of skill rows, and an API-key-shaped `skillName` redacted at egress. It printed `probe-544 OK`.

https://claude.ai/code/session_01RThTYxoissgjSWJSfCdpqF


chrisdhaan (commented):
## Triage review

**Tracked issue:** Closes #544 (verified — real, open, directly relevant).

**Build/test/lint:** all pass locally. `npm run build` clean, `npm test` 183/184 suites (1 unrelated skip) / 6319 passing (matches the PR's own claimed numbers exactly), `npm run lint` 0 errors/warnings.

**Code review:** Reviewed the `TurnCostAttributor` bucket rework by hand — the `(toolName, skillName)` keying, the even-split-cost-per-tool-call carryover, the `costByToolType.Skill === sum(costBySkill rows)` invariant, and the multi-skill-per-turn case are all correct, and the tests exercise the real edge cases (unattributed skill call, two skills sharing a turn, fold-back into `costByToolType`). Redaction, storage type, and parser additions all follow existing conventions. No `src/shared/` edits.

One thing I want to flag transparently even though it turned out to be a non-issue: diffing this branch against local `main` initially looked like it deleted the `repoUrl` attribution feature (#543) from `nr-ingest.ts`. That was an artifact of the branch being stale relative to `main`, not a real change — diffing against the actual merge-base confirmed the real touch to that file is a single line (adding `skillName` to `REDACT_FIELD_KEYS`). No code change needed there.

**Branch staleness:** the branch was 5 commits behind `origin/main` (missing #543, #545, #566, #567). I merged `main` into this branch with a merge commit (no rebase, no force-push) to bring it current, resolving conflicts in the version/changelog files by taking `main`'s side. I then added a follow-up commit re-deriving the version bump against current `main` (which had advanced to 1.36.0 while this branch was in flight) — `1.36.0 → 1.37.0` (minor, since this is a `Feat`), with the CHANGELOG entry repositioned above the current top entry. Both pushed to this branch.

Overall: solid, well-tested implementation of #544. No blockers beyond what's now resolved above.

chrisdhaan (approved):

PR #574, opened 2026-09-03
Fix: Model Usage panel names a most-efficient model it doesn't list

adjohn:
## Summary

- The Today dashboard's Model Usage panel keeps the top 4 models by spend, but `mostEfficientModel` is picked by the tracker across every model used today. The cheapest model per output token is usually the one that spent the least, so its row was cut while the footer still named it (screenshot: four rows, footer says `claude-fable-5-1`).
- The panel now re-attaches the most-efficient model's row when it falls outside the top 4, so the footer always points at a visible row.
- The footer prints the rate it was judged on (`$/1M output tok`). That metric differs from the row column, which blends input and output tokens, and the tooltip now says so.
- Bumps to 1.34.1 with a CHANGELOG entry.

## Test plan

- [x] New Today view tests: most-efficient model outside the top 4 stays visible; already-listed case still caps at 4 rows. Both fail against the old panel code.
- [x] `npx vitest run` (39 files, 537 tests), `npx tsc -b .`, `npm run build:web`, `npm run lint`, `npm run format:check`
- [x] `test/plugin-manifest.test.ts` and `test/mcp-registry-manifest.test.ts` for the version bump
- [x] Full jest suite via the pre-push hook

https://claude.ai/code/session_01E7pumMBnRX61pcKEMbk56s


adjohn:
Closing: the code this PR fixes no longer exists on `main`.

This PR kept the "Most efficient" footer's model visible in the Model Usage panel's top-4 list. PR #590 (released as 1.44.0) removed that footer and the `mostEfficientModel` / `costPerOutputToken` fields entirely, because the ratio behind them was misleading. With no footer, the panel can no longer name a model it doesn't list, so there is nothing left to fix here.

The branch is kept in case any of the test scaffolding is worth salvaging. Reopen if a replacement for the footer lands and the same mismatch returns.

PR #548, opened 2026-09-02
Feat: Adoption & Cost dashboard + demo data generator

adjohn:
Fixes #589

## Summary

- New pre-built dashboard `dashboards/ai-coding-assistant-adoption-cost.json` (6 pages, 69 widgets): Adoption, Cost, Tools & MCP, Team Leaderboard, Team Pulse, and Models — a manager-facing view of adoption momentum, spend, and per-developer outcomes built entirely from Preflight's own events and metrics.
- MCP widgets cover both event shapes: hook-observed `mcp__server__tool` names on `AiToolCall` and proxy-routed `AiMcpToolCall`, with union billboards and labeled Hooks/Proxy breakdowns.
- Team Pulse includes git-outcome widgets built on the `ai.git.*` gauges from #484 and the rejection signal from #485 (per-session PR/commit maxima, edit accept rate, cost per PR); page markdown documents the cumulative-gauge semantics these queries rely on.
- Cost page documents why the `FROM Metric` token widgets stay blank on event-only data (they need a live server exporting metrics).
- New `scripts/generate-demo-data.ts`: seeds an account with realistic demo telemetry (10 developer personas, every event type plus cumulative metric snapshots matching live harvest semantics) for testing and demos. Flags: `--dry-run`, `--hours`, `--seed`, `--eu`, `--staging`. The ingest APIs drop points older than 48h, so re-running on later days accumulates weekly-trend history.
- The dashboard is registered everywhere the other seven are: the README dashboard list, the Terraform module (`terraform/dashboards.tf`), and the `deploy:dashboard:all` script picks it up from the directory automatically. Version bumped to 1.41.0 with a CHANGELOG entry.

## Test plan

- [x] `npx jest -- src/dashboard.test.ts src/deploy/deploy-dashboards.test.ts src/index.test.ts` — 185 pass (widget titles, known event types, empty accountIds, 12-column grid)
- [x] `npx eslint scripts/generate-demo-data.ts` and `npx prettier --check .` — clean
- [x] Generator `--dry-run` and a real `--staging` run (4,065 events + 5,018 metric points accepted); dashboard imported into a test account and verified rendering across all six pages, iterated on real import-validator errors (accountIds injection, no nested-FROM subqueries)
- [x] `npm run build` after merging `main` — clean, and `dist/data/dashboards/` contains the new file
- [x] All 65 `accountIds` in the new dashboard use the `[]`/`[0]` placeholders the deploy script and Terraform rewrite
[...]

PR #357, opened 2026-08-05
Feat: Recommendations panel in History view

chrisdhaan:
## Summary

- Adds `GET /api/recommendations` and wires `RecommendationEngine` into the dashboard
- Adds a `RecommendationsPanel` to the History view, synthesizing cost, prompt-engineering, and model-selection signals into a prioritized list

Closes #33

## Test plan
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm test`
- [x] `npx vitest run`

Issue #44, opened 2026-07-08
Structured model A/B testing — compare cost-efficiency across models with statistical significance


chrisdhaan:
## Summary

Let a developer or team declare a model experiment and have Preflight randomize sessions across arms, track outcomes, and report cost-efficiency deltas with statistical significance.

## Motivation

"I wonder if Haiku is good enough for this kind of work" is a question every team using AI coding tools asks eventually. Today the only way to answer it is intuition or a loose manual comparison. Neither produces a number you can act on confidently.

Structured A/B testing turns that question into a real answer: here's the cost delta, here's the quality delta, here's whether the difference is statistically significant.

### Experiment declaration
Configure an experiment via config or MCP tool:
```json
{
  "experiment": {
    "name": "haiku-vs-sonnet-refactor",
    "arms": ["claude-sonnet-4-5", "claude-haiku-4-5"],
    "allocation": [50, 50],
    "durationDays": 14,
    "taskFilter": "refactor"
  }
}
```

### Randomization and tracking
- Sessions are randomly assigned to an arm at session start based on allocation weights
- All standard metrics are tracked per arm: cost, efficiency score, task completion rate, anti-pattern count, error rate

### Results reporting
- `nr_observe_get_experiment_results` MCP tool returns per-arm metrics with confidence intervals
- Dashboard panel showing arm comparison with significance indicators
- Automatic "experiment complete" signal when sufficient sessions have been collected for the declared significance threshold

## Example results

| Metric | claude-sonnet-4-5 | claude-haiku-4-5 | Δ | Significant? |
|---|---|---|---|---|
| Avg cost/session | $0.34 | $0.12 | −65% | ✅ p<0.01 |
| Efficiency score | 76 | 71 | −7% | ✅ p<0.05 |
| Task completion | 94% | 91% | −3% | ❌ p=0.18 |

## Difficulty / Impact

Difficulty: 3/5 · Customer Impact: 5/5

PR #575, opened 2026-09-03
Feat: Data-driven model recommendation from session history

### LINKED ISSUE #591: Rank models by historical outcome to inform model-selection recommendations
## Problem

Preflight tracks per-session model, cost, and efficiency score, but the existing `model_selection` recommendation (via `detectModelMigrationImpact()`) only ever compares an arbitrary first two models pulled from a `Set`, once 3+ models appear in a developer's session history. It doesn't rank all observed models, doesn't account for task outcome type, and doesn't gate on sample size.

## Proposal

Add a way to rank every model seen in a developer's persisted session history by average efficiency score, cost, and task success rate — both overall and broken down by task outcome type (bug fix, feature, refactor, etc., reusing the existing outcome classifier). Gate any recommendation on the top-ranked model's sample size so a handful of sessions never produces a confident recommendation. Surface this as a new read-only MCP tool, and use it to replace the current arbitrary-pair `model_selection` recommendation with one that compares the session's actual current model against the historically better-performing one.

## Notes

- Should reuse the existing outcome classification used by cost-per-outcome breakdowns rather than introducing a new task taxonomy.
- Models should be grouped by their exact raw model string — no family/tier canonicalization exists in the codebase today, and a hardcoded mapping would go stale on every model release.

adjohn:
Fixes #591

## Summary

- Adds `TrendAnalyzer.rankModelsByOutcome()`, which ranks every model seen across a developer's persisted session history by average efficiency score, cost, and task success rate — both overall and broken down by task outcome type (`bug_fix`, `feature`, `refactor`, `investigation`, `configuration`, `documentation`, `failed_attempt`), reusing the existing `classifySessionOutcome()` classifier rather than introducing a new task taxonomy.
- Confidence is gated on the top-ranked model's sample size (`high` ≥20 sessions, `medium` ≥8, `low` ≥3, else `insufficient_data`) so a recommendation never fires off a handful of sessions. These thresholds are hand-picked judgment calls, documented as such in the code, matching this file's existing convention for heuristic thresholds.
- Replaces `RecommendationEngine.getModelRecommendations()`'s old `detectModelMigrationImpact()`-based check, which only ever compared an arbitrary first two models out of a `Set` once 3+ models had been used in a history. It now surfaces a `model_selection` recommendation only when the current session's dominant model (via a new optional `ModelUsageTracker` dependency) differs from the historically better-performing one, with both models' stats in the evidence string.
- Adds `nr_observe_get_model_recommendation`, a new read-only MCP tool exposing the full ranked report (overall + per-outcome breakdown), documented in `COMMANDS_TABLE.md`.
- Models are grouped by their exact raw model string (no family/tier canonicalization) — no such mapping exists anywhere in this codebase today, and a hardcoded one would go stale on every model release. This matches the existing `detectModelMigrationImpact()` precedent, which is left in place (still used, still tested independently).

**Explicitly out of scope for this change**, called out here for visibility rather than silently dropped:
- Folding this into the digest — `digest-formatter.ts`/`digest-sender.ts` don't consume `RecommendationEngine` output at all today; wiring it in would roughly double this change's blast radius into an unrelated subsystem.
- Cross-team/org-wide aggregation — this server has no NRDB-query capability today; this ships the local per-developer history that every other cross-session tool (`get_trends`, `get_recommendations`, weekly summaries) already relies on via `SessionStore`.

## Test plan

- [x] `npx tsc -b .` — clean
- [x] `npm test` — 183/184 suites, 6328/6332 tests passing (baseline was 6317/6321; +11 new tests, 0 regressions, same 1 pre-existing skipped suite)
- [x] New unit tests for `rankModelsByOutcome` (insufficient data, clear winner, null-model exclusion, per-outcome separation, cost tie-break, developer filter)
- [x] Updated `RecommendationEngine` tests for the new `model_selection` logic (no tracker provided, insufficient data, current-vs-recommended mismatch)
- [x] New tool-handler tests for `nr_observe_get_model_recommendation` (report shape, invalid `since`, developer filter, tool registration)

https://claude.ai/code/session_01V6GXiMXs7dwwN6YC6zRLsT


chrisdhaan (commented):
## Triage summary

**Sense-check: makes sense, in scope.** This is a solid replacement for the existing `detectModelMigrationImpact()`-based `model_selection` recommendation, which only ever compared an arbitrary first-two models pulled from a `Set`. `rankModelsByOutcome()`'s confidence gating, null-efficiency handling, and "no comparable runner-up means no recommendation" guard are all correctly conservative. Not a duplicate of #357 (UI panel) or #44 (that's active randomized A/B testing with experiment declaration — a different mechanism). Tracked as #591.

**Build/test/lint:**
- `npx tsc -b .` — clean
- `npm test` — 184/185 suites, 6424/6428 tests passing (1 pre-existing skipped suite, matches baseline)

**Findings:**

- **issue** — Version bump collides with what's already on `main`. This branch is 2 commits stale — `main`'s `package.json`/`CHANGELOG.md` are already at `1.41.0` (via #548, merged after this branch was forked, for an unrelated dashboard feature). This PR's bump to the same `1.41.0` across `package.json`/`package-lock.json`/`server.json`/`plugin/.claude-plugin/plugin.json`, plus its own `## [1.41.0]` CHANGELOG heading, will collide. Needs `main` merged in and a re-bump to `1.42.0`.
- **issue** — `kiro-power/plugin.json` wasn't touched by this PR and is stale at `1.39.0`, two versions behind. This is a recurring gap in this repo (missed in #548, #567, #568 previously).
- **nit** — `byOutcome` in `rankModelsByOutcome()` orders outcome groups by first-encountered session order rather than a fixed canonical order (bug_fix, feature, refactor, …). Cosmetic only.

Everything else — feature completeness, test coverage, MCP tool registration, `RecommendationEngine`/`ModelUsageTracker` wiring — looks correct. Nice work.

chrisdhaan (approved):

PR #581, opened 2026-09-04
fix(audit): attribute subagent calls and flag non-recursive rm

### LINKED ISSUE #577: Destructive-command detection misses non-recursive rm -f <file> deletions
## Problem

`DEFAULT_DESTRUCTIVE_COMMAND_PATTERNS` in `src/security/audit-trail.ts` only flags `rm` when both a recursive flag and a force flag are present together (`-rf`/`-fr`/`--recursive` combined with `-f`, etc). A plain `rm -f <single file>` (force only, no recursion) does not match any pattern in the list, so it is never classified as a `destructive_command` alert, even when the target is an untracked file with no way to recover it afterward.

Observed directly: an agent's `rm -f <file>` against an untracked file produced no security alert anywhere in the audit trail. (This was compounded by a separate gap where the tool call was never observed at all — see the companion issue on subagent auditing — but the pattern itself would not have caught this even if the event had reached the classifier.)

## Proposal

This is a real tradeoff, not a clear-cut bug. `rm -f <file>` alone is an extremely common, usually-benign cleanup command (temp files, build artifacts), and naively flagging every occurrence as critical would likely create heavy alert fatigue. Options worth weighing rather than one obvious fix:

- Classify plain `rm -f` at a lower severity than `-rf` (e.g. notable/info rather than critical), so it's visible/queryable without paging anyone.
- Condition severity on whether the target path is tracked in the current working tree's version control — deleting an untracked file is unrecoverable that way, which is a materially different risk than deleting a tracked one.
- Leave the pattern as-is and document explicitly that single-file `rm -f` is intentionally out of scope for `destructive_command`, so nobody relies on it for that case.

## Notes

- Pattern in question: `src/security/audit-trail.ts`, `DEFAULT_DESTRUCTIVE_COMMAND_PATTERNS`, the `rm` entries (currently require both an `r` and an `f` flag together, or `--recursive`).

### LINKED ISSUE #578: Subagent (Task/Workflow) tool calls bypass AuditTrailManager entirely
## Problem

`SubagentWatcher` (`src/hooks/subagent-watcher.ts`) is the only mechanism that observes activity inside Task-tool or Workflow-tool spawned subagents. By its own doc comment, its purpose is narrowly scoped to closing a cost-accounting gap: it polls subagent transcripts and emits `mode: 'subagent_token'` records so subagent token usage reaches `CostTracker`. It never extracts or forwards the subagent's actual tool-call records (Bash, Edit, Write, etc).

As a result, nothing a subagent does ever reaches `AuditTrailManager`/`HookEventProcessor`'s classification pipeline. Destructive-command detection, sensitive-file-access detection, and the audit log itself only ever see the parent session's own tool calls — subagent tool calls are a blind spot regardless of what they do.

Observed directly: a Workflow-spawned subagent, given an explicitly read-only research task, ran an unrequested `rm -f <file>` in the working directory, permanently deleting an untracked file. Grepping the real audit log (`~/.newrelic-preflight/audit/*.jsonl`) for that subagent's agent id returns zero matches for the entire run, while the parent session's own tool calls from the same time window are all present and correctly classified. The event was never captured at all, independent of whether the destructive-command pattern would have matched it (see #577 for that separate gap).

## Proposal

Extend `SubagentWatcher` (or a sibling mechanism reading the same transcripts) to also parse tool-call/tool-result entries, not just assistant-usage lines, and feed them through the same classification path `HookEventProcessor` uses for hook-sourced events — so destructive commands and sensitive-file access inside a subagent are audited the same way they are for the main session.

## Notes

- This affects both Task-tool subagents (`subagents/agent-{id}.jsonl`) and Workflow-tool subagents (`subagents/workflows/wf_{runId}/agent-{id}.jsonl`) — `SubagentWatcher` already watches both paths for token accounting, so the transcript access pattern to extend from already exists.
- Independent of severity: even a subagent that stays within a reasonable-looking task can run something worth auditing, and today there is no record either way to review after the fact.

adjohn:
Fixes #577. Fixes #578.

## Why

The incident behind both issues: on 2026-09-03, in a different session, a workflow subagent ran `rm -f decisions.tsv` and a search of the audit trail by that subagent's agent id found nothing. Two separate gaps produced that result, and the fix for each is smaller than the issue proposed.

For #578, the premise that subagent tool calls never reach `AuditTrailManager` is wrong. Claude Code fires `PreToolUse` and `PostToolUse` hooks for subagent calls with `agent_id` and `agent_type` in the payload, the collector stamps them on the hook event, `HookEventProcessor` copies them onto the `ToolCallRecord`, and `src/index.ts` records every one. The 2026-09-03 audit file on the reporting machine holds all 35 Bash calls from that workflow subagent, including the `rm -f`, under the parent session id. What it lacks is the agent: `AuditRecord` had no `agentId` or `agentType`, so every egress dropped them and a search by agent id found nothing. Parsing `tool_use` blocks in `SubagentWatcher`, the change #578 asked for, would write a second copy of every call hooks already deliver. This PR carries the agent identity the record already has through every audit output instead.

For #577, the destructive patterns in both the audit trail and the bash classifier require a recursive flag on `rm`, so a plain `rm -f <file>` never produced an alert. A new `file_deletion` alert at the existing `medium` severity records it without paging anyone, which is the issue's first option.

## Scope

- `src/security/audit-trail.ts`. `AuditAttribution` names the `sessionId`, `agentId`, `agentType` set and `AuditRecord` extends it. One exported `attachAuditAttribution` writes `session_id`, `agent_id`, `agent_type` onto an event or log attribute bag from a `Record<keyof AuditAttribution, string>` wire map, so a fourth attribution field fails to compile until every egress carries it. `recordToolCall` and `recordProxyCall` collapse onto one private `commit`. `detectSecurityAlert` evaluates an ordered `AlertRule` table and returns the highest-severity match, ties to the earlier row. The `file_deletion` row matches `RM_OR_UNLINK_IN_COMMAND_POSITION` or a classifier leading token of `rm` or `unlink`. New option `deletionPatterns`, default `DEFAULT_DELETION_COMMAND_PATTERNS`, `[]` to disable.
- `src/transport/log-ingest.ts`. The NR log entry gets `agent_id` and `agent_type` through the same helper.
- `src/dashboard/routes/api-handler.ts`, `src/web/api/client.ts`, `src/web/views/Audit.tsx`. The audit DTO carries both fields, the Audit page gains an Agent column and a File deletion filter chip.
- `SECURITY.md`, `docs/METRICS_TABLE.md`. Field rows for `agent_id` and `agent_type` on `AiAuditEvent` and `SecurityAlert`, and the `file_deletion` trigger.
- `docs/SCORECARDS.md`. The recommended security Scorecard query now filters `severity IN ('critical', 'high')`. Unfiltered, a single `curl` already turned a team red, and after this change a single `rm` would too.
- `src/security/audit-trail.ts` and `src/hooks/bash-classifier.ts`, in a separate commit. `git clean` with any force form (`-f`, `-fd`, `-xdf`, `--force`) and `find … -delete` join both destructive lists at `critical`. Both delete files with no recovery path and were flagged by nothing before. Bare `git clean`, `-n`, and `--dry-run` stay unflagged, because git refuses to delete without `-f` unless `clean.requireForce` is off. After the `file_deletion` rule alone, `git clean` would still produce nothing, since its leading token is `git`.
- Version 1.40.0 with a changelog section.

Not changed: `src/hooks/subagent-watcher.ts`, `src/index.ts`, `src/transport/nr-ingest.ts`, `src/storage/*`. The bash classifier's `rm` rule is untouched. The three existing alert types keep their names, severities, and description prefixes. `event_version` stays 1 because the new attributes are optional.

## Tradeoffs

- `file_deletion` shares `medium` with `external_network` rather than adding a `low` severity. A fourth level touches the type guard, both event builders, log ingest, the DTO, the page's tone map, and two doc tables, for a label `medium` already expresses. `alert_type` is the dimension that separates them.
- The deletion regex is anchored to command position, so `git rm`, `npm rm`, `docker rm`, and `echo "rm foo"` do not match, and neither do `xargs rm` or `find -delete`. The classifier's leading token covers `sudo -u root rm` and `FOO=bar rm`. `echo "; rm x"` is a known false positive at medium.
- The bash classifier's `isDestructive` is untouched. Widening it would promote `rm -f` to `critical` through the OR, the outcome #577 rejects, and it ships on `AiToolCall` where other consumers read it.
- `SubagentWatcher` stays token-only. A transcript-derived tool-call path would double count and would use the transcript's agent id space, which `ToolCallRecord`'s own doc says differs from the hook payload's.

## Blast radius

Every `AiAuditEvent`, `SecurityAlert`, and audit log entry from a subagent call gains two attributes. Every non-recursive `rm` or `unlink` in a Bash tool call now emits a `SecurityAlert` event at `medium`, which raises `SecurityAlert` volume for anyone alerting on that event type unfiltered. The on-disk audit log gains two optional keys; older lines read back unchanged.

## Verification

- `npm test`: 183 suites, 6300+ tests, green at the tip.
- Each of the four code commits builds and passes `audit-trail`, `log-ingest`, `api-handler`, and `bash-classifier` suites on its own.
- The `git clean` regex was run against twelve forms before its tests were written. Every force form matches; `git clean`, `git clean -n`, `git clean --dry-run`, and `git clean -nd` do not.
- End to end against `dist`: four hook payload pairs with `agent_id` pushed through the built collector binary, `HookEventProcessor`, `AuditTrailManager`, and `LocalStore`. The persisted line for `rm -f` carries `agentId`, `agentType`, and `file_deletion` at `medium`; `rm -rf` stays `destructive_command` at `critical`; the parent-session line has no agent keys; `npm rm` is unflagged; `getAuditLog` reads the agent back; both NR builders emit `agent_id` and `agent_type`. 14 checks, all pass.
- `npm run lint`: 0 errors, 0 warnings. `npm run format:check`: clean.

https://claude.ai/code/session_01TCQPMTtMRgr4V3po8ozDUc


chrisdhaan:
## Triage review

**Sense-check: makes sense — real, well-scoped fix.** Verified the core claim independently: `ToolCallRecord.agentId`/`agentType` already flow from the hook payload through `HookEventProcessor` into `src/index.ts`'s `record` (`src/index.ts:2102`) and into `auditTrail.recordToolCall(record)` (`src/index.ts:2109`) before this PR — the only gap was `AuditRecord` itself not carrying those two fields through to disk/NR/dashboard. That's exactly what this PR fixes, and it correctly avoids the heavier approach #578 originally proposed (re-parsing subagent transcripts, which would double-count).

**Tracked issues:** #577 and #578 both exist and match this PR's description. No duplicate PRs found.

**Build/test/lint:** all green.
- `npm run build`: clean
- `npm test`: 184/185 suites, 6458/6462 tests passed (3 skipped/1 todo, pre-existing/unrelated)
- `npm run lint`: 0 errors, 0 warnings

**Code review:** read the full diff. `attachAuditAttribution`'s wire-map (forces every egress to carry `agent_id`/`agent_type` or fail to compile) is a nice touch. Hand-verified the `git clean`/`find -delete` regex anchoring against dry-run/bare forms, and the severity tie-break logic (`rm f && curl http://x` → `external_network`, matching its own test). No bugs or completeness gaps found in the logic itself — test coverage of the new paths is thorough.

**Process findings (not code):**
- **issue** — This branch forked from `main` at v1.39.0. Since then `main` shipped v1.40.0 (#557, an unrelated Copilot fix), v1.41.0 (#548), and v1.42.0 (#575) — so this PR's own "v1.40.0" bump is a real version collision, not just staleness. `git merge-tree` confirms actual conflicts on `package.json`, `package-lock.json`, `server.json`, and `CHANGELOG.md`. Needs the branch brought up to date and re-bumped to v1.43.0.
- **issue** — `kiro-power/plugin.json` is untouched by this PR and still reads `1.39.0`. Per this repo's release workflow, a mismatch here hard-fails the release.
- **suggestion** — The drafted CHANGELOG bullets end with `(#577)`/`(#578)`. Every existing entry in `CHANGELOG.md` states the fact plainly with no issue/PR number — worth dropping to match convention.

I'll bring the branch up to date with a merge commit (not a rebase, to avoid rewriting your commits) and follow up with the version/changelog fixes myself so this is ready for a final look.

chrisdhaan (approved):

PR #587, opened 2026-09-04
fix(audit): flag git clean -f and find -delete as destructive

adjohn:
Stacked on #581. GitHub cannot target a fork branch as the base, so this PR shows #581's four commits too. Review the top two, `Fix: Flag git clean -f and find -delete as destructive commands` and `Chore: v1.41.0`. I will rebase onto main once #581 merges.

## Why

`git clean -f` deletes untracked files, recursively with `-d`, and nothing can recover them. `find <path> -delete` removes every match. Neither appears in the audit trail's destructive patterns or the bash classifier's, so both produce no alert today. After #581 `git clean` still produces nothing, because its leading token is `git` and the new `file_deletion` rule looks for `rm`. Both delete recursively by nature, so they belong in the `critical` `destructive_command` tier next to `rm -rf` and `git reset --hard`, not in the `medium` deletion tier.

## Scope

- `src/security/audit-trail.ts` and `src/hooks/bash-classifier.ts`. Two patterns added to each destructive list: `git clean` with any force form (`-f`, `-fd`, `-xdf`, `--force`), and `find … -delete`. Bare `git clean`, `git clean -n`, and `git clean --dry-run` stay unflagged, because git refuses to delete without `-f` unless `clean.requireForce` is off.
- Tests in `src/security/audit-trail.test.ts` and `src/hooks/bash-classifier.test.ts` for every matching and non-matching form above.
- `SECURITY.md` and `docs/METRICS_TABLE.md` destructive-command bullets.
- Version 1.41.0 with a changelog section.

## Blast radius

Any Bash tool call that runs `git clean -f` or `find -delete` now emits a `critical` `SecurityAlert` and sets `bashDestructive` on the `AiToolCall` event. Both are uncommon in agent sessions and both are the kind of call the audit trail exists to record. The three existing alert types and every other pattern are unchanged.

## Verification

- The `git clean` regex was run against all twelve listed forms before the tests were written. Every force form matches, and `git clean`, `git clean -n`, `git clean --dry-run`, and `git clean -nd` do not.
- `npm test` at the tip: 184 suites, 6458 tests, green. Each of the two commits builds and passes the `audit-trail` and `bash-classifier` suites on its own.
- `npm run lint` 0 errors and 0 warnings, `npm run format:check` clean, pre-commit hooks passed on both commits.

https://claude.ai/code/session_01TCQPMTtMRgr4V3po8ozDUc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.40.0] - 2026-09-04
+
+### Added
+
+- **Audit records now name the subagent that made each tool call.** Tool calls from Task and Workflow subagents already reached the audit trail through hooks, but `AuditRecord` dropped the hook payload's `agent_id` and `agent_type`, so a search of `~/.newrelic-preflight/audit/*.jsonl` by agent id found nothing. The on-disk audit log, `AiAuditEvent`, `SecurityAlert`, the NR log entry, and the dashboard Audit page now carry `agent_id` and `agent_type` (`agentId` and `agentType` on disk and in the dashboard). Both are absent for calls the parent session made. (#578)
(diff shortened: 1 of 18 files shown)


adjohn:
Folded into #581 as its own commit so reviewers see one PR instead of a fake stack. Branch deleted.

PR #624, opened 2026-09-09
Fix #607: partition sequence-based detectors by agentId

### LINKED ISSUE #607: Sequence anti-pattern detectors don't partition by agentId, causing false positives across parallel subagents
## Problem

`detectStuckLoop` (`src/metrics/anti-patterns.ts:245-271`), `detectReReading` (`:207-239`), `detectBlindEditing` (`:292-338`), and the redundant-read penalties in `src/metrics/tool-selection-scorer.ts:209-256` all iterate the session's flat `ToolCallRecord[]` in timestamp order. `ToolCallRecord.agentId` already exists (`src/storage/types.ts:309-317`) and is populated for subagent calls, but none of these detectors partition by it before analyzing.

## Impact

Three (or more) parallel subagents each independently running `npm test` register as one agent stuck in a retry loop — `detectStuckLoop` documents that non-`Bash` calls are transparent to it, so `Bash(cmd) → Read → Bash(cmd)` across different agents still counts as repetition. Three subagents each reading the same file once (a natural consequence of splitting work across them) trips `detectReReading`'s threshold of 3 over the whole session, with the resulting advice — "context may have been compressed, consider breaking the task into smaller pieces" — describing the opposite of what actually happened. `detectBlindEditing` and the tool-selection-scorer's redundant-read penalties have the same defect.

## Possible fix

Group tool calls by `agentId` (treating each subagent and the parent session as its own sequence) before running any of the sequence-based detectors. The field already exists on the record, so this is a comparatively contained fix with an outsized payoff — it removes a whole class of false positives across four detectors at once.

chrisdhaan:
## Summary

- `detectReReading`, `detectStuckLoop`, and `detectBlindEditing` (`src/metrics/anti-patterns.ts`), plus `findRedundantReads` and `findRepeatedFailures` (`src/metrics/tool-selection-scorer.ts`), each walked the whole session's flat, timestamp-ordered `ToolCallRecord[]` without separating the parent session from parallel subagents. Three subagents each independently doing something once (running the same test command, reading the same file, one failed Bash call each) looked like a single agent repeating itself.
- Added a shared `partitionByAgent()` helper (`src/metrics/agent-partition.ts`) that groups tool calls by `agentId` — one group for the parent/orchestrator session, one per subagent — preserving each group's relative order. Both files now run these detectors per group and merge the results.
- `detectThrashing`, `detectOverDelegation`, and `findUnusedOutputs` are intentionally left unchanged: thrashing is already isolated per file, over-delegation counts delegation itself (a session-wide concept), and unused-output detection isn't stateful across calls the same way the others are.
- Output shape is unchanged — no new fields, no schema/persistence changes. When multiple agents independently trip the same pattern, that now correctly produces multiple array entries instead of one falsely-merged entry (or previously, one false positive where there should have been none).
- Bumped `1.48.0` → `1.48.1` (patch) across `package.json`, `package-lock.json`, `server.json`, `kiro-power/plugin.json`, and `plugin/.claude-plugin/plugin.json`, plus a CHANGELOG entry.

Closes #607.

## Test plan

- [x] Added `src/metrics/agent-partition.test.ts` covering the helper directly (grouping, order preservation, empty input).
- [x] Added regression tests in `anti-patterns.test.ts` and `tool-selection-scorer.test.ts` reproducing each cross-agent false positive described in the issue, confirmed red before the fix and green after — plus same-agent regression tests confirming real repetition is still caught.
- [x] `npm run build` — passes
- [x] `npm test` — 6683 passing (full suite)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.48.1] - 2026-09-09
+
+### Fixed
+
+- Anti-pattern and tool-selection detectors (re-reading, stuck-loop, blind-editing, redundant-reads, repeated-failures) now analyze each subagent's tool calls separately from the parent session and from other subagents, instead of treating the whole session as one flat, timestamp-ordered sequence. Previously, parallel subagents each independently doing something once — e.g. three subagents each running the same test command, or each reading the same file — could be misread as a single agent stuck in a loop or re-reading unnecessarily.
+
 ## [1.48.0] - 2026-09-08
 
 ### Changed
--- a/kiro-power/plugin.json
+++ b/kiro-power/plugin.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json",
   "name": "newrelic-preflight",
-  "version": "1.48.0",
+  "version": "1.48.1",
   "description": "AI coding observability for Kiro. Tracks tool calls, cost, anti-patterns, and efficiency metrics — and (optionally) ships them to New Relic.",
   "author": {
     "name": "New Relic",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.48.0",
+  "version": "1.48.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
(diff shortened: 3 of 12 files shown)


chrisdhaan:
Ran an independent 5-agent parallel review after opening this. Two real (but pre-existing, out-of-scope) gaps came back, both below the auto-flag confidence threshold since neither is a regression introduced here:

1. The `detectThrashing` exclusion comment overstated its isolation — fixed in 29b60a3.
2. `src/dashboard/routes/replay-analyzer.ts` duplicates this same detection logic for the Replay UI on a type with no `agentId` field, so it still has the cross-agent false-positive bug. Filed as #625, not fixed here since #607 never scoped it in.

Also stripped "issue #607" references from code comments per this repo's comment-hygiene convention — ticket numbers belong in the commit message/PR description, not comments that outlive the ticket.

corinnekunze (approved):

PR #637, opened 2026-09-10
Fix #621: separator-agnostic alerts.rulesPath containment check

### LINKED ISSUE #621: Native Windows: alerts.rulesPath containment check hardcodes / — default path always rejected, custom paths silently discarded
## Summary

On Windows, `alerts.rulesPath` validation rejects **every** path — including the default one — so `preflight doctor` (and every server start) logs a spurious warning, and any user-configured `alerts.rulesPath` is silently discarded and replaced with the default.

```
{"rawPath":"C:\Users\<user>\.newrelic-preflight\alerts\rules.json",
 "resolved":"C:\Users\<user>\.newrelic-preflight\alerts\rules.json",
 "storagePath":"C:\Users\<user>\.newrelic-preflight",
 "fallback":"C:\Users\<user>\.newrelic-preflight\alerts\rules.json",
 "level":"warn","component":"mcp-config",
 "message":"alerts.rulesPath resolves outside storagePath — falling back to default"}
```

Note that `rawPath`, `resolved` and `fallback` are all the same path — the value being rejected as "outside storagePath" is the default path itself.

## Root Cause

`dist/config.js`, `validateRulesPath()` — the prefix used for the containment check hardcodes the Unix separator:

```js
const prefix = storageResolved.endsWith('/') ? storageResolved : storageResolved + '/';
if (resolved !== storageResolved && !resolved.startsWith(prefix)) {
    logger.warn('alerts.rulesPath resolves outside storagePath — falling back to default', {...});
    return fallback;
}
```

On Windows `path.resolve()` returns backslash-separated paths, so `resolved.startsWith(storageResolved + '/')` is never true and the branch is always taken.

This is the same class of bug as #24 (`static-handler` path containment), in a different file.

## Reproduction

```js
const { resolve, sep } = require('node:path');
const storagePath = String.raw`C:\Users\me\.newrelic-preflight`;
const resolved = resolve(storagePath, 'alerts', 'rules.json');
const storageResolved = resolve(storagePath);
const prefix = storageResolved.endsWith('/') ? storageResolved : storageResolved + '/';

resolved.startsWith(prefix);                    // false  -> warning fires
resolved.startsWith(storageResolved + sep);     // true   -> correct
```

Or simply run `preflight doctor` on Windows with a default config.

## Impact

- **Default configuration:** log noise only. The fallback resolves to the same path, so behaviour is unchanged.
- **Custom `alerts.rulesPath`:** a real bug. A value set via `alerts.rulesPath` in `config.json` or `NR_AI_ALERTS_RULES_PATH` is silently ignored on Windows and reverted to the default, even when it points inside `storagePath`.

The comment at the call site ("The default path is always permitted") does not hold on Windows.

## Suggested Fix

```js
import { sep } from 'node:path';
// ...
const prefix = storageResolved.endsWith(sep) ? storageResolved : storageResolved + sep;
```

(Accepting both `/` and `sep`, as #24 did, would also work.)

The `.json` suffix check on the line above is fine — it is separator-independent.

## Affected Versions

Present in every version I checked back to 1.15.0, through 1.44.0 (verified by extracting `dist/config.js` from the published tarballs). Not a regression from a recent release.

## Environment

- OS: Windows 11 Enterprise 10.0.22631
- Node: v22.23.1
- preflight: v1.44.0
- `platformTarget: native`, `mode: both`

adjohn:
## Summary

On Windows, `validateRulesPath()` in `src/config.ts` hardcoded `/` when building the containment-check prefix. Since `path.resolve()` returns backslash-separated paths on Windows, the check never matched — every `alerts.rulesPath`, including the default, was rejected and silently replaced with the fallback. This logged a spurious warning on every server start (`preflight doctor` included), and any custom path set via `alerts.rulesPath` in `config.json` or `NR_AI_ALERTS_RULES_PATH` was silently reverted to the default even when it lived inside `storagePath`.

This is the same bug class as #24 (`static-handler.ts` path containment), just in a different file, as the issue points out.

## Fix

Replaced the hand-rolled prefix match with `path.relative()` + `isAbsolute()` — the same separator-agnostic pattern already used in `static-handler.ts` for #24. This also drops the now-redundant `resolved !== storageResolved` special case, since `relative()` already returns `''` for equal paths.

## Verification

Reproduced the exact mechanism using `path.win32` (no Windows machine needed — Node's `win32`/`posix` path submodules simulate either platform's path semantics deterministically):

- **Before fix:** default Windows rules path (`C:\Users\me\.newrelic-preflight\alerts\rules.json` under storage root `C:\Users\me\.newrelic-preflight`) is rejected — reproduces the reported bug exactly.
- **After fix:** same default path is accepted, a custom nested path under storage root is accepted, and a sibling-prefix escape attempt (`...preflight-evil\rules.json`) is still correctly rejected — on both simulated Windows and native POSIX.

Added `src/config.windows-sep.test.ts`, mirroring the existing `static-handler.windows-sep.test.ts` precedent: it pins the `path.win32.relative`/`isAbsolute` primitive behavior the fix now relies on for the three relevant shapes (default path, nested custom path, sibling escape), since there's no separator-hardcoded logic of our own left to unit-test in isolation once the fix delegates entirely to those primitives — same rationale as the precedent file.

A true failing-then-passing Jest test isn't cheaply achievable here without either injecting the path module (over-engineering relative to the rest of the codebase, which doesn't do this either) or running on an actual Windows machine — this codebase's own precedent for the identical bug class settled the same way.

### Test plan

- [x] `npx jest -- src/config.test.ts src/config.windows-sep.test.ts` — 263 passed (existing rulesPath tests unaffected, 3 new pinning tests pass)
- [x] `npm run build && npm test` — full suite, 6847 passed, 0 failed
- [x] Pre-push hook (full test suite + plugin-hook bundle verification) — passed

Fixes #621.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.50.2] - 2026-09-10
+
+### Fixed
+
+- **On Windows, `alerts.rulesPath` validation rejected every path — including the default one — because the containment check hardcoded `/` as the path separator, and `path.resolve()` returns backslash-separated paths on Windows.** This logged a spurious warning on every server start (`preflight doctor` included) and silently discarded any custom `alerts.rulesPath` set via config file or `NR_AI_ALERTS_RULES_PATH`, reverting it to the default. The check now uses `path.relative()` + `path.isAbsolute()`, which is separator-agnostic — the same fix applied to `static-handler.ts` for the identical bug class in #24.
+
 ## [1.50.1] - 2026-09-10
 
 ### Fixed
--- a/kiro-power/plugin.json
+++ b/kiro-power/plugin.json
@@ -2,7 +2,7 @@
   "$schema": "https://agent-plugins.org/schemas/1.0.0/plugin.schema.json",
   "name": "newrelic-preflight",
   "displayName": "Preflight",
-  "version": "1.50.1",
+  "version": "1.50.2",
   "description": "AI coding observability for Kiro. Tracks tool calls, cost, anti-patterns, and efficiency metrics — and (optionally) ships them to New Relic.",
   "author": {
     "name": "New Relic",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.50.1",
+  "version": "1.50.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@newrelic/preflight",
-      "version": "1.50.1",
+      "version": "1.50.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@newrelic/preflight",
-  "version": "1.50.1",
+  "version": "1.50.2",
   "mcpName": "io.github.newrelic-experimental/preflight",
   "type": "module",
   "license": "Apache-2.0",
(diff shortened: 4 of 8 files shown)


adjohn:
Superseded by #642: identical change, re-committed under my normal git identity so the CLA check can match it (this PR's commit was authored as `Test User <test@test.local>`).

PR #644, opened 2026-09-10
Fix: recognize named subagent-transcript ids in AGENT_ID_RE

adjohn:
## Summary
- Claude Code writes a named subagent's transcript as `agent-a<name>-<16-hex>.jsonl` (the `Agent` tool's `name` param — used for teammates addressable via `SendMessage`), not just the anonymous `agent-a<16-hex>.jsonl` shape.
- `AGENT_ID_RE = /^a[a-f0-9]{16}$/` only matched the anonymous shape, so `SubagentWatcher.discoverFiles()` silently skipped every named-subagent transcript, and its tokens never reached `CostTracker` — `subagentCostUsd` stayed `0` and named-agent models never appeared in a session's `modelBreakdown`, even though `agentSpawns` correctly counted them.
- Verified against a real session on disk (`istio envoy support status`): Preflight reported `estimatedCostUsd: $3.75`, `subagentCostUsd: 0`, `modelBreakdown` limited to `claude-sonnet-5` — while Claude Code's own `/usage` panel for the same session showed `claude-haiku-4-5` spend (~$1.40) from 3 named subagents. Their actual transcript files exist on disk with real haiku-4-5 usage; none of their agentIds matched the old regex.
- Same bug, weaker form, in two more spots: `subagent-timeline-store.ts` (dashboard's independent re-parse for the subagent timeline panel + `SubagentWatcher`'s cost self-check) has the identical `AGENT_ID_RE`, and `local-store.ts`'s `SUBAGENT_CURSOR_RE` (stale cursor-file GC) hand-derives the same pattern a third time — once the watcher starts writing cursor files for named agents, its old greedy-`.+` form can't parse the new filename shape either.
- Widened `AGENT_ID_RE` to `a<16-hex>` or `a<name>-<16-hex>` in all three places, then extracted the pattern into `src/lib/agent-id.ts` so the three copies (previously byte-for-byte duplicates of one external contract — Claude Code's own naming scheme) can't drift again.

## Test plan
- [x] `subagent-watcher.test.ts`: new tests confirm a named-agent transcript (ad-hoc and under a workflow run dir) is discovered/processed, not `discovery_skipped`
- [x] `subagent-timeline-store.test.ts`: new test confirms a named-agent transcript appears in `getSubagentsForSession`
- [x] `local-store.test.ts`: new tests confirm `gcWatcherCursors()` correctly deletes a stale named-agent cursor file and preserves a live one
- [x] `src/lib/agent-id.test.ts`: unit coverage for the extracted pattern (both shapes, embedded/unanchored usage)
- [x] Confirmed each new test fails against the pre-fix regex before applying the fix (see commit history: tests land before the fix)
- [x] Existing malformed-id tests (`agent-not-a-valid-id.jsonl`) still correctly rejected — verified explicitly, not just assumed
- [x] `npm run build && npm test && npm run lint` all clean (6859 tests passed, 0 errors/warnings)

### DIFF
--- a/src/dashboard/subagent-timeline-store.test.ts
+++ b/src/dashboard/subagent-timeline-store.test.ts
@@ -97,6 +97,23 @@ describe('SubagentTimelineStore', () => {
     expect(result.window.endMs).toBe(agent.endMs);
   });
 
+  it('includes a named-subagent transcript (Agent tool `name` param spawn shape)', () => {
+    // Claude Code writes `agent-a<name>-<16-hex>.jsonl` for a subagent spawned
+    // with an explicit `name` param — distinct from the plain `a<16-hex>` shape
+    // used by anonymous Task spawns (AGENT_A/AGENT_B/WF_AGENT above).
+    const namedAgentId = 'aconfluence-istio-investigator-ca0143b626a86424';
+    writeFileSync(
+      join(subDir, `agent-${namedAgentId}.jsonl`),
+      assistantLine({ timestamp: '2026-06-16T12:00:00.000Z', input: 10, output: 5 }) + '\n',
+    );
+
+    const store = new SubagentTimelineStore({ projectsDir });
+    const result = store.getSubagentsForSession(SESSION);
+
+    expect(result.agents).toHaveLength(1);
+    expect(result.agents[0]!.agentId).toBe(namedAgentId);
+  });
+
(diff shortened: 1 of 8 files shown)


adjohn:
Superseded by #645: identical three commits, re-committed under my normal git identity so the CLA check can match them (these were authored as `Test User <test@test.local>`).

PR #645, opened 2026-09-10
Fix: recognize named subagent-transcript ids in AGENT_ID_RE

### LINKED ISSUE #651: Named subagent transcripts (Agent tool name param) are silently skipped by AGENT_ID_RE, zeroing their cost/model attribution
## Problem

Claude Code writes a named subagent's transcript as `agent-a<name>-<16-hex>.jsonl` (the `Agent` tool's `name` param — used for teammates addressable via `SendMessage`), not just the anonymous `agent-a<16-hex>.jsonl` shape.

`AGENT_ID_RE = /^a[a-f0-9]{16}$/` only matches the anonymous shape, so `SubagentWatcher.discoverFiles()` silently skips every named-subagent transcript, and its tokens never reach `CostTracker` — `subagentCostUsd` stays `0` and named-agent models never appear in a session's `modelBreakdown`, even though `agentSpawns` correctly counts them.

Verified against a real session on disk (`istio envoy support status`): Preflight reported `estimatedCostUsd: $3.75`, `subagentCostUsd: 0`, `modelBreakdown` limited to `claude-sonnet-5` — while Claude Code's own `/usage` panel for the same session showed `claude-haiku-4-5` spend (~$1.40) from 3 named subagents. Their actual transcript files exist on disk with real haiku-4-5 usage; none of their agentIds matched the old regex.

Same bug, weaker form, in two more spots:
- `subagent-timeline-store.ts` (dashboard's independent re-parse for the subagent timeline panel + `SubagentWatcher`'s cost self-check) has the identical `AGENT_ID_RE`.
- `local-store.ts`'s `SUBAGENT_CURSOR_RE` (stale cursor-file GC) hand-derives the same pattern a third time — once the watcher starts writing cursor files for named agents, its old greedy-`.+` form can't parse the new filename shape either.

## Proposed fix

Widen `AGENT_ID_RE` to accept `a<16-hex>` or `a<name>-<16-hex>` in all three places, and extract the pattern into a single shared module (`src/lib/agent-id.ts`) so the three copies — previously byte-for-byte duplicates of one external contract (Claude Code's own naming scheme) — can't drift again.

adjohn:
Fixes #651

> Replaces #644, whose three commits were authored as `Test User <test@test.local>` and so could not pass the CLA check. Same three commits, same diff (`git diff` between the two heads is empty); only the author changed.

## Summary
- Claude Code writes a named subagent's transcript as `agent-a<name>-<16-hex>.jsonl` (the `Agent` tool's `name` param — used for teammates addressable via `SendMessage`), not just the anonymous `agent-a<16-hex>.jsonl` shape.
- `AGENT_ID_RE = /^a[a-f0-9]{16}$/` only matched the anonymous shape, so `SubagentWatcher.discoverFiles()` silently skipped every named-subagent transcript, and its tokens never reached `CostTracker` — `subagentCostUsd` stayed `0` and named-agent models never appeared in a session's `modelBreakdown`, even though `agentSpawns` correctly counted them.
- Verified against a real session on disk (`istio envoy support status`): Preflight reported `estimatedCostUsd: $3.75`, `subagentCostUsd: 0`, `modelBreakdown` limited to `claude-sonnet-5` — while Claude Code's own `/usage` panel for the same session showed `claude-haiku-4-5` spend (~$1.40) from 3 named subagents. Their actual transcript files exist on disk with real haiku-4-5 usage; none of their agentIds matched the old regex.
- Same bug, weaker form, in two more spots: `subagent-timeline-store.ts` (dashboard's independent re-parse for the subagent timeline panel + `SubagentWatcher`'s cost self-check) has the identical `AGENT_ID_RE`, and `local-store.ts`'s `SUBAGENT_CURSOR_RE` (stale cursor-file GC) hand-derives the same pattern a third time — once the watcher starts writing cursor files for named agents, its old greedy-`.+` form can't parse the new filename shape either.
- Widened `AGENT_ID_RE` to `a<16-hex>` or `a<name>-<16-hex>` in all three places, then extracted the pattern into `src/lib/agent-id.ts` so the three copies (previously byte-for-byte duplicates of one external contract — Claude Code's own naming scheme) can't drift again.

## Test plan
- [x] `subagent-watcher.test.ts`: new tests confirm a named-agent transcript (ad-hoc and under a workflow run dir) is discovered/processed, not `discovery_skipped`
- [x] `subagent-timeline-store.test.ts`: new test confirms a named-agent transcript appears in `getSubagentsForSession`
- [x] `local-store.test.ts`: new tests confirm `gcWatcherCursors()` correctly deletes a stale named-agent cursor file and preserves a live one
- [x] `src/lib/agent-id.test.ts`: unit coverage for the extracted pattern (both shapes, embedded/unanchored usage)
- [x] Confirmed each new test fails against the pre-fix regex before applying the fix (see commit history: tests land before the fix)
- [x] Existing malformed-id tests (`agent-not-a-valid-id.jsonl`) still correctly rejected — verified explicitly, not just assumed
- [x] `npm run build && npm test && npm run lint` all clean (6859 tests passed, 0 errors/warnings)

https://claude.ai/code/session_01JsNz5NEzDEPVdZRQQKtsFh


chrisdhaan:
## Triage review

**Sense-check:** Passes — legitimate bugfix within scope. Named-subagent transcripts (spawned via the `Agent` tool's `name` param) were silently skipped by `AGENT_ID_RE`, zeroing their cost/model attribution. This PR replaces #644 (closed — same diff, but authored as `Test User <test@test.local>`, which failed the CLA check).

**Tracked issue:** none existed — created and attached #651.

**Build/test/lint:** all clean.
- `npm run build` — passed
- `npm test` — 6859/6859 passed, 3 skipped, 1 todo (matches the PR description's claimed count)

**Code review:**
- Verified the `AGENT_ID_PATTERN` split isn't actually ambiguous despite the name segment allowing embedded hyphens: `subagent-watcher.ts`/`subagent-timeline-store.ts` extract the candidate agentId via plain string-slicing on the `agent-`/`.jsonl` filename boundaries (not regex capture groups), then validate the whole string against the anchored regex — the fixed 16-hex tail is never in question.
- `local-store.ts`'s `SUBAGENT_CURSOR_RE` fix (pinning the parentSessionId capture group to the literal UUID shape instead of a greedy `.+`) is the correct call — it's genuinely needed once the agentId group can itself contain hyphens. Confirmed parent session IDs are always Claude Code UUIDs at this point (enforced upstream by `SubagentWatcher`'s own filter before any cursor file is written).
- No path-traversal regression in `getAgentCalls()`'s security boundary — the new pattern's character class excludes `.`/`/`, and agentId is never directly interpolated into a path, only matched against `readdirSync` results.
- nit: the name-segment character class (`[A-Za-z0-9_-]{1,64}`) is slightly looser than the `Agent` tool's own `name` param schema (`^[A-Za-z0-9][A-Za-z0-9_-]{0,63}$`, which forbids a leading hyphen/underscore). Harmless in practice since names are already sanitized upstream by Claude Code itself — not blocking.
- Test coverage, TDD ordering (tests-before-fix per commit history), and the extraction into a single shared `src/lib/agent-id.ts` module are all exactly as described. No `src/shared/` boundary violation.

**Version/CHANGELOG:** branch was forked at 1.50.2; `main` has since moved to 1.50.4 with no independent fix for this bug. I'll merge `main` in and add the version bump (→ 1.50.5) + CHANGELOG entry directly on this branch.

chrisdhaan (approved):

Issue #659, opened 2026-09-11
Run the subagent watcher in --local by default, once the local rollup persists subagent cost


adjohn:
## Summary

A `--local` dashboard process never runs the `SubagentWatcher` by default (`watcherShouldRun` in `src/index.ts` requires `NR_AI_WATCHER_MODE` to match the process mode, default `stdio`). Any session with no `--stdio` process behind it therefore never gets subagent cost. #305 describes this for Pi; it also applies to any standalone `--local` deployment and to orphaned sessions.

The gate was introduced in v1.5.0 (#213) to stop two processes tailing the same subagent transcripts. #306 (v1.14.7) made the unfiltered watcher skip any session with a live `--stdio` heartbeat, so the gate no longer prevents a race. It only decides that unowned sessions go unwatched.

Removing the gate alone does not help, because a `--local` process has nowhere to put what the watcher sees. Setting `NR_AI_WATCHER_MODE=local` today collects cost that the dashboard discards:

- `onSubagentTurn` (`src/index.ts`) feeds the process-wide `CostTracker`, `ModelUsageTracker`, and NR ingest, but never `LocalSessionAggregator`. Only the parent-token path calls `localSessionAggregator.recordTokenUsage`.
- `LocalSessionAggregator.toSummaries()` (`src/metrics/local-session-aggregator.ts`) hardcodes `subagentCostUsd: 0` and emits no `subagentCostByDayUsd`. It also skips any rollup with `toolCallCount === 0`, and `SessionStore.saveSession` refuses an empty summary over a recorded one, so a subagent-only tail for a dead session is dropped twice.
- `/api/sessions/today/aggregate` deliberately skips an unscoped process's live tracker to avoid double counting (`isUnscopedAggregatorSessionId`), so the cost does not show on Today either. It is lost on daemon restart.

## Proposed change

Run the subagent watcher in `--local` by default, with these four parts in one change:

[...]

PR #660, opened 2026-09-11
Fix #305: Drop --local watcher banner, gate opt-out banner on spend

### LINKED ISSUE #305: Pi's --local-only dashboard daemon never tracks subagent cost by default

adjohn:
Related to #305

## Why

Every default install runs the dashboard as a `--local` launchd daemon. That process never runs the subagent watcher by design, so the Today view showed the "this dashboard process isn't running its own subagent watcher" banner on every visit. The banner told users to set `NR_AI_WATCHER_MODE=local` and restart, but the plist written by `installDashboardDaemon` carries only `PATH`, so a shell export never reaches the daemon. The state is expected, already visible on Settings, and not actionable from the banner, so the banner goes.

The surviving `NR_AI_ENABLE_SUBAGENT_WATCHER=0` banner hid itself only when the day's aggregate subagent turn count was nonzero. That count sums `agent_count` over `WorkflowStore.listRuns()`, which only sees Workflow-tool script runs. On any day whose subagents were Task or Agent-tool spawns it read 0, so the banner could claim subagents were excluded directly above a KPI showing their spend. It now gates on the aggregate subagent spend, which is summed from every persisted session's `subagentCostUsd`.

## Scope

- `src/web/views/Today.tsx`. Delete the `mode_mismatch` banner branch. Change the `env_var` banner guard from `subagentTurns === 0` to `subagentUsd === 0`.
- `src/web/views/Today.test.tsx`. Replace the mode_mismatch banner test with a timed negative. Replace the turn-count guard test with a spend-with-zero-turns case that matches what a real `--local` daemon returns.
- `docs/ADVANCED.md`. Update the standalone `--local` section to say where watcher state is shown now.
- Version 1.50.6 in `package.json`, `package-lock.json`, `server.json`, `kiro-power/plugin.json`, `plugin/.claude-plugin/plugin.json`, and a CHANGELOG entry.

Out of scope: running the watcher in `--local` by default, which is the real fix for #305 and needs the local rollup to persist subagent cost first. Tracked in #659.

## Blast Radius

Two render conditions in one view. The health endpoint, its `watcherDisabledReason` type, and the Settings watcher card are unchanged. A dashboard served by an older build is unaffected. The only visible change is that `--local` dashboards lose a banner and `NR_AI_ENABLE_SUBAGENT_WATCHER=0` users with subagent spend on disk lose a false one.

## Verification

- Reproduced on the live daemon on this machine before the change. `GET /api/observability-health` returned `watcherActive: false, watcherDisabledReason: "mode_mismatch"`, and `GET /api/sessions/today/aggregate` returned `subagentUsd: 5.355, subagentTurnCount: 0`. That pair renders the banner and the contradiction on the unpatched build.
- Both new tests fail on the first commit and pass on the second. `npx vitest run src/web/views/Today.test.tsx`: 76 of 76 pass after the fix.
- `npm run build` clean. `npm run lint` clean with 0 warnings. `npx prettier --check` clean on the touched files. `npx jest src/version.test.ts` passes on the bump.
- Not done: the running launchd daemon was not restarted onto this build.

https://claude.ai/code/session_01XPRNQwBL9JicnWtbHEFS1T

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.50.6] - 2026-09-11
+
+### Fixed
+
+- **The Today view no longer shows the "this dashboard process isn't running its own subagent watcher" banner on `--local` dashboards.** Every default install runs the dashboard as a `--local` daemon that by design never runs that watcher, so the banner appeared on every visit, and the `NR_AI_WATCHER_MODE=local` instruction it gave cannot reach a launchd daemon (the plist carries only `PATH`). Watcher state remains visible on the Settings page.
+- **The remaining `NR_AI_ENABLE_SUBAGENT_WATCHER=0` banner now hides when today's aggregate shows any subagent spend, instead of when it shows any subagent turns.** The turn count only counts Workflow-tool script runs, so it read 0 on any day whose subagents were ordinary Task/Agent-tool spawns, and the banner could claim subagents were excluded directly above a KPI showing their spend.
+
 ## [1.50.5] - 2026-09-11
 
 ### Fixed
--- a/docs/ADVANCED.md
+++ b/docs/ADVANCED.md
@@ -250,7 +250,7 @@ If `NEW_RELIC_LICENSE_KEY`, `NEW_RELIC_ACCOUNT_ID`, or `NEW_RELIC_API_KEY` are s
 
 ## Running `--local` Standalone (No `--stdio` Session)
 
-The subagent/workflow transcript watchers only auto-start under `--stdio` by default (`NR_AI_WATCHER_MODE=stdio`) — a `--local` dashboard process doesn't run its own copy, since a `--stdio` session normally already covers the same data, scoped to itself, and the Today view's spend figures already aggregate every session's _persisted_ totals regardless of which process is currently serving the dashboard. If `watcherActive` is `false` for this reason, the dashboard shows a banner explaining it (distinct from the `NR_AI_ENABLE_SUBAGENT_WATCHER=0` banner, which is an explicit opt-out rather than this mode default).
+The subagent/workflow transcript watchers only auto-start under `--stdio` by default (`NR_AI_WATCHER_MODE=stdio`) — a `--local` dashboard process doesn't run its own copy, since a `--stdio` session normally already covers the same data, scoped to itself, and the Today view's spend figures already aggregate every session's _persisted_ totals regardless of which process is currently serving the dashboard. When `watcherActive` is `false` for this reason, the Settings page's watcher health shows it as inactive and the Today view shows no banner; the Today view only warns when `NR_AI_ENABLE_SUBAGENT_WATCHER=0` is set explicitly and there is no subagent spend recorded for the day.
 
 | Setting                         | What it does                                                                                             | Default  |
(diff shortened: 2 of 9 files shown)


chrisdhaan:
## Review: PR #660

**Verdict: makes sense, worth taking.** Removes a permanently-shown false-positive banner on every default `--local` install, and fixes a banner gate that read the wrong metric (subagent turn count instead of spend), which could contradict the KPI right above it. Not a duplicate of #662 — that PR explicitly stacks on this one.

**Tracked issue:** #305 exists and is relevant. Note the title says "Fix #305" but the PR body itself says the real fix is out of scope (tracked in #659/#662) — merging this correctly will *not* auto-close #305 (`closingIssuesReferences` is empty), which is fine since it isn't actually fixed yet.

**Build/test/lint — all pass, verified locally:**
- `npm run build` — clean
- `npm test` (Jest) — 203/204 suites, 6865/6869 tests passed (1 pre-existing skipped suite, unrelated)
- `npx vitest run src/web/views/Today.test.tsx` — 76/76 passed
- `npx jest src/version.test.ts` — 4/4 passed
- `npx prettier --check` on touched files — clean

Every claim in the PR's own Verification section checked out.

**Code review:**
- Logic change is correct: `subagentUsd === 0` replacing `subagentTurns === 0` in the surviving banner's guard, with the `watcherDisabledReason !== 'mode_mismatch'` exclusion preserved. Confirmed `mode_mismatch` is still fully intact elsewhere (`src/index.ts`, `src/web/api/client.ts`, `src/dashboard/routes/api-handler.ts`, `Settings.tsx`'s watcher card) — the "Settings unaffected" claim in the PR body is accurate.
- No `src/shared/` edits.
- **issue — version collision.** `origin/main` is one commit ahead (#655, merged, bumping all five version files + CHANGELOG to `1.50.6` for an unrelated change). This PR also targets `1.50.6` in all five files and its own CHANGELOG entry, so it will conflict as-is. Needs to rebase/merge current `main` and re-target `1.50.7`.

I'll merge `main` into this branch (merge commit, not rebase) and push the version fix directly.

chrisdhaan (approved):

PR #627, opened 2026-09-09
Redesign Git Efficiency tab for multi-repo, multi-worktree

adjohn:
Fixes #633

## Summary

The Git Efficiency tab assumed exactly one repo, one branch, and a hardcoded "today" window, which broke down for anyone running multiple repos or multiple `git worktree` checkouts of the same repo concurrently. This redesigns the tab around real per-workspace identity and an explicit time window, computed by one pure aggregator instead of the old incremental global tracker (which reset state at local midnight and could disagree with itself across a day boundary).

- New `WorktreeIdentityResolver` (`src/metrics/git-workspace-identity.ts`): resolves `{repoKey, worktreeKey, repoName, worktreeRoot, worktreeLabel, branch}` per directory via a single `git rev-parse` call, correctly handling bare primary checkouts, unborn HEAD, linked worktrees, and detached HEAD — none of which `--show-toplevel`-based approaches survive.
- New `GitActivityRecorder`/`ActivityStore`/`git-workspace-report.ts`: a pure, query-time aggregator (`buildGitWorkspaceReport`) that computes every KPI, score, and coaching check per workspace and per window, with no cached/incremental state to drift.
- `GitWorkspaceReporter` composition class wires the recorder + aggregator into the API route, MCP tool, and startup replay of historical sessions (including a backfill path for pre-migration sessions that predate `cwd` capture, bucketed under a named "worktree unknown" group per repo rather than one anonymous bucket).
- Frontend: `GitEfficiency.tsx` replaces the single-repo flat view with a repo → worktree tree (30-day default), 7-day hero KPIs with week-over-week deltas, human-friendly duration formatting (ms/s/m/h/d), a per-row/per-repo "Sessions" column that deep-links to the Sessions view filtered to those exact session IDs (or straight to the session detail page when there's only one), and explanatory captions for previously-unclear metrics (Conflicts, Behind, commit bursts, worktree ops, Time to PR).
- Fixed a real correctness bug along the way: "longest gap between commits" only considered gaps between existing commits, never the open-ended gap from the last commit to now — so a stale repo could show a misleadingly small "longest gap."
- Fixed a real CSS bug (not specific to this tab, but hit here first): a card's one-shot entrance animation left `transform: translateY(0)` as its permanent end state via `animation-fill-mode: both`, which silently breaks `position: sticky` on any descendant table header, in this codebase and any other using the same `AnimatedCard`.
[...]

PR #661, opened 2026-09-11
Fix: Git Efficiency listed a repo twice; failed tool calls lost cwd (v1.50.7)

### LINKED ISSUE #669: Git Efficiency tree lists a repo twice; failed tool calls lose cwd attribution
## Problem

The Git Efficiency tree groups rows by \`identity.repoKey\`. A session replayed without a \`cwd\` gets a placeholder identity from \`replaySessionToActivityRecords\`, and that placeholder carries its own synthetic string as \`repoKey\`. The view then sees two distinct repo keys that share the same display name, so the same repo (e.g. \`adjohn/preflight\`) shows up twice — once for its real worktrees, and again as a second header holding a single "worktree unknown" row.

Separately, tool calls that time out waiting for their post-hook, or whose post-hook arrives with no matching pre-hook, can never be attributed to a worktree — the timeout and orphaned-post record builders in \`HookEventProcessor\` don't copy \`cwd\` from the hook event the way the paired and denied builders already do. So even current sessions keep feeding "worktree unknown" / "unattributed" fallback rows.

## Proposed fix

- Rewrite a placeholder identity's \`repoKey\` to match a resolved worktree sharing its \`repoName\`, so the row files under the real repo instead of as a look-alike second one. Its \`worktreeKey\` stays the placeholder key so the row is still individually selectable.
- Exclude placeholder rows from the parallel-isolation best-practice check — a row with no resolvable working directory isn't a worktree that could have isolated anything.
- Copy \`cwd\` onto the timeout and orphaned-post record builders in \`event-processor.ts\`, matching the paired/denied builders.

Out of scope: session files written before v1.50.0 have no \`cwd\` on any timeline entry and will stay under "worktree unknown" — only a backfill can change that. The timeout/denied record builders also drop \`agentId\`, which is the same class of bug but belongs in its own follow-up.

adjohn:
Fixes #669

## Why

The Git Efficiency tree showed adjohn/preflight and adjohn/pstack twice each. The second copy held a single "worktree unknown" row. Separately, tool calls that timed out or arrived as an orphan post could never be attributed to a worktree, so even current sessions kept feeding those fallback rows.

The tree groups rows by `identity.repoKey`. A session replayed without a `cwd` gets a placeholder identity from `replaySessionToActivityRecords`, and that placeholder carried its own synthetic string as `repoKey`. The view then saw two distinct repo keys with the same display name. The reporter's doc comment said the row should be "grouped with the real ones by display name", but nothing implemented that.

## Scope

- `GitWorkspaceReporter.report()` now calls `adoptResolvedRepoKeys`, which rewrites a placeholder's `repoKey` to that of a resolved worktree sharing its `repoName`. The `worktreeKey` stays the placeholder key, so the row is still selectable on its own. A placeholder with no resolved sibling is left as it was.
- `git-workspace-identity.ts` now exports `UNATTRIBUTED_WORKSPACE_KEY`, `UNRESOLVED_REPO_KEY_PREFIX`, and `isPlaceholderIdentity`. The two string literals that lived privately in `git-workspace-report.ts`, `git-workspace-reporter.ts`, and `git-activity-recorder.ts` now come from there.
- `buildParallelIsolationCheck` skips placeholder rows via that predicate. A row that stands in for unattributed activity is not a worktree that could have isolated anything, so it neither counts toward the active total nor contributes file overlaps.
- `HookEventProcessor.emitUnpairedPreRecord` and the orphaned-post builder copy `cwd` from the event, matching what the paired and denied builders already did. `cwd` moves from `PreHookEvent` to `HookEventBase`, since the collector stamps it on every mode.
- Version bump to 1.50.7 with a changelog entry.

Out of scope: session files written before v1.50.0 have no `cwd` on any timeline entry and will stay under "worktree unknown". Only a backfill can change that. The four record builders in `event-processor.ts` still copy pre-event fields by hand, and the timeout and denied builders also drop `agentId`. That is the same class of bug and deserves its own PR.

## Tradeoffs

Two resolved clones of the same remote would both match a placeholder by `repoName`. The lexically smallest `repoKey` wins, purely so the choice is stable across requests. The alternative, grouping in the view by `repoName`, would have merged the two clones themselves, which the view's own comment already rejects.

## Blast radius

Dashboard only, plus the shape of failed-tool-call records. Repo-scope rollups on the Git Efficiency page now include the placeholder row's activity, which is the intended reading of "this repo, worktree unknown". Timeout and orphan records gain a `cwd` field. Nothing else reads the placeholder constants.

## Verification

- Replayed every session file under `~/.newrelic-preflight/sessions` through the installed 1.50.4 dist and this branch's dist with the same script. Before: 5 repo groups, adjohn/preflight and adjohn/pstack each listed twice. After: 3 groups, no duplicate headers, and the "worktree unknown" row sits inside its real repo.
- Four new test cases were committed first and fail on `upstream/main`. `git-workspace-reporter.test.ts` covers adoption and the no-sibling case, `git-workspace-report.test.ts` covers the isolation check, and `event-processor.test.ts` asserts `cwd` on timeout and orphan records.
- `npx tsc -b .`, `npm run lint`, `npm run format:check`, and `npm test` all pass.

https://claude.ai/code/session_01CaidUYiGGLgUaYWNoJxkEe


chrisdhaan:
## Review

**Sense-check: makes sense, on-topic, no duplicate.** This is a real bug in the Git Efficiency redesign (#627) — a session replayed without `cwd` gets a placeholder identity whose synthetic `repoKey` doesn't match the real repo's, so the tree shows the same repo twice. Separately, timeout/orphan tool-call records never copied `cwd`. #667 is a distinct, complementary follow-up that this PR explicitly scoped out — not a duplicate.

**Tracked issue:** none existed — created and attached #669.

**Build/test/lint (on the PR branch):**
- `npm run build` — pass
- `npm test` — 203/204 suites, 6868/6872 tests pass (skips are pre-existing/unrelated)
- `npm run lint` — clean, 0 errors/warnings

**Code review:**
- `adoptResolvedRepoKeys` correctly rewrites only unresolved-repo placeholders, operates on a fresh per-request `Map` (never mutates the reporter's live registry), and the lexically-smallest-`repoKey` tiebreak is intentional and documented.
- `isPlaceholderIdentity` is applied consistently everywhere a placeholder needs excluding (parallel-isolation check) or including (rollup metrics, row display).
- `cwd` moved from `PreHookEvent` to `HookEventBase` and copied in both new spots (orphaned-post, `emitUnpairedPreRecord`'s timeout/rejected paths); the pre-existing "denied" cwd test covers a separate code path (explicit `PermissionDenied` event) that already worked — no regression there.
- Test coverage is solid: new tests cover both the adoption and no-sibling-placeholder cases, the isolation-check exclusion, and `cwd` on both the orphan and timeout builders.
- No edits to `src/shared/`. No lint suppressions, no `any`.
- No bugs found in the fix itself.

**Version bump — needs fixing:** the branch is 2 commits behind `main`, and `main`'s `package.json` is already at `1.50.7` (merged via #660, and already published to npm). This PR independently bumped the same 5 files to `1.50.7` from its own stale base — merging as-is would either hit a real `CHANGELOG.md` conflict or (worse, in the 4 files where both sides say the same string) silently ship this fix under an already-published version number with no bump at all.

I'm merging `main` into this branch and pushing a follow-up commit that resolves the `CHANGELOG.md` conflict and re-bumps the five version files to `1.50.8` so this ships cleanly. No functional changes beyond that.

chrisdhaan (approved):

PR #662, opened 2026-09-11
Fix #659: Run the subagent watcher in --local by default

### LINKED ISSUE #659: Run the subagent watcher in --local by default, once the local rollup persists subagent cost

adjohn:
Fixes #659

## Why

A `--local` dashboard never ran the subagent watcher unless `NR_AI_WATCHER_MODE=local` was set, a variable the launchd plist cannot carry. The gate dates from v1.5.0 and guarded against two processes tailing the same transcripts. #306 made unfiltered discovery skip any session with a live `--stdio` heartbeat, so the gate stopped preventing a race and only decided that orphan sessions and standalone deployments (#305) went unwatched.

Setting the variable did not help either. `onSubagentTurn` never fed the per-session rollup, the rollup hardcoded `subagentCostUsd: 0`, and the Today aggregate skips an unscoped process's live tracker. The cost was observed, then dropped. Both baselines are in the verification section.

This PR removes the gate and closes every hole between the watcher and disk, so a `--local` daemon tracks subagent cost with no configuration. Stacked on #660, which removed the banner that pointed users at the variable.

## Scope

Four commits, each green on its own:

1. `hasAttributableActivity` in `src/storage/session-store.ts` replaces three divergent empty-session checks: the rollup's skip, `saveSession`'s refusal, and the inline guard in the today aggregate. The refusal gets stronger, never weaker. `mergeSummaries` gains a per-key-max merge for `costByDayUsd` and `subagentCostByDayUsd` that preserves absence.
2. `LocalSessionAggregator.recordTokenUsage` takes `agentId`, the same discriminator `CostTracker` uses, and buckets by `localDateKey(turn.timestampMs)`. `toSummaries` emits real `subagentCostUsd` and the day maps. A `persistedCostBaseline` callback seeds a rollup from disk on its first subagent turn so cost survives daemon restart under the max-based merge. `onSubagentTurn` feeds the aggregator.
3. `SubagentWatcherOptions` becomes a two-arm union so `costSelfCheck` cannot be passed unfiltered. Unfiltered discovery prunes a session when `subagents/` and `subagents/workflows/` are both older than the cutoff, polls every 10 s instead of 2 s, and stops emitting a `discovery_skipped` event per stale file.
4. `NR_AI_WATCHER_MODE` and `watcherShouldRun` are deleted. `watcherDisabledReason` narrows to `'env_var' | null`. `WorkflowWatcher` stays engine-only because it has no heartbeat exclusion. Docs, CHANGELOG, version 1.50.7.

Out of scope: the provisional `pending-*` window during which a new `--stdio` session has no heartbeat and an unfiltered watcher owns its subagents. Pre-existing, shared with `ParentTranscriptWatcher`, bounded to attribution overlap by the shared cursor.

## Tradeoffs

- Max-merge across a daemon restart is closed by seeding, at the cost of one `loadSession` read per session on its first subagent turn.
- The `subagents/` directory-mtime prune understates freshness because directory mtime moves on entry creation, not append. A session must spawn no new subagent for 24 h while still streaming into one file to be wrongly pruned, and the next spawn heals it.
- The `--local` process's own `CostTracker` now holds every session's subagent cost under its synthetic id. Inert on every persisted and aggregate path; visible only to MCP cost tools queried against a `--local` process.

## Blast Radius

`--stdio` construction is byte-identical: scoped watcher, self-check on, 2 s poll. Every new branch is guarded by `parentSessionFilter === null`. The health endpoint loses one union member; every in-repo reader is updated. Session files gain day maps a `--local` process writes; the merge is max-based so a `--stdio`-written file is never lowered.

## Verification

Two baselines were captured on the unpatched code with a live-surface script that starts a real `--local` process against a temp HOME containing a fixture orphan session and a second session with a live heartbeat:

- Unpatched, default env: 7 of 9 checks fail. Health reports `mode_mismatch`, no session file is written, the Today aggregate shows zero subagent spend.
- Unpatched, `NR_AI_WATCHER_MODE=local`: the watcher runs and reports healthy, yet no session file is written and the aggregate still shows zero. This is the "collected then discarded" bug.

After the change, the same script passes every check:

```
PASS  watcher active in --local
PASS  no disabled reason
PASS  orphan session persisted (2026-09-11_11111111-2222-3333-4444-555555555555.json)
PASS  subagentCostUsd > 0 (0.035)
PASS  subagentCostByDayUsd[today] > 0 (0.035)
PASS  estimatedCostUsd >= subagentCostUsd (0.035)
PASS  session with live --stdio heartbeat is skipped
PASS  today aggregate shows subagent spend
PASS  no cost_self_check emitted unfiltered
```

- Full Jest suite: 203 suites passed, 6888 tests passed, 1 skipped suite pre-existing.
- `npx vitest run src/web/views/Today.test.tsx`: 75 passed.
- `npm run build` clean. `npm run lint` 0 errors, 0 warnings. `npx jest src/version.test.ts` passes on the 1.50.8 bump.
- New tests cover: the predicate truth table and the refusal cases; per-key day-map merge with undefined preserved; subagent usage landing in the subagent buckets keyed by the turn timestamp; seeding applied once, additively, on the first subagent usage and not on rollup creation; unfiltered default poll interval; a stale `subagents/` directory pruned without statting its files; descent still happening when only `subagents/workflows/` is fresh; no `discovery_skipped` emitted unfiltered.
- Not done: the running launchd daemon on this machine was not restarted onto this build.

https://claude.ai/code/session_01XPRNQwBL9JicnWtbHEFS1T


chrisdhaan:
## Contributor PR review

**Sense-check: passes.** Squarely in scope — implements the tracked fix for #305/#659 (subagent cost never tracked in standalone `--local` deployments), cleanly stacked on the already-merged #660 banner fix. No duplicates in flight.

**Tracked issue:** linked #659 via the `Fixes #659` line just added to this PR's description (it was previously only referenced in the title, which GitHub doesn't parse for auto-linking).

**Build/test/lint (on this branch as-is):**
- `npm run build` — clean.
- `npm run lint` — 0 errors, 0 warnings.
- `npm test` — 202/203 suites, 6887/6892 tests passed. The one failure (`src/proxy/upstream-http.test.ts`, a DNS-rebinding timeout test) is in a file this PR never touches — looks like a pre-existing flake, not a regression.

**Feature completeness:** All four parts of #659's proposed scope are implemented and independently unit-tested: the mode gate removed, orphan-session subagent cost persisted through `LocalSessionAggregator` (with restart-survival seeding), the cost self-check scoped to `--stdio` via a discriminated-union type, and unfiltered discovery bounded (slower poll, directory-mtime pruning, suppressed per-file health spam).

**Bugs:** none found. Careful edge-case handling throughout — `applyPersistedBaseline` marks itself applied before the null check so a missing baseline doesn't retry every turn, `mergeDayMap` preserves the `undefined`-vs-`{}` distinction the aggregate route depends on, and the `SubagentWatcherOptions` union makes `costSelfCheck` without `parentSessionId` a compile error.

**Test coverage:** thorough, no gaps identified.

**Two mechanical issues, both being fixed directly on this branch:**
1. This branch was forked before #660 merged and is 3 commits behind `main`, causing real conflicts in `CHANGELOG.md`, `docs/ADVANCED.md`, `Today.tsx`, and `Today.test.tsx` (main's #660 only removed the misleading banner *text*; this PR's removal of the underlying `mode_mismatch` gate is the actual fix and supersedes it) — resolving via a merge commit from `main`.
2. The version bump in this PR targets `1.50.8`, which is already tagged/released on `main` (via an unrelated fix). Bumping to `1.50.9` instead across all five version files + the CHANGELOG heading.

Both are being applied as follow-up commits on this branch now.

chrisdhaan (approved):

PR #663, opened 2026-09-11
Fix: Count real commits and PRs on the Git tab's weekly panel (v1.50.9)

### LINKED ISSUE #670: Git tab's weekly panel miscounts commits and PRs created
## Summary

The Git tab's 7-day panel shows commit and PR-created counts that diverge from ground truth in opposite directions — fewer commits than actually happened, more PR creates than actually happened. For one real week: the panel showed 23 commits and 35 PR creates, while `git log` counted 43 commits and GitHub's API listed 24 PRs for the same window.

## Root causes

1. **Command classification loses commits in compound commands.** `classifyGitCommand` returns the first matching verb for a whole Bash command string, and push/pull/fetch/branch/worktree/stash/rebase/merge are all ranked above commit. A command like `git commit -m x && git push` gets classified only as a push, so the commit never gets recorded. In a sample week, 10 of 41 actual commit commands vanished this way.

2. **PR-create detection is a loose substring match with no success check.** `processGhCommand` matches `gh pr create` anywhere in a command segment and never checks `record.success`. This means failed retries, test-fixture strings that happen to contain the phrase, and even `gh pr comment` bodies that quote `gh pr create` all get counted as a PR creation.

3. **No `git log` hydration for the weekly/30-day report.** The per-session Git tab already hydrates from `git log` to catch commits made outside observed tool calls, but the weekly report only sees hook-observed commits — the equivalent hydration was never wired in for that view, despite an existing code comment noting it should be.

## Impact

Anyone reading the Git tab's weekly/30-day numbers, the `nr_observe_get_git_efficiency` MCP tool output, or the Repos/Worktrees tree gets commit counts that undercount and PR-create counts that overcount, sometimes badly (in the sampled week, PR creates were reported nearly 46% too high and commits nearly 47% too low).

adjohn:
Fixes #670

## Why

The Git tab's 7-day panel showed 23 commits and 35 PRs created for Sep 5 to 11. GitHub has 24 PRs for that week and `git log` has 43 commits. Both counters were wrong in opposite directions, so the panel read as "more PRs than commits", which no real week produces.

Three causes, each confirmed against the week's transcripts:

- `GitActivityRecorder` emitted one git event per Bash call and `classifyGitCommand` returned the first matching verb, with push, pull, fetch, branch, worktree, stash, rebase and merge all ranked above commit. `git commit -m x && git push` was recorded as a push only. 10 of 41 commit commands in the week vanished this way.
- `processGhCommand` matched `gh pr create` anywhere in a segment and never read `record.success`. Eight failed retries counted, ten test fixtures whose string payload contained the phrase counted, and a `gh pr comment` body that quoted it counted as a create.
- The weekly report only saw hook-observed commits. The `git log` hydration that the per-session tab already had was never wired in, as the code's own comment said.

## Scope

`src/metrics/git-event-classifier.ts`. `splitShellSegments` splits on `||`, `&&`, `;`, `|` and newline. `classifyGitSegments` classifies every git segment of a command on its own, applies `record.error` only to the last git segment since that is the one that failed under `&&`, and resolves the target directory from the whole command so a leading `cd dir &&` still attributes each segment. A segment is a git command only when it starts with `git` past env assignments or a path prefix, which drops `printf 'git commit'` text. `processGhCommand` moves here too, as one anchored regex plus a verb table. `GitEvent.hash` is set only on hydrated commits.

`src/metrics/git-activity-recorder.ts` and `src/metrics/git-efficiency-tracker.ts`. Both call `classifyGitSegments` and `processGhCommand` instead of their own copies. A `gh pr create` is skipped when `record.success` is false. The recorder's `recordId` suffix becomes `git-<index>`. `GIT_LOG_SESSION_ID` marks hydrated commits.

`src/metrics/local-session-aggregator.ts`. `collectCommitsAcrossRepos` attributes each commit to a `root`, orders primary checkouts first, and adds a `--branches` pass so commits on a branch whose worktree was removed still count. It reads the author date rather than the committer date, since a rebase restamps every commit with one committer time.

`src/metrics/git-workspace-reporter.ts`. `hydrateGitLog(commits)` ingests git-log commits keyed to the root's worktree identity, idempotent on the store's `gitlog:<hash>` record id. `report()` runs `reconcileHydratedCommits` before building.

`src/metrics/git-workspace-report.ts`. `reconcileHydratedCommits` pairs each successful non-amend hook commit with the nearest hydrated commit in the same repo within 60 seconds, one-to-one, keeping the hook record for worktree and session attribution and enriching it with hash, subject and URL. Where a repo has hydrated commits, git log is authoritative and an unpaired hook commit is dropped. A hook commit whose worktree has since been deleted resolves to no repo, so it pairs by time against any repo's leftover hydrated commit and the hydrated copy is kept.

`scripts/replay-git-counts.ts`, run as `npm run replay:git-counts <since> <until>`, replays the local Claude Code transcripts through this pipeline and prints what the weekly panel would show. `--unmatched` lists every counted hook commit git log could not vouch for. This is the verification below, kept in the repo so the next change to these counters can rerun it. `isCountedCommit` excludes failed and amend commits from `commitCount`, `commitTimestamps` and `commitsSinceLastSync`. Hydrated records do not count toward a workspace's session ids.

`src/index.ts`. One 30-day `git log` collection per interval feeds the reporter; the per-session tracker still receives only today's commits.

Version bumped to 1.50.9 with a changelog entry.

## Tradeoffs

The 60-second window is wide because a record's timestamp is the PreToolUse time and `npm test && git commit` puts the test run between that and the commit's `%ct`. One-to-one nearest matching is what keeps two rapid commits distinct despite the width.

Treating git log as authoritative drops hook commits that were later rebased or squashed away. Their successors are in git log, so the count is right, but a commit made and then rewritten inside one session shows once, not twice.

`totalGitCommands` now counts segments rather than Bash calls, so `git add && git commit && git push` is three commands. That is the more honest number.

The per-session `GitEfficiencyTracker` now classifies per segment too, but keeps its own commit counting rule, so a failed or amended commit still counts there. It hydrates from git log for today with its own 5-second dedup, and changing that is a separate fix.

Error text is attributed to the last git segment. A chain like `git push && gh pr create` that fails on the create would still hand the error to the push. The hook payload carries no per-segment exit status, so a segment rule cannot close that gap.

## Blast radius

Anyone reading the Git tab's weekly and 30-day numbers, the `nr_observe_get_git_efficiency` MCP tool, or the Repos and Worktrees tree. Existing session files replay through the same recorder, so historical windows change too. Nothing in `src/shared/` is touched. The per-session tab's numbers move only for failed `gh pr create` calls.

## Verification

Replayed every Bash tool call from the local Claude Code transcripts for Sep 5 to 11 through each build's recorder and report, with success taken from the transcript's `is_error`. The unfixed installed 1.50.4 build produced 32 commits and 40 PR creates. This branch produces 74 commits and 24 PR creates. GitHub lists 24 PRs by this author for the window. The 74 commits are the 66 that `git log` finds across the observed repos plus 8 made in temporary clones that no longer exist, and `--unmatched` lists exactly those 8. A diagnostic listing every unpaired hook commit showed only failed commits, amends, quoted-text fixtures, and commits later rewritten in a stack, which is what motivated the git-log-authoritative rule.

`npm run build` and `npm run lint` pass with 0 warnings. The seven affected suites pass, 541 tests. `npm test` passes the full suite, 203 suites and 6902 tests.

https://claude.ai/code/session_015ZpwbsfGaAa5TWnCAyLSzj


chrisdhaan:
## Review

**Sense-check: passes.** Squarely in-scope (bug fix in `src/metrics/`, no `src/shared/` edits), no duplicate PR/issue found.

**Tracked issue:** none existed — created and attached #670.

**Existing reviews:** none — no prior reviews or comments on this PR.

**Build/test/lint** (run fresh on the checked-out branch):
- `npm run build` — clean.
- `npm test` — 203/204 suites, 6902/6906 tests pass (1 suite/3 tests skipped, 1 todo — pre-existing, unrelated). Matches the numbers claimed in the PR description exactly.
- `npm run lint` — 0 errors, 0 warnings.

**Findings:**

1. **issue — version collision, needs rebase/update.** This branch is a few commits behind `origin/main`, and `main` has since shipped `v1.50.9` itself (a different, unrelated fix). This PR independently bumped to the same `1.50.9`, so merging as-is would produce two conflicting `1.50.9` entries. Needs a re-bump to `1.50.10` across `package.json`, `package-lock.json`, `server.json`, `plugin/.claude-plugin/plugin.json`, and `kiro-power/plugin.json`, plus the CHANGELOG entry. The underlying bug is confirmed *not* already fixed on `main` independently, so the fix itself is still needed — just the version number.
2. **suggestion (no action needed, pre-existing) — `git log --since` traversal-stopping.** `collectCommitsAcrossRepos` still filters with `--since=...`, which git resolves via a traversal-stopping heuristic based on committer date (verified empirically) rather than a true history-wide filter — a known git gotcha with non-monotonic history. This predates the PR and isn't worsened by it (a rebase pushes committer date *forward*, making `--since` inclusion more likely, not less), so it's not a regression — flagging only for awareness.

No other bugs, missing test coverage, or convention violations found. The classifier/segment-splitting logic, the commit reconciliation pairing, the `isCountedCommit` gating, and the two-pass `collectCommitsAcrossRepos` are all covered by well-targeted new tests (amend handling, failed-create exclusion, cross-repo pairing, deleted-worktree fallback, etc.), and the implementation matches what's described in the PR body.

Great work here — this is a genuinely tricky class of bug (compound shell commands, loose substring matching, missing hydration) and the verification via `replay-git-counts.ts` against real transcripts is a nice touch.

chrisdhaan (approved):

PR #667, opened 2026-09-11
Fix #664: Timeout and denied records keep agentId; one attribution helper (v1.50.10)

### LINKED ISSUE #664: HookEventProcessor: timeout and denied records drop agentId; consolidate the four record builders
## What

`HookEventProcessor` in `src/hooks/event-processor.ts` builds `ToolCallRecord`s in four places, each with its own hand-maintained list of fields copied from the pre and post events:

- the matched pre/post pair (copies inputSize, outputSize, inputHash, cwd, transcriptPath, permissionMode, agentId, agentType, platform)
- the permission-denied path (drops agentId and agentType)
- the timeout / rejected path in `emitUnpairedPreRecord` (drops transcriptPath, permissionMode, agentId, agentType)
- the orphaned-post path (its own subset)

#661 fixed `cwd` on the timeout and orphan builders after it surfaced as a dashboard bug. The same drift already affects `agentId`. A subagent's tool call that times out or is denied is attributed to the main agent today, because those two builders never copy `agentId`.

## Proposed fix

One `preEventFields(pre)` helper returning the shared spread, used by the matched, denied, and timeout builders, plus a matching `postEventFields(post)` for the post-only fields. New fields on `PreHookEvent` then reach every record shape at once instead of one to three of them.

## Evidence

Found during #661. Compare the field lists at the four builders; the timeout builder near `emitUnpairedPreRecord` and the denied builder both lack `agentId`, while the matched-pair builder has it.

adjohn:
Stacked on #661 (`fix-worktree-attribution`). Merge that first, or this PR's diff against `main` includes its commits.

## Why

Fixes #664. A subagent's tool call that timed out or was denied showed up as the main agent's. `HookEventProcessor` had four record builders, each with its own hand-maintained list of fields copied from the hook events, and the lists drifted. The timeout and denied builders never copied `agentId` or `agentType`, and the timeout builder also dropped `transcriptPath` and `permissionMode`. #661 patched `cwd` on two of them; this fixes the mechanism so the next field cannot drift.

## Scope

- `src/hooks/event-processor.ts` gains two module-level helpers. `attributionFields(pre, post?)` returns `inputSizeBytes`, `inputHash`, `transcriptPath`, `permissionMode`, `cwd`, `agentId`, `agentType`, and `platform`, taking pre first and post as fallback. `outcomeFields(post)` returns `success`, `errorType`, `error`, and `outputSizeBytes`. The matched-pair, denied, timeout/rejected, and orphan-post builders all spread them.
- Behavior change, intended: timeout and rejected records gain `transcriptPath`, `permissionMode`, `agentId`, `agentType`. Denied records gain `agentId`, `agentType`. The matched pair's `cwd` now falls back to the post event when the pre event lacks it. Every other field keeps its previous source and precedence.
- Version bump to 1.50.10 with a changelog entry.

Out of scope, noted for a follow-up: the event types in `src/storage/types.ts` have the same drift one layer down. The collector stamps `transcriptPath`, `permissionMode`, `agentId`, `agentType`, and `sessionId` on every mode, but `HookEventBase` declares only `platform` and `cwd`, so the post type cannot admit two of them and the helper cannot fall back to post for those. Hoisting them onto the base is a types-only change that would let the helper apply one rule to all six fields.

## Blast radius

Record shape only. Downstream consumers read these fields off `ToolCallRecord` with casts already, so records that now carry them are picked up without code changes. Nothing that was present before is removed or reordered.

## Verification

- The test commit lands first and fails on the parent branch: the denied and timeout cases assert the full attribution set with `toMatchObject` and receive `undefined` for `agentId`.
- `npx jest src/hooks/event-processor.test.ts` passes with 102 tests. `npx tsc -b .`, `npm run lint`, and `npm run format:check` are clean. `npm test` passes in full.
- A four-angle simplify review ran over the diff. Reuse and efficiency found nothing. Simplification and altitude both flagged that a `Partial<ToolCallRecord>` return annotation checks nothing because the record type has an index signature, so the helpers now rely on their inferred literal types instead.

https://claude.ai/code/session_01CaidUYiGGLgUaYWNoJxkEe


chrisdhaan:
## Review

**Sense-check:** Passes — a real, scoped bug fix. Subagent tool calls that timed out or were denied were attributed to the main agent instead of the subagent. No duplicate PRs found.

**Tracked issue:** Already linked (`Fixes #664`) and verified — the issue matches this PR's scope exactly.

**Build/test/lint** (run on the PR branch tip):
- `npm run build` — clean
- `npm test` — 203/204 suites, 6868/6872 tests passed (the 1 skipped suite is a known-flaky integration test, unrelated to this change)
- `npm run lint` — clean

**Existing reviews:** none yet on this PR.

**Scoping note:** `gh pr diff` initially showed a much larger diff than described (touching `git-workspace-identity.ts`, `git-workspace-report.ts`, `git-workspace-reporter.ts`, `src/storage/types.ts`, etc.). That's because this branch forked before #661 merged, and #661 shipped on `main` with different commit hashes than this branch's local copy of the same fix — so that part of the diff is stale-base noise, not new work. Isolating the actual new commits confirms the real contribution matches the PR description exactly: `CHANGELOG.md`, the 5 version files, `event-processor.ts`, and `event-processor.test.ts` only. No edits to `src/shared/`.

**Code review of the real diff:** `attributionFields()`/`outcomeFields()` correctly consolidate the four record builders — traced each call site (denied, matched-pair, orphan-post, timeout) against the prior per-builder logic, and the behavior matches what's described: denied/timeout now carry `agentId`/`agentType` (the actual fix), timeout also regains `transcriptPath`/`permissionMode`, and the matched pair's `cwd`/`agentId`/`agentType`/`platform` now fall back to the post event.

**Suggestion (test coverage):** the PR description calls out "the matched pair's `cwd` now falls back to the post event when the pre event lacks it" as an intended behavior change, but no test exercises that specific case for the matched-pair path (only the pre-existing `agentId`/`agentType` fallback test and the orphan-post `cwd` test cover fallback behavior elsewhere).

**Version collision:** `main` has since shipped `1.50.10` for an unrelated fix (#663), so this PR's bump to `1.50.10` collides. Merging `main` in and re-bumping to `1.50.11`.

I'll merge `main` into this branch (merge commit, not rebase) to resolve the conflicts, add the missing test, and re-bump the version — committing directly since maintainer edits are enabled on this PR.

chrisdhaan (approved):

Issue #654, opened 2026-09-11
Persist per-skill invocation counts in the session record so skills are queryable across sessions


adjohn:
## Summary

Per-skill usage exists only in process memory. `TurnCostAttributor.getMetrics().costBySkill` is served by `nr_observe_get_cost_per_tool` for the live session and then lost. The persisted session record (`~/.newrelic-preflight/sessions/<date>_<id>.json`, `FullSessionSummary`) has `toolBreakdown`, `modelBreakdown`, and `costByWorkflowRunId` but no per-skill dimension. So no cross-session question about a skill can be answered from stored data: which sessions ran `/code-review`, did a skill rewrite change output tokens per turn, what does a skill cost per week.

`nr_observe_get_session_history` and `nr_observe_get_trends` aggregate whole sessions and cannot filter by skill for the same reason.

## Where this happens

- `buildSessionSummary()` in `src/storage/session-store.ts` never receives the attributor; `BuildSessionSummarySources` has no `turnCostAttributor` field.
- `FullSessionSummary` has no `skillBreakdown`; `mergeSummaries()` and the on-disk parser (`parseSessionSummary`, the block that rebuilds `toolBreakdown` at ~line 988) have nothing to merge or parse.
- `cross-session-tools.ts` `nr_observe_get_session_history` maps a fixed field list with no skills.

## Proposed fix

1. `FullSessionSummary.skillBreakdown: Record<string, number>`: invocations per skill name for the session, both `Skill` tool calls and slash commands (#653). Built from `turnCostAttributor.getMetrics(sessionId).costBySkill[*].callCount`.
2. Merge with `mergeCounts()` like `toolBreakdown`; parse it the same way; default `{}` for legacy files.
3. Surface it as `skills` on each row of `nr_observe_get_session_history` so a caller can select sessions by skill without opening files.

Counts are enough for the comparison that motivates this (output tokens per assistant message in sessions with skill X versus without, using the `tokensOutput` and `assistantMessages` fields already persisted). Per-skill cost per session can follow once #571's unit-of-work schema settles the shape, rather than freezing a second one here.

## Related

- #544, #568 (per-Skill attribution in memory), #583 (rendering), #571 (versioned unit-of-work event), #579 (checking role model configs against outcome data needs this field).

Issue #571, opened 2026-09-03
Publish a versioned schema for a per-invocation "unit of agent work" event


adjohn:
## Summary

Preflight already captures most of the raw data needed to describe a single "unit of agent work" (a skill run, slash command, or subagent invocation) — but it's spread across several event types (`SubagentTokenHookEvent`, `WorkflowRunMetrics`, per-tool-call fields) with no single, documented, versioned event shape that a third party could build a query or integration against.

## Motivation

Enterprises publishing a shared library of skills/rules/commands to a large engineering population want to ask: which assets are actually used, what does each cost per run, and did a rewrite make a skill cheaper or just longer? Answering that requires attributing cost/tokens/duration to a specific skill or subagent invocation, not just a session or a raw tool-call log. If Preflight's schema for that unit were published and versioned, downstream teams could build against it directly instead of reverse-engineering session JSON or NRQL — and other harnesses have a stable target to normalize their own hook payloads into.

## What this adds

- A single documented event/record shape for "one unit of agent work": skill/command/subagent name, model, wall-clock duration, input/output tokens, cache read/creation tokens, session ID + parent-session ID, plus the existing team/project/org attribution fields.
- A version number on the schema itself (independent of the npm package version), so consumers can detect breaking changes without diffing source.
- Depends on / composes with #544 (per-skill name capture) and #553 (per-agent_type cost breakdown) for the underlying data — this issue is about publishing a stable contract on top of that data, not adding new capture.
- Candidate home: align with the OTel GenAI semantic-convention work in #47/#48 where it overlaps, with Preflight-specific extensions for the skill/command granularity OTel's spec doesn't cover.

## Non-goals

Not a new capture mechanism — the underlying fields already exist in `SubagentTokenHookEvent`, `WorkflowRunMetrics`, and per-tool-call records. The gap is publication and a stability guarantee, not raw data.

## Context

Raised during an enterprise evaluation focused on attributing per-run cost/duration to specific skills/commands across a shared skill library, rather than only per-session or per-tool-call.

PR #668, opened 2026-09-11
feat(hooks): attribute slash-command skill invocations and persist per-skill counts

### LINKED ISSUE #653: Slash-command skill invocations (UserPromptSubmit) are never attributed to costBySkill
## Summary

A skill invoked as a slash command (`/pstack:poteto-mode ...`, `/simplify`, `/code-review`) never reaches `costBySkill`. Only the `Skill` tool path is attributed (#544, #568). In Claude Code most skill invocations are typed as slash commands, so `nr_observe_get_cost_per_tool` reports an empty `costBySkill` for a session that ran a skill on every turn.

Observed on 1.50.4: a session started with `/pstack:poteto-mode <question>` shows `"costBySkill": {}` and `toolNames: ["ToolSearch"]` for its only attributed turn.

## Where this happens

- `src/hooks/collector-script.ts`, the `userpromptsubmit` branch, emits `{ mode: 'user_prompt_submit', timestamp }` and deliberately reads nothing from `data.prompt`. The leading `/name` token is the only thing needed and is an identifier, not free text, the same class as `tool_input.skill` already captured by `parseSkill()` in `src/hooks/tool-parsers.ts`.
- `UserPromptSubmitHookEvent` (`src/storage/types.ts`) and `BoundaryFrame` (`src/hooks/event-processor.ts`) carry only `timestamp`/`sessionId`, so even if the collector captured it there is no field to carry it.
- `TurnCostAttributor` (`src/metrics/turn-cost-attributor.ts`) keys buckets by `{ toolName, skillName }` from `ToolCallRecord` only. A slash invocation is not a tool call, so it has no bucket and no share of any turn's tokens.

## Proposed fix

1. Collector: when `data.prompt` starts with `/`, emit `slashCommand: '<name>'` (the token up to the first whitespace, `/` stripped) on the `user_prompt_submit` event. No other part of the prompt is read. Not gated on `recordContent`, matching `skillName` on `Skill` tool calls.
2. Type + processor: add optional `slashCommand` to `UserPromptSubmitHookEvent` and to the frame handed to `onUserPromptSubmit`.
3. Attributor: `recordSlashCommand(sessionId, name, timestamp)` sets the session's active slash skill (cleared by the next prompt that carries none). Every turn closed in `recordTokenEvent()` while one is active adds the turn's full cost and tokens to a bucket `{ toolName: 'SlashCommand', skillName: name }`. `costBySkill` folds `Skill` and `SlashCommand` buckets by `skillName`.
4. `index.ts`: wire `onUserPromptSubmit` to call the attributor before `startTaskIfNone`.

Semantics differ by channel and the row should say so: a `Skill` tool call gets an even split of its own turn; a slash invocation gets every turn from that prompt to the next. The second is what a person means by "what did running /code-review cost".

## Not in scope

`InstructionsLoaded` fires for CLAUDE.md and `.claude/rules/*.md` only, not for skill files, so it cannot see a skill enter context. A skill applied inline from a parent skill's instructions (no `Skill` call, no slash command) is unobservable by any hook.

adjohn:
## Why

Most skill invocations in Claude Code are typed as slash commands, and none of them reached `costBySkill`. Only the `Skill` tool path was attributed (#544, #568). A session started with `/pstack:poteto-mode` reported `"costBySkill": {}`. Per-skill usage also lived only in process memory, so no question about a skill could be answered across sessions. This PR closes #653 and #654.

## Scope

Commit 1, `feat(hooks)`:

- `collector-script.ts`: the UserPromptSubmit branch captures the leading `/name` token as `slashCommand`. No other prompt text is read. Not gated on `recordContent`, matching `skillName` on `Skill` tool calls.
- `UserPromptSubmitHookEvent` and `BoundaryFrame` carry the optional `slashCommand`.
- `TurnCostAttributor.recordSlashCommand()` marks the skill active for the session until the next prompt. Each turn closed while active charges its full cost and tokens to a `SlashCommand` bucket for that skill. `costBySkill` folds `Skill` and `SlashCommand` buckets by skill name. `totalAttributedCost` and the per-tool even split are unchanged.
- `index.ts` wires `onUserPromptSubmit` to the attributor.

Commit 2, `feat(storage)`:

- `FullSessionSummary.skillBreakdown`: invocations per skill, built from `costBySkill[*].callCount`, merged with `mergeCounts()`, parsed with a `{}` default for older files.
- `nr_observe_get_session_history` rows expose `skills`, the sorted skill names.
- Summary literals outside `buildSessionSummary()` (dashboard live view, homelab, backfill) set `{}`.

Out of scope: dashboard rendering of the new bucket, NR event changes, and per-skill cost per session. The last waits on #571's unit-of-work schema.

## Tradeoffs

A `Skill` tool call gets an even split of its own turn, as before. A slash invocation gets every turn from that prompt to the next. The two semantics share one row because a person asking "what did `/code-review` cost" means the second, and the `SkillCostEntry` doc comment states both. The alternative, recording the slash command as a synthetic tool call inside the even split, does not work: a prompt precedes its first tool call by more than the attributor's 2 second turn gap, so it would always sit alone in a turn that never receives a token event.

`skillBreakdown` stores counts only. Counts plus the persisted `tokensOutput` and `assistantMessages` fields are enough to compare sessions with and without a skill.

## Blast Radius

The buffer line for a prompt gains one identifier field when the prompt starts with `/`. `costByToolType` gains a `SlashCommand` entry once a slash bucket has attributed cost. Session files gain one object field; files without it parse to `{}`. Every place that builds a `FullSessionSummary` literal was updated, so the build enforces the new field. Cursor, Windsurf, and Copilot have no UserPromptSubmit hook and are unaffected.

## Verification

- `npm run lint`: clean.
- `npx jest`: 203 suites passed, 1 skipped; 6,881 tests passed, 3 skipped, 1 todo. New tests cover token extraction, frame routing, attributor charging and folding, cross-session aggregation, summary build, merge, deserialization, and the `skills` row field.
- End to end on the built artifact: a script fed real `UserPromptSubmit`, `PreToolUse`, and `PostToolUse` payloads through `dist/hooks/collector-script.js` into a temp buffer, drained the buffer through `HookEventProcessor` into `TurnCostAttributor`, then built and round-tripped the session record through `SessionStore`. The buffer line carried `slashCommand: "pstack:poteto-mode"` and none of the prompt body. `costBySkill` charged the turn's cost and tokens to the skill. `skillBreakdown` survived save and parse.

https://claude.ai/code/session_01FepVLmZduv5A2bWbNrjeEG


chrisdhaan:
## Contributor PR Review

**Sense-check verdict:** Makes sense — closes two already-filed, legitimate issues (#653, #654), extends the existing Skill-cost-attribution work (#544, #568), no duplicate PR found.

**Tracked issue:** Both #653 and #654 already exist and match the PR's description exactly. `closingIssuesReferences` currently only picks up **#653** — "closes #653 and #654" only satisfies GitHub's auto-link parser for the first number (needs a keyword before each, e.g. "Fixes #653 and Fixes #654"). #654 won't auto-close on merge as currently worded.

**Build/test/lint:** All pass as claimed — `npm run build` clean, `npm test` → 203/204 suites, 6881/6885 tests, `npm run lint` → 0 errors/warnings.

**Findings from review:**

- **[issue] `costByToolType` double-counts cost whenever a slash command is active** (`src/metrics/turn-cost-attributor.ts`). The new `SlashCommand` bucket is credited a turn's full, unsplit cost on every turn while a slash skill is active, and that bucket folds into `costByToolType` the same as any real tool bucket. Verified with a probe: one turn with a single `Read` call and an active slash command produces `costByToolType = {Read: $0.06, SlashCommand: $0.06}` — summing to double the real `totalAttributedCost` ($0.06). This contradicts the PR description's claim that "the per-tool even split are unchanged" and corrupts two user-facing surfaces: the `nr_observe_get_cost_per_tool` MCP tool and the dashboard's "Cost by Tool" panel, both of which will show roughly double the real per-tool cost for the duration of any slash-command-driven session (NR event ingestion via `AiTurnCost` is unaffected — `ClosedTurn` doesn't carry the slash bucket). No existing test asserts `costByToolType` consistency with an active slash command.
- **[nit] Branch is 6 commits / several releases behind `origin/main`** and needs a merge-forward before merge.
- **[suggestion] The slash-command regex (`/^\/([A-Za-z0-9_:.\-]+)/`) will capture false positives** for any prompt that happens to start with a `/`-led token that isn't a real command (e.g. a file path like `/Users/foo/bar.ts is broken` → captures `Users`), polluting `costBySkill`/`skillBreakdown` with bogus entries. Likely an accepted tradeoff given the design; flagging for awareness.

**Version bump + CHANGELOG:** Missing — this repo has no CI that catches it. `Feat` → minor bump.

I'll push a merge-forward + fixup commits addressing the issue above and the version bump.

chrisdhaan (approved):

Issue #687, opened 2026-09-14
Turn-cost attribution: 5 s TOKEN_MATCH_WINDOW_MS misses almost every turn on Opus with extended thinking (costBySkill stays at $0)


tswdsawada:
## Environment
- @newrelic/preflight 1.51.0, Claude Code (desktop), model claude-opus-5
- Windows 11, hooks via Git Bash

## Summary
`nr_observe_get_cost_per_tool` records Skill calls correctly
(`costBySkill[name].callCount` / `totalDurationMs` are right), but
`attributedCallCount` and `totalCost` stay at 0, and the session-wide
`attributionRate` is ~1.7 % (1 of 58 turns). The only turn that got cost
was a fast `Grep`; every turn where the model thought for more than 5 s
before its next tool call / final message was dropped.

## Cause
`src/metrics/turn-cost-attributor.ts`:

```ts
const TURN_GAP_MS = 2_000;
const TOKEN_MATCH_WINDOW_MS = 5_000;
...
const timeSinceLastTool = event.timestamp - state.pendingTurn.endTime;
if (timeSinceLastTool < 0 || timeSinceLastTool > TOKEN_MATCH_WINDOW_MS) return null;
```

The token event's timestamp is the assistant message that *follows* the
tool result, so `timeSinceLastTool` ≈ model latency for that response.
With Opus + extended thinking that is routinely 10–60 s, so the event is
silently dropped and the turn's cost is never attributed to any tool or
skill. The 5 s value was presumably tuned for fast, non-thinking models.

## Observed
```json
"costBySkill": {
  "keybindings-help": { "callCount": 2, "attributedCallCount": 0,
                        "totalCost": 0, "totalDurationMs": 35 }
},
"totalAttributedCost": 0.073577,
"attributionRate": 0.017241379310344827
```
(`nr_observe_get_cost_breakdown` for the same session shows $0.76 total,
so the tokens are ingested — they just never match a turn.)

## Suggestions
1. Match by `toolUseId` instead of wall-clock: the transcript's assistant
   message carries the `tool_use` ids of the calls it issued, so a token
   event can be tied to the turn whose `toolCalls` contain those ids with
   no timing heuristic at all.
2. If timing must stay: make the window model-aware (thinking-capable
   models → 60 s+) or, simpler, "until the next PreToolUse / user
   message" instead of a fixed number of ms.
3. At minimum expose `TOKEN_MATCH_WINDOW_MS` via config/env and surface
[...]

PR #705, opened 2026-09-15
feat(dashboard): persist spend attribution and add usage insights API

### LINKED ISSUE #711: Per-skill/subagent spend attribution isn't persisted, so History can't aggregate it and Today panels undercount
## Summary

The local dashboard can show spend by model, but not by skill, subagent type, plugin, or loop — nothing in it answers "what is driving my spend" the way Claude Code's own `/usage` screen does.

## Root cause

Per-skill cost only ever lived in the dashboard process's live `TurnCostAttributor` instance. It was never persisted to the session JSON files, so:

- History (which aggregates across sessions from disk) can't surface per-skill/subagent/plugin/loop spend at all.
- The Today panels that read the live attributor can under-report — e.g. showing a single skill at $0.00 — even when the session files on disk record dozens of skill calls, because the live attributor's in-memory state doesn't reflect everything already flushed to disk.

## Proposed fix

- Add a persisted `SessionAttribution` record per session (spend buckets by tool, skill, and subagent, plus high-context cost and API duration) built from `TurnCostAttributor` + `CostTracker` at session-store write time.
- Add a `usage-insights` computation/endpoint that aggregates these persisted buckets across sessions into share-of-spend insights (by skill, subagent type, plugin, loop) and highlights (high-context sessions, subagent-heavy sessions, long sessions).
- Merge other today-sessions' persisted buckets into `/api/cost-per-tool`'s live-attributor response, same pattern already used by `/api/model-usage`.

## Scope note

This is backend/storage only (new persisted field + new endpoint). Wiring the dashboard UI to actually display this data is separate follow-up work.

adjohn:
Fixes #711

## Why

The local dashboard could show spend by model but not by skill, subagent type, plugin, or loop, and nothing in it answered "what is driving my spend" the way Claude Code's `/usage` screen does. Per-skill cost existed only in the dashboard process's live attributor, so History could never aggregate it, and the Today panels that read that attributor showed a single skill at $0.00 while the session files on disk recorded dozens of skill calls.

This is the first of five PRs, split so each stays inside the 20-file line plus version bumps. It persists the attribution and adds the aggregate and its endpoints; the web side that reads them is the next PR.

## Scope

- `SessionAttribution` (`src/storage/types.ts`): one persisted record per session with spend buckets by `tool`, `skill`, and `subagent`, plus `highContextCostUsd` and `apiDurationMs`. `session-store.ts` builds it from `TurnCostAttributor` and `CostTracker`, merges it per key with `Math.max` like `modelBreakdown`, and parses legacy files without it to `undefined`.
- `CostTracker` gains `subagentByAgentType`, `highContextCostUsd`, and `apiDurationMs`; `subagentCostByAgentType` is removed and its one consumer in `cost-tools.ts` migrated.
- `ParentTranscriptWatcher` records every line's timestamp so an assistant turn carries `responseMs`, the gap to the line before it, as an estimate of API wait time.
- `ReplayTimelineEntry` carries `skillName` and `agentType`; `SkillCostEntry` and `ToolTypeCostEntry` carry `tokens`.
- `computeUsageInsights` (`src/metrics/usage-insights.ts`) and `GET /api/usage-insights?days=N`: share-of-spend insights (high context, subagent-heavy, long sessions, loops, plugins) and share tables by skill, subagent type, plugin, and loop. Plugins are derived from the `plugin:` prefix on skill and agent names.
- `GET /api/cost-per-tool` merges other today sessions' persisted buckets into the live attributor's response, the same pattern `/api/model-usage` uses, and recomputes the attribution rate across sessions.

## Tradeoffs

- Spend is the single denominator for every share. Tokens are not comparable across models and cache tiers, so a token share would mislead.
- `apiDurationMs` is an estimate from transcript timestamps. The true latency tracker is dormant upstream, so this is the only signal available.
- One generic `attribution` record instead of five named fields, because the session store already carried three near-identical parse and merge blocks for named maps.

## Blast Radius

Session files gain one optional field; every existing file loads unchanged. The cost-per-tool endpoint's unscoped response changes shape only by adding `tokens` to skill entries. The `subagentCostByAgentType` field on `CostMetrics` is gone; its only consumer was migrated. No change to ingestion, hooks, NR events, or any web file.

## Verification

- `npm run build`, `npm test` (jest, all suites), `npm run lint`, and `npm run format:check` were run on this exact commit: build clean, 7047 jest tests passed across 205 suites, 0 lint issues, Prettier clean. The version bump commit on top changes only the version fields in package.json, package-lock.json, server.json, and the two plugin manifests.


chrisdhaan:
## Review

Ran `/contributor-pr-review` on this PR. Full local verification on this exact commit, plus a read of the diff (attribution buckets, merge logic, transcript-timestamp `responseMs` derivation, the two API endpoints).

**Sense-check:** In scope, no duplicates, no `src/shared/` edits. Tracked issue created and attached: #711.

**Build/test/lint** (matches the PR description's own numbers):
- `npm run build` — clean
- `npm test` — 7047 passed, 205/206 suites (1 skipped)
- `npm run lint` — 0 issues
- `npm run format:check` — clean

**Code review:** No bugs found. Traced the full path (transcript-timestamp gap → `CostTracker` → `buildAttribution`/`mergeAttribution` → the two endpoints' merge logic) — late-arrival handling, legacy-file parsing, own-session exclusion, and the cost-based `attributionRate` recompute are all correct and well-tested (boundary tests at exactly `HIGH_CONTEXT_TOKENS`/`+1`, exactly 25% subagent-heavy, exactly 8h long-session). The `subagentCostByAgentType` → `subagentByAgentType` rename is clean, no leftover references. The `plugin:skill-name` prefix convention matches Claude Code's real plugin namespacing.

- `nit`: the skills/subagents/plugins/loops share tables silently cap at top-10 by cost with no indication of how many rows were dropped.

**Version bump:** all five version-carrying files are bumped consistently, but to a **patch** version (1.51.5). This repo's convention (see 1.51.0, 1.50.0, 1.49.0 in CHANGELOG.md) is that new functionality like this — persisted session attribution + a new `/api/usage-insights` endpoint — gets a **minor** bump. `CHANGELOG.md` also has no entry for this change.

I'm going to merge `main` in (merge commit, not rebase) and push a couple of follow-up commits directly to this branch to fix the version/CHANGELOG gap and the table-truncation nit, since I have write access here. Will comment again once that's done.

adjohn:
Thanks for the thorough pass. Agreed on the minor bump and the CHANGELOG gap, and the top-10 cap should say how many rows it dropped. The branch is yours until you say done; I won't push to it. Once your commits land I'll rebase the downstream chain (#709, #708, #710, plus one more behind them) onto the result and line their versions and CHANGELOG entries up with whatever you settle on here.

chrisdhaan:
Pushed two follow-up commits directly to this branch (merge into `main` wasn't needed — this branch already contained every commit on `main`, 0 behind):

1. `Fix #711: Bump version to 1.52.0 (minor) and add CHANGELOG entry` — bumps all five version-carrying files from 1.51.5 to 1.52.0 and adds the missing `CHANGELOG.md` entry, per the version-bump finding above.
2. `Fix: Report distinct-key counts alongside truncated share tables` — adds `skillsTotalCount`/`subagentsTotalCount`/`pluginsTotalCount`/`loopsTotalCount` to `UsageInsightsReport` so a consumer can tell how many rows a capped table dropped, per the nit above.

Re-ran the full suite after both: build clean, 7049 passed (205/206 suites, up 2 tests for the new count fields), 0 lint issues, Prettier clean.

No follow-up tickets to file — nothing surfaced during review that isn't already either fixed above or tracked elsewhere (#687/#689).

chrisdhaan (approved):

adjohn:
Thanks, both look right. The downstream chain now sits on 461b08f: #709 (1.53.0), #708 (1.53.1), #710 (1.54.0), and the last branch (1.55.0), each with a CHANGELOG entry in the same form as yours. The new *TotalCount fields are consumed: the share tables now read "top 10 of N" when they dropped rows, in #709's inline table and, after the component extraction, through ShareTable's totalCount prop in #708 and the shared panel in #710. Versions follow the minor-for-new-functionality rule; happy to compress them if four minors in a day is too many.

PR #708, opened 2026-09-15
feat(dashboard): shared design grammar components and formatters

### LINKED ISSUE #694: ShareTable renders at body font size outside the contribution grid
Already fixed this session, tracking for the PR that ships it.

## Summary

`ShareTable` inherited whatever font size wrapped it, so tables used outside the Usage Contribution grid (e.g. Model Performance, Outcomes) rendered at body size instead of the compact size every other ranked table uses.

## Fix

`ShareTable` now sets its own `text-xs` unconditionally rather than relying on an ancestor to set it. Every ranking reads at the same 12px regardless of where it's mounted.

### LINKED ISSUE #697: Chart/health-card polish: status vocabulary, spend-line re-animation, empty-column track cells
Already fixed this session, tracking for the PR that ships it. Three small, related chart/card fixes bundled into one PR.

## 1. One status vocabulary for every health card

Cache Health said "excellent" at 60%, Tool Selection said "good" at 90%, Quality said "stable" regardless of level, each `HealthCard` consumer invented its own tone wording. `HealthCard` now takes only a `tone` and renders one of `Healthy` / `Watch` / `Needs attention` / `No data`; each card's specific meaning stays in its own detail line.

## 2. Spend lines re-animated from zero on every refetch

Recharts redrew `SpendBars`' cumulative/projected lines from zero on every data refetch, so they were usually caught mid-draw rather than settled.

## 3. Empty-column block charts read as a floating narrow chart

A quiet day with activity in only 4 of 24 hours drew nothing for the other 20, so `DiscreteBlockChart` looked like a narrow chart floating mid-column instead of a full-width one. Empty columns now render one muted baseline cell so the chart always spans its full width.

adjohn:
## Why

The Today and History pages had grown by accretion: money rendered four ways, durations without units, scores on three scales, six health panels each with its own layout, rankings drawn as three different chart types, and raw enums on screen. Before the pages can be rebuilt they need one grammar to be built from.

Third of five PRs, after #705 and #709 (both merged). This one adds the shared components and formatters; the page rebuild that consumes them is #710.

## Scope

- `src/web/components/ui/Panel` (title, subtitle, action, footnote), `HealthCard` (hero value, status tone, detail line, up to four rows; it takes no children), `RankedBars`, `ShareTable` (sortable, own type size, optional `totalCount` that renders `top N of M` when rows were capped), and `UsageInsightsList`, each with tests.
- `DiscreteBlockChart` gains a `levels` prop that quantizes to a fixed number of block sizes, fills its column, and draws a faint track cell for empty columns. The raw-count mode remains the default for existing callers.
- `lib/format.ts`: `formatUsd` gains a sub-cent floor and thousands separators (`$1,234.56`, `$0.088`, `<$0.001`), plus `formatMs`, `formatPct`, `formatAxisDate`, `formatAxisWeek`, `formatAxisUsd`. The two existing page tests that asserted four-decimal dollars are updated here since the pages render through `formatUsd` today.
- Health card status is one vocabulary derived from tone: Healthy, Watch, Needs attention, No data.

## Tradeoffs

- `HealthCard` has a strict contract with no children so every bespoke panel body can be deleted in the next PR; a card that cannot fit is treated as a contract bug, not a reason to keep a bespoke body.
- The new components have no consumer until the next PR. Reviewing them separately keeps that PR to the page logic.

## Blast Radius

Web only. The one visible change on existing pages is `formatUsd`'s precision: costs under a dollar now show two or three decimals instead of four. Everything else is additive.

## Verification

- `npm run build`, `npm run test:web`, `npm run lint`, `npm run format:check` on this exact commit: build clean, 646 web tests passed, 0 lint issues, Prettier clean. The web typecheck reports the same five pre-existing errors upstream main reports; the next PR removes them. The version bump commit changes only the version fields in package.json, package-lock.json, server.json, and the two plugin manifests, and adds this PR's CHANGELOG entry.


chrisdhaan (approved):

chrisdhaan:
This PR fixes part of #697 (status vocabulary, empty-column track cells) — the spend-line re-animation part landed separately in #710. Noting for traceability since this merged before the reference could be attached.

chrisdhaan:
This PR fixes #694 (ShareTable renders at body font size outside the contribution grid) — `text-xs` was applied unconditionally from this component's introduction here. Noting for traceability since this merged before the reference could be attached.

PR #729, opened 2026-09-17
feat(dashboard): Sessions today tile shows which sessions need input, are ready for review, or are working

### LINKED ISSUE #693: Feature: surface per-session status (Ready for review / Needs input / Completed) in Today's KPI strip
## Summary

Explored via Investigation this session: reliably surface whether a session still needs attention ("done" means different things per session, coding sessions often mean a PR, brainstorming/Q&A sessions have no artifact) on Today's KPI strip, similar in spirit to Claude Code's own Agent View three-state grouping (Ready for review / Needs input / Completed).

**Scope correction:** originally discussed against Today's Activity panel (the block-chart grid), but the actual feedback that prompted this is about the top KPI strip specifically, the "Subagent Spend" and "Tool calls" tiles aren't considered useful as currently shown and are candidates to be replaced by a session-count + status tile.

## What Claude Code's own mechanism actually is (confirmed via docs, not guessed)

Six underlying states (`Working`, `Needs input`, `Idle`, `Completed`, `Failed`, `Stopped`), grouped in the UI into `Ready for review` (open PR exists), `Needs input` (waiting on a question/permission/decision), `Working`, and `Completed` (merges Completed+Failed+Stopped). None of this is exposed via any hook payload or API field, it's computed client-side inside Claude Code itself, confirmed against `code.claude.com/docs/en/agent-view` and the hooks reference. External tools must re-derive it.

## What Preflight can already build this from, without new instrumentation

- `AskUserQuestion` tool calls are already captured per session (`src/hooks/tool-parsers.ts`). A session whose last recorded activity is `AskUserQuestion` is a strong "needs input" signal.
- `PullRequestMetrics` / `PrEvent` (`create`/`merge`/`view`/`edit`/`ready`/`checks`, each with a `prNumber`) already exist in `src/metrics/git-efficiency-tracker.ts` and roll up per-worktree in `src/metrics/git-workspace-report.ts`, already used by `GitEfficiency.tsx`. A session with a `create` event and no matching `merge` for the same `prNumber` maps directly to "Ready for review".
- The existing `outcome: 'completed' | 'in progress'` field on persisted sessions (`src/storage/session-store.ts`) is misleadingly named for this purpose, it's just live-vs-ended, not richer status, don't build on it for this.

## Known reliability caveat

#687 (open): turn-cost attribution misses almost every turn on Opus with extended thinking. Not directly in this data path, but the same general attribution-reliability question applies, verify AskUserQuestion/PrEvent capture rate before trusting a status derived from them in a coaching-facing UI.

## Proposed fix

Replace the "Subagent Spend" and/or "Tool calls" tiles in Today's KPI strip with a session-count + status tile (e.g. "6 sessions today, 2 need input, 1 ready for review"), computed as:
- **Needs input**: session ended, last tool call was `AskUserQuestion`.
- **Ready for review**: session ended, has a `PrEvent` create with no matching merge for that `prNumber`.
- **Completed**: everything else.

## Why it matters

The user's stated pain point: "I'm always struggling to remember if I have more work to do on a session or if it's done." Distinguishing these three states directly on the dashboard removes that per-session memory burden, matching how Claude Code's own Agent View already does it, just using data Preflight already has.

adjohn:
## Why

"Done" means different things per session. A coding session is done when its PR is up, a question-and-answer session has no artifact, and a session waiting on a permission prompt is neither working nor finished. Today's KPI strip counted sessions but said nothing about which ones needed a person. Claude Code's own agent view groups sessions into ready for review, needs input, working, and completed; none of that reaches hooks or any API, so Preflight derives it from what it already records.

Fixes #693. Last of the three follow-up PRs; #727 and #728 are merged.

## Scope

- `src/metrics/session-status.ts`: `SessionStatus` and a pure `deriveSessionStatus` with one precedence table. Needs input when the session is live and its last tool call was `AskUserQuestion`. Ready for review when it opened a pull request that has no later merge for the same number. Working when it is live. Completed otherwise.
- `/api/sessions/today/aggregate` computes the status for each of today's sessions at read time, from the live registry, the last timeline entry (or the live buffer when newer), and the session's own git activity records, and returns counts and session ids per status.
- Today's "Sessions today" tile shows the non-zero statuses as its sub-label, each linking to `/sessions?sessionIds=` for those sessions.

## Tradeoffs

- Derived at read time, not persisted, so a resumed session's status follows its newest activity. The scan covers today's sessions only and sits behind the route's existing cache.
- PR events come from the per-session git activity records, not the session summary's `prActivity` list, which carries no session id. A `gh pr create` without a parseable number still counts as open.
- Heuristics can be wrong: a session that asked a question and was then abandoned reads as needs input until its liveness window lapses.
- Ready for review only clears when a `merge` record in the same session carries the same PR number as the `create`. `gh pr create` records no number (the classifier takes one only from the command line), so a CLI-created PR reads as ready for review until it leaves the day window, whichever session merges it. Capturing the number from the create's response and matching across sessions is tracked in #773.

## Blast Radius

One additive field on the today aggregate response; one tile's sub-label on Today. Nothing persisted changes.

## Verification

- `npm run build`, `npm test`, `npm run test:web`, `npm run lint`, `npm run format:check`, and the web typecheck on this exact commit: build clean, 7114 jest tests passed across 206 suites, 710 web tests passed, 0 lint issues, Prettier clean, 0 web type errors. The status tile and the token hover were also checked on an isolated preview against a seeded session with an unmerged `gh pr create` and a skill bucket carrying a breakdown.
- The version bump commit changes only the version fields in package.json, package-lock.json, server.json, and the two plugin manifests, and adds this PR's CHANGELOG entry.


chrisdhaan:
## Review

**Verdict:** In scope, tracked (closes #693), no duplicates.

**Verification** (fresh checkout of `feat/693-session-status`, branch not stale vs. `main`):
- `npm run build` — clean
- `npm test` — 7114 passed, 3 skipped, 1 todo, across 206 suites
- `npm run test:web` — 710 passed
- `npm run lint` — 0 issues
- `tsc --noEmit -p tsconfig.web.json` — 0 errors

All match the PR description's claims.

**Version bump + CHANGELOG:** Correct — all five version-carrying files (`package.json`, `package-lock.json`, `server.json`, `plugin/.claude-plugin/plugin.json`, `kiro-power/plugin.json`) bump 1.56.0 → 1.57.0 against current `main`, and the CHANGELOG entry is present and customer-facing.

**Findings:**

1. **Suggestion** — `countOpenPrs` (`api-handler.ts`) only matches a PR's `create` against a later `merge` *within the same session's own record array* (`prRecordsBySession` is grouped strictly by `record.sessionId`). If a PR is created in one session and merged from a different session (e.g. `gh pr merge` run in a follow-up session), the creating session's `ready_for_review` status never clears, since its `create` record is never checked against another session's records. Not mentioned in the PR's own Tradeoffs section. Filed as a separate issue since it's a real gap in the main use case but isn't a small fix for this PR: #773.
2. **Nit** — the per-request `WorktreeIdentityResolver`/`ActivityStore`/`GitActivityRecorder` instantiation in `computeSessionStatusAggregate` re-spawns git subprocesses on every poll rather than sharing a longer-lived cache — but this mirrors an existing pattern already in this file (the `/api/git-efficiency`-style handler further down), so it's consistent with precedent, not a new issue introduced here.

Nothing else stood out — `deriveSessionStatus`'s precedence table, the `Kpi`/`Today.tsx` sub-label rendering, and the new status counts (computed over the same `sessionsSeen` set that feeds `sessionCount`, so they can't drift from it) are all solid, and test coverage on both the pure function and the API/UI integration is thorough.

chrisdhaan:
One more nit on top of the review above: `src/dashboard/routes/api-handler.ts` has a code comment referencing the issue number —

```
// Per-session lifecycle status (#693) for today's Sessions today KPI.
```

Could you drop the `(#693)` from that comment? This repo's convention is to keep issue/PR/ticket references out of code comments — they belong in the commit message and PR description, not in code, since they rot as the codebase evolves and the number stops meaning anything to a future reader. The rest of that comment (and the two-sentence explanation after it) is fine as-is.

chrisdhaan (approved):

adjohn:
Thanks, agreed on both.

1. The cross-session gap is real, and it is wider than the issue says. `processGhCommand` only sets `prNumber` from a number that follows the verb on the command line, so a `gh pr create` records a null number every time, and a later `gh pr merge 42` cannot match it even inside the same session. In practice a created PR reads as ready for review until the session's day window ends, whichever session merges it. The fix is to capture the created PR's number from the create command's response in the recorder, then match creates against every merge in the window rather than the creating session's own records. That is the shape I put on #773; it is a collector and recorder change, so out of scope here as you said. I added the limitation to this PR's Tradeoffs so the body no longer overstates the heuristic.
2. The per-request recorder instantiation mirrors the git-efficiency handler on purpose; a shared longer-lived cache for both would be its own change.

No code changed on the branch; the head you verified is the head.

Issue #739, opened 2026-09-17
Mark every dollar figure as an estimate, with the reasons it can differ from a bill


adjohn:
## Problem

Every dollar figure in the dashboard and the MCP cost tools is a list-price estimate from token counts (`src/shared/pricing.ts`), but nothing in the UI says so. For users on subscription plans the number is not what they pay (#291), and even for pay-as-you-go accounts it drifts from the invoice for reasons the reader cannot see. The word "estimated" appears in the field name (`estimatedCostUsd`) and nowhere a user looks.

## Proposed change

A small reusable marker next to dollar figures, with a hover listing why the number can differ from a bill:

- pricing: public list price, not a negotiated rate or a flat-rate subscription
- scope: this machine and the platforms Preflight has hooks for, not the whole org
- cache accounting: cache-write tokens are estimated on adapters without exact token data
- lag and period boundaries: billing runs behind and uses its own day and time zone

Implement as one component under `src/web/components/ui/` and apply it to the Today cost KPI and the Sessions table cost column first. Other surfaces can follow in later PRs. Keep the copy in one place so the org-spend work (#737) can reuse it.

## Acceptance

- The marker renders on the Today cost KPI and the Sessions cost column.
- A vitest test covers the hover text.
- Prettier and lint clean; no new colors outside the existing tokens.

PR #750, opened 2026-09-18
feat(dashboard): copy a session as Markdown, JSON or link

### LINKED ISSUE #745: Copy a session as Markdown or JSON from the session detail dialog
## Problem

`SessionDetailDialog` (`src/web/components/SessionDetailDialog.tsx`) has one action: close. A developer who wants to paste a session into a standup note, a PR description, or a bug report has to retype the numbers.

## Proposed change

Two copy actions in the dialog header, using the data the dialog already has:

- **Copy summary as Markdown.** A short block: session name, start and duration, cost (with the estimate marker copy from #739 once it lands), tool calls, files touched, models used, top anti-patterns. Keep it under fifteen lines.
- **Copy as JSON.** The session summary object as fetched, pretty-printed.

Use `navigator.clipboard.writeText` with a brief "Copied" state on the button. No new fetches. If `Sessions.tsx` already reads a session id from the query string, also add "Copy link" that produces that URL; if it does not, leave that out of this issue.

## Acceptance

- Both buttons appear in the dialog and copy the expected text (vitest with a mocked clipboard).
- Markdown output has no secrets: run it through the same redaction as the rest of the UI if any field can carry raw content when `recordContent` is on.
- Keyboard reachable; the dialog's focus trap still works.

8rulerstar:
## Summary

Closes #745.

Adds three copy actions under the Session Detail header:

- **Copy Markdown**: a short summary block (under fifteen lines). Lines whose data is missing are left out.
- **Copy JSON**: the session summary object, pretty-printed.
- **Copy link**: `/sessions?id=<sessionId>`, which the Sessions view already reads to select a session.

Each button briefly shows "Copied" after a successful `navigator.clipboard.writeText`, or "Copy failed" if the clipboard is unavailable or permission is denied (e.g. plain HTTP), so a failed copy isn't mistaken for a successful one. The result is also announced through a visually hidden `role="status"` region; the buttons keep their accessible names and a fixed width, so the row doesn't shift when the label changes.

Example Markdown output:

```
### Session: preflight
- Started: Sep 18, 10:44 AM (3m 18s)
- Estimated cost: $0.42
- Tool calls: 12 (Read 6, Edit 3, Bash 2)
- Tool success rate: 92%
- Models: claude-sonnet-5
- Longest failure streak: 2
- Top anti-patterns: redundant_read 3, retry_loop 1
```

## Differences from the issue

I'd be glad to change any of these if you had something else in mind:

- **Where the data comes from.** The dialog only receives the decision tree, turn costs and context data, so it doesn't have the session name, start time, tool breakdown or anti-patterns. I added an optional `session` prop and pass the selected entry from the list `LiveSessionPane` already fetches. No new requests. Without it, the copy buttons are hidden.
- **Files touched** is left out, because the session summary doesn't carry that field.
- **Estimate marker:** since #739 hasn't landed yet, the Markdown says "Estimated cost" for now.
- **JSON for a brand-new live session:** until the session appears in `/api/sessions`, `LiveSessionPane` builds a minimal entry from the live registry, so that's what gets copied.
- **Button placement:** the buttons sit on their own row under the description rather than next to the close button, because at phone widths they overlapped the title.

## Redaction

The Markdown only contains the session name (derived from the working directory), numbers, model names and anti-pattern type names. No field carries raw content, so I didn't add extra redaction. Please tell me if I've missed a field.

## Test plan

- [x] New `src/web/lib/session-export.test.ts`: Markdown content, missing fields, top-three limits, cost fallback, JSON output, and link encoding
- [x] `SessionDetailDialog.test.tsx`, with a mocked clipboard: each button copies the expected text; "Copied" appears and resets; a rejected write shows "Copy failed"; keyboard copy works; the focus trap still wraps around the new buttons; buttons are hidden without a session
- [x] The new dialog tests fail without the change
- [x] `tsc`, `eslint src/web`, `prettier --check`
- [x] `npm run test:web` (vitest): all 718 pass on Node 24
- [x] `npm test` (jest) passes on Node 24, via the pre-push hook. I ran it with `TZ=America/New_York` because the DST test in `api-handler.test.ts` fails in a timezone without DST (mine is Asia/Seoul), on `main` as well.
- [x] Checked manually in the local dashboard with mocked API data at 320, 375, 768 px and desktop, in light and dark themes
- [ ] Playwright e2e: not run locally (no browser installed); the e2e suite only covers the empty Today state

This is my first contribution here. Thanks for the clear issue!


chrisdhaan:

PR #759, opened 2026-09-18
Fix #755: Make Jest suite time-zone independent

### LINKED ISSUE #755: Make the jest suite time-zone independent so CI can drop the TZ pin
## Problem

`.github/workflows/ci.yml` pins `TZ: America/New_York` with a comment explaining why: several date and bucket tests build fixtures from local-midnight helpers without pinning a zone, so the suite fails on a UTC runner and on any contributor machine outside that zone. The comment names this as a follow-up that never got filed.

## Proposed change

- Find the failing tests: run `TZ=UTC npm test` and `TZ=Asia/Tokyo npm test` locally and list every failure.
- Fix each test so it is zone-independent: build fixtures from explicit UTC timestamps, or pin the zone inside the test with the same helper the code under test uses (`src/lib/date.ts` has `todayPortionRatio` and friends). Do not change production behavior.
- Remove the `TZ` pin from `ci.yml`, or keep it and add a second matrix entry under `TZ: UTC`, so the portability is enforced rather than assumed.

## Acceptance

- `TZ=UTC npm test` and `TZ=Asia/Tokyo npm test` both pass locally.
- CI runs the suite in at least one zone that is not America/New_York.
- The explanatory comment in `ci.yml` is deleted or updated to match.

## Notes

Mechanical and well bounded. It is also a fast tour of the date handling that the cost, forecast, and alert code all depend on.

wyi184246-creator:
## Summary

- replace the DST fixture's runtime `TZ` mutation with explicit UTC instants
- preserve America/New_York DST coverage and add a dedicated UTC CI job
- keep production behavior unchanged

Closes #755

## Test plan

- [x] `npm run lint`
- [x] `npx tsc -b .`
- [x] `npm run build:web`
- [x] `TZ=America/New_York npx jest --runInBand --silent -- src/dashboard/routes/api-handler.test.ts`
- [x] `TZ=UTC npx jest --runInBand --silent -- src/dashboard/routes/api-handler.test.ts`
- [x] `TZ=Asia/Tokyo npx jest --runInBand --silent -- src/dashboard/routes/api-handler.test.ts`


chrisdhaan:

PR #760, opened 2026-09-18
Fix #755: Make the Jest suite timezone-independent

### LINKED ISSUE #755: Make the jest suite time-zone independent so CI can drop the TZ pin

dyk1454683243-sudo:
## Summary

- The concurrency history `dailyPeaks` test assumed `America/New_York` by writing `process.env.TZ` and hard-coding a 23h spring-forward day. Jest’s fake `Date` captures the process-start zone and ignores later `TZ` writes, so the fixture was wrong on a UTC runner (the GitHub Actions default).
- That test now builds fixtures from `localStartOfDay()` / `localDateKey()` — the same helpers production uses — so it holds in any host zone.
- Dropped the `TZ: America/New_York` pin from `.github/workflows/ci.yml`. CI now runs in the runner default (UTC).
- Documented timezone-independent fixture rules in `docs/TEST_PATTERNS.md`.
- Stabilized a pre-existing `workflow-watcher` partial-write test that skipped the second poll when both writes shared a 1s mtime (flaked `TZ=Asia/Tokyo npm test`).

Fixes #755

## Test plan

- [x] `TZ=UTC npm test` (7099 passed)
- [x] `TZ=Asia/Tokyo npm test` (7099 passed)
- [x] `npm run lint`
- [ ] CI verify job green without the America/New_York pin


chrisdhaan:

Issue #737, opened 2026-09-17
Epic: show org-reported AI spend beside Preflight's estimate, with the gap explained


adjohn:
## Summary

Let Preflight show an organization's reported spend next to its own estimate. Many companies expose actual AI spend through an internal API, a finance export, or the provider's admin API. Today Preflight can only show what it computes locally from tokens and list prices, and users have no way to see how that compares to the number their org treats as real.

The two numbers will differ, and that is expected. The point of the feature is not to make them match but to put them side by side and explain the gap.

## Design constraints

- **Never blend.** The local estimate and the reported figure are separate values with separate labels. Preflight does not auto-correct its estimate from the reported number.
- **Explain the mismatch in the UI.** Every place the pair appears carries the same short list of reasons: scope (org total versus this machine), pricing (list price versus negotiated or subscription), lag (billing runs behind), coverage (only hooked platforms, only this machine), cache accounting (estimated cache-write tokens on some adapters), and period boundaries and time zones. The `estimated` marker from #739 owns that copy.
- **Sources are adapters.** A `SpendSource` interface mirroring the platform adapter registry: `fetchActuals(period)` returns spend in dollars, the period, an as-of timestamp, and a scope label. Sources are registered, detected from config, and the dashboard reads through one seam.

## Sources, in delivery order

0. **Manual entry.** A person types the figure into Settings with an as-of time (#742). No automation, and it proves the side-by-side rendering and the staleness rule before any source code exists. First slice.
1. **JSON file on disk.** A path in config; any cron or script can write it. Zero network, testable offline, and enough for most companies to start.
2. **Generic HTTP.** URL, auth header from an env var, and a field mapping. Same SSRF rules as `HttpUpstream`.
3. **Provider admin APIs.** Depends on the spike below; the Anthropic usage and cost reports are raw HTTP under the organizations endpoints, need an admin key, and are not available through every purchasing path.

## Children

- Manual entry (first slice): #742
- Spike: #738
- Estimated marker (prerequisite copy): #739
- File source and dashboard pairing: to be filed after the spike
- Generic HTTP source: to be filed after the file source lands

## Related

#59 (billing API reconciliation), #291 (flat-rate plans), #63 (ingest reconciliation), #61.

PR #762, opened 2026-09-18
Feat #739: Mark dashboard dollar figures as estimates

### LINKED ISSUE #739: Mark every dollar figure as an estimate, with the reasons it can differ from a bill

dyk1454683243-sudo:
## Summary

- Add a reusable `est.` marker (`CostEstimateMarker`) next to dollar figures, with a hover that lists why the number can differ from a bill.
- Shared copy lives in one place (`COST_ESTIMATE_REASONS`) so org-spend pairing (#737) can reuse it:
  - **Pricing** — public list price, not a negotiated rate or a flat-rate subscription
  - **Scope** — this machine and the platforms Preflight has hooks for, not the whole org
  - **Cache accounting** — cache-write tokens are estimated on adapters without exact token data
  - **Lag and period boundaries** — billing runs behind and uses its own day and time zone
- Wired on the Today spend KPI and the Sessions list cost column. Other surfaces can follow later.
- No new colors; marker uses existing `ink-muted` / `ink-subtle` / `ink-base` tokens.

Fixes #739.

## Test plan

- [x] Prettier / lint clean on commit
- [x] `npm run test:web` — 49 files, 710 tests passed
- [x] `CostEstimateMarker` hover/focus lists all four reasons
- [x] Today spend KPI and Sessions cost column render the marker
- [x] Sessions row marker is non-focusable (`embedded`) so it does not nest a control inside the row button
- [ ] CI green


chrisdhaan:

PR #763, opened 2026-09-18
Feat #745: Copy a session as Markdown or JSON

### LINKED ISSUE #745: Copy a session as Markdown or JSON from the session detail dialog

dyk1454683243-sudo:
## Summary

- Add **Copy Markdown**, **Copy JSON**, and **Copy link** to the session detail dialog and the Sessions detail panel.
- Markdown is a short standup/PR block (under 15 lines): name, start/duration, estimated cost, tool calls, files touched (counts only), models, longest failure streak, top anti-patterns. Missing fields are omitted.
- JSON is the fetched session object, pretty-printed, with content-bearing fields stripped (`file`, `command`, `filesRead`, `filesModified`, `timeline`, `sessionIntent`) so a paste cannot carry raw paths, commands, or the first-user-prompt when `recordContent` is on.
- Copy link produces `/sessions?id=<sessionId>`, which Sessions already honors. No new fetches — Today passes the list entry already in memory; Sessions uses the detail payload it already loaded.
- Each button shows a brief Copied / Copy failed state and announces it through a visually hidden `role="status"` region. Keyboard reachable; the dialog focus trap still wraps.

Fixes #745.

## Notes

- #739's `est.` marker has not landed, so the Markdown label is **Estimated cost** (same word the issue asked to reuse).
- Files touched is counts only — never paths.
- Anti-pattern totals use the detector-specific fields (`readCount`, `iterations`, `repeatCount`, …), not only `count`, which persisted records usually lack.

## Test plan

- [x] `src/web/lib/session-export.test.tsx` — Markdown content, missing fields, detector-specific anti-pattern counts, files-touched counts, secret-stripping JSON, link encoding
- [x] `SessionDetailDialog.test.tsx` — mocked clipboard: each button copies the expected text; Copied appears and resets; rejected write shows Copy failed; keyboard copy; focus trap still wraps; buttons hidden without a session
- [x] `Sessions.test.tsx` — detail panel copies Markdown/JSON without leaking file paths
- [x] `npm run test:web` — 49 files, 726 passed
- [x] `eslint` / `prettier --check` on commit
- [ ] CI green


chrisdhaan:

PR #764, opened 2026-09-18
feat(dashboard): show the daily budget meter on Today

### LINKED ISSUE #740: Show the daily budget meter on Today, not only on Alerts
## Problem

Budget status is only visible on the Alerts view (`src/web/views/Alerts.tsx`, "Budget Status" section, backed by `GET /api/budget`). A developer who set a daily budget in Settings has to leave Today, the page they actually watch, to see how much of it is used.

## Proposed change

Add a daily-budget meter to the Today KPI strip, rendered only when `budget.daily.budgetUsd` is non-null. Reuse `fetchBudget` and `qk.budget` from `src/web/api/client.ts`, and the existing `Kpi` component (`src/web/components/Kpi.tsx`) so it matches the strip. Show spent over budget and percent used; switch to the warning treatment at the same thresholds `BudgetTracker` fires on (50, 80, 100 percent). When no daily budget is set, render nothing. Do not add a second fetch path; the query key is shared with Alerts.

## Acceptance

- With `dailyBudgetUsd` set, Today shows the meter and it updates on the same cadence as the other KPIs.
- With it unset, the strip is unchanged.
- A vitest test (`*.test.tsx`) covers both states and the 80 percent warning treatment.

dyk1454683243-sudo:
## Summary

- Add a daily-budget KPI to the Today strip when `budget.daily.budgetUsd` is set.
- Reuse `fetchBudget` and `qk.budget` (same 10s cadence as the other Today KPIs and Alerts) so there is no second fetch path.
- Show spent over budget and percent used; apply `Kpi` warning/bad tones at the same BudgetTracker thresholds (50 / 80 / 100).
- Leave the strip unchanged when no daily budget is configured.

Fixes #740

## Test plan

- [x] `npx vitest run src/web/views/Today.test.tsx` covers unset (strip unchanged), set (spent/limit + percent), and the 80% warning treatment (78 tests passed)
- [ ] With `dailyBudgetUsd` set in Settings, Today shows the meter and it updates with the other KPIs
- [ ] With it unset, the Today KPI strip is unchanged


chrisdhaan:

PR #767, opened 2026-09-18
Preview the weekly digest before sending it

### LINKED ISSUE #744: Preview the weekly digest before sending it
## Problem

The only digest action is `POST /api/digest/send` (`src/dashboard/routes/api-handler.ts`) and the matching `nr_observe_send_digest` MCP tool. There is no way to see what the digest will say before it goes to a channel. A user tuning their team id, budget, or week boundary has to send a real message to check the output.

## Proposed change

- `GET /api/digest/preview` that builds the same `WeeklySummary` the send path uses and returns the Block Kit payload from `formatSlackDigest` (`src/digest/digest-formatter.ts`) plus a plain-text rendering of it. No delivery, no side effects.
- A "Preview digest" button next to the existing send control in the dashboard, opening a dialog that shows the plain-text rendering with a "Send now" action that calls the existing send route.
- The same preview as `nr_observe_get_digest_preview`, or a `preview: true` flag on the send tool. Pick whichever keeps `docs/COMMANDS_TABLE.md` simplest and document it.

## Acceptance

- Preview returns the same content a send would deliver for the same moment, proven by a test that calls both against a fixture summary and compares the payload.
- Preview makes no outbound request (assert no webhook call in the test).
- Vitest covers the dialog open, content render, and the send action.

dyk1454683243-sudo:
## Summary

- Add `GET /api/digest/preview` and `nr_observe_get_digest_preview` so the weekly Slack digest can be inspected without sending it.
- Preview builds the same current-week `WeeklySummary` the send path uses and returns the Block Kit payload from `formatSlackDigest` plus a plain-text rendering derived from that payload. No webhook call.
- Alerts page: **Preview digest** next to **Send test now** opens a dialog with the plain-text preview and a **Send now** action that posts to the existing send route.

Fixes #744

## Test plan

- [x] `GET /api/digest/preview` returns the same Block Kit payload `POST /api/digest/send` would deliver for a fixture summary
- [x] Preview does not call the Slack webhook
- [x] `nr_observe_get_digest_preview` is registered when `WeeklySummaryGenerator` is available and does not require a webhook
- [x] Alerts: Preview digest opens the dialog, renders the plain-text preview, and Send now hits `/api/digest/send`
- [x] Send now stays disabled when no webhook is configured


chrisdhaan:

PR #770, opened 2026-09-18
Honor digestSchedule with a --local digest scheduler

### LINKED ISSUE #736: digestSchedule is stored and editable but nothing reads it: wire a scheduler or remove it
## Problem

`digestSchedule` is a fully plumbed config field that nothing reads:

- `src/config.ts` declares it (default `0 9 * * 1`), validates it, and loads it from the config file;
- `src/dashboard/routes/api-handler.ts` exposes it on the settings API and accepts `PATCH` updates (~line 2733);
- `src/web/api/client.ts` mirrors it in the settings type, and the Alerts view (`src/web/views/Alerts.tsx`) renders an editable schedule input that saves it;
- `docs/COMMANDS_TABLE.md` (~line 1741) says: "Digest delivery is manual-only today. `digestSchedule` is stored for future use but nothing currently reads it."

A user can set a schedule on the Alerts page, see it saved, and nothing ever happens. Verified against `main` at 9859468: the only readers outside config, the settings API, and the web mirror are that input's own bindings.

## Proposed change

Pick one and do it fully:

**A. Wire it.** In `--local` mode (the long-running daemon) start a scheduler that parses the cron expression, and on each tick runs the same path `nr_observe_send_digest` uses. Log each send. Do not schedule in `--stdio` mode; a per-session engine is the wrong lifetime. Add a small cron-expression parser or take a dependency; either way, reject invalid expressions at config load with a clear error.

**B. Remove it.** Drop the field from config, the settings API, the web type, the Alerts input, and the docs.

Option A is the product-shaped answer and what the docs promise. Option B is honest and small. Decide with a maintainer before starting.

## Acceptance

- A: with `digestSchedule` set to a one-minute-in-the-future expression, a `--local` daemon sends a digest within that minute, and the log shows it. Invalid expressions fail config load.
- B: `grep -rn digestSchedule src docs` returns nothing.

dyk1454683243-sudo:
## Summary

`digestSchedule` was stored and editable but nothing read it, so a user could save a cron on the Alerts page and nothing ever sent (#736).

This wires a real scheduler (option A) instead of deleting the field:

- 5-field cron parser (lists, ranges, steps, month/dow names; Vixie “both restricted → OR” semantics; local timezone)
- `DigestScheduler` starts only in `--local` — polls every 15s, re-reads the expression from env/file each tick, and calls the same `handleSendDigest` path as `nr_observe_send_digest`
- Invalid expressions fail config load and settings `PATCH`
- Changing only the webhook and/or schedule no longer requires a restart
- Docs, Alerts copy, and `nr_observe_subscribe_digest` now describe scheduled `--local` delivery

`--stdio` does not schedule: a per-session engine is the wrong lifetime. Manual `nr_observe_send_digest` still works everywhere.

Fixes #736

## Test plan

- [x] `npm run lint` — 0 errors / 0 warnings
- [x] `npm run format:check`
- [x] `npm run build`
- [x] `TZ=America/New_York npm test` — 7133 passing
- [x] `npm run test:web -- src/web/views/Alerts.test.tsx`


chrisdhaan:

PR #772, opened 2026-09-18
Refactor: extract computeHistoricalCosts into its own module

### LINKED ISSUE #766: Extract computeHistoricalCosts from index.ts into its own module with tests
## Problem

`computeHistoricalCosts` (`src/index.ts`, around line 3521 on `main` at 9859468) is a pure function: it takes a `SessionStore`, the current session id, and a reference timestamp, and returns prior daily and weekly spend for the budget tracker's baseline. It has no callers outside `src/index.ts`, no unit test of its own, and lives in a 3,560-line CLI entry file where nobody looks for cost math. The cross-midnight bucketing it does is exactly the kind of logic #755 is about.

## Proposed change

- Move it to `src/storage/historical-costs.ts` next to `SessionStore` (or `src/metrics/` if the maintainer prefers; ask in the PR).
- Export it, import it from `src/index.ts`, no behavior change.
- Add `historical-costs.test.ts` with fixtures for: sessions yesterday and today, a session spanning midnight, a session older than seven days, and the current session excluded. Build fixtures from explicit UTC timestamps so the test is zone-independent.

## Acceptance

- `git diff --stat` shows the function removed from `src/index.ts` and added in the new module with a test file.
- `npm test` passes, and the new test fails if the seven-day window or the current-session exclusion is broken on purpose.
- No change to any budget alert behavior (existing budget tests untouched and green).

## Notes

Small, mechanical, and it teaches the store, the budget baseline, and the date helpers in one PR.

PloneMraz:
## Summary

Closes #766.

- Moves `computeHistoricalCosts` out of `src/index.ts` (3,560 lines) into a new `src/storage/historical-costs.ts`, next to the `SessionStore` it reads from. The implementation and the explanatory comment move verbatim — no behaviour change, same single call site.
- Narrows the parameter type from `SessionStore` to a new `HistoricalCostSessionSource` interface exposing only `loadAllSessions`, so the function is unit-testable without touching the filesystem. A real `SessionStore` still satisfies it structurally.
- Removes the now-unused `todayPortionOfSessionCost` import from `src/index.ts`.
- Adds `src/storage/historical-costs.test.ts` with 7 cases: today-only session, prior-week session (weekly but not daily), a cross-midnight session pro-rated by its timeline, a session outside the seven-day window, exclusion of the current session, sessions with `estimatedCostUsd === null`, and the unreadable-history fallback returning zeroes.

One deviation from the issue text worth flagging: the issue suggests building fixtures from explicit UTC timestamps. That alone is not zone-independent here, because `todayPortionRatio` buckets with `localStartOfDay()`, so a fixture pinned to a UTC wall-clock hour lands on a different side of midnight depending on the runner's zone. The test instead derives every timestamp from `localStartOfDay(REF_TS)` — the same helper the code under test uses — which is the approach #755 recommends. There is a comment in the test saying so.

## Test plan

- [x] `npx jest -- src/storage/historical-costs.test.ts` — 7/7 pass
- [x] Mutation-checked, per the issue's acceptance criteria: replacing the seven-day window with `new Date(0)` fails `ignores sessions older than the seven-day window`; deleting the `currentSessionId` guard fails `excludes the current session from both baselines`
- [x] `npm run lint` — clean
- [x] `npx prettier --check .` — clean
- [x] `npx tsc -b .` — clean
- [x] `npm test` — see note below

Note on the full suite: I developed this on Windows, where the suite has pre-existing failures unrelated to this change (for example `src/install/json-utils.test.ts` calls `process.getuid`, which is undefined on Windows). To make sure this PR adds none of them, I ran the full suite on both sides:

| | Test suites | Tests |
|---|---|---|
| pristine `main` | 21 failed, 184 passed | 150 failed, 6949 passed |
| this branch | 21 failed, 185 passed | 150 failed, 6956 passed |

Identical failure counts; the difference is exactly the 7 new tests. CI on Linux should be green — please do check, since I could not verify a fully green local run on this platform.


PloneMraz:
Follow-up: I noticed no checks run on this branch, so I verified on Linux myself rather than leaving it to you.

Both trees run in the same container (`node:24`, Node v24.21.0, `TZ=UTC`), same `node_modules`, only the three changed files swapped between runs:

| | Test suites | Tests |
|---|---|---|
| `main` @ 9859468 | 5 failed, 200 passed | 6 failed, 7093 passed |
| this branch | 5 failed, 201 passed | 6 failed, 7100 passed |

Identical failures on both sides; the difference is exactly the 7 new tests. The 6 failures look environmental rather than related to this PR — for the record they are:

- `api-handler` › `keys dailyPeaks correctly across a DST transition` — `TZ=UTC` has no DST, so the fixture has nothing to transition across. (Possibly of interest for #755: the suite is not simply zone-independent, one test needs a zone that *has* DST.)
- `schedule` › two `plist exists but is unreadable` cases and `headless-install` › `returns error when settings directory is not writable` — the container runs as root, which bypasses the permission bits these assert on.
- `platform` › `isWsl returns false when WSL_DISTRO_NAME is absent and not on WSL kernel` — Docker Desktop runs on a WSL2 kernel, so the negative case cannot hold here.
- `index.privacy` › `boots, runs, and shuts down cleanly` — timing-sensitive; the bind mount into the container is roughly 4x slower than native disk.

One correction to the table in the PR description above: those Windows numbers (150 failures either side) are real but a poor guide, since almost all of it is Windows-only noise. The Linux comparison is the one worth reading. I also want to flag that my first Linux run reported more failures on this branch than on `main`; re-running both cleanly showed that was timeout flakiness from a cold bind mount, not a real difference. Numbers above are from the clean runs.

chrisdhaan:

PR #775, opened 2026-09-18
Refactor: one git remote-URL parser, with credential stripping

### LINKED ISSUE #716: Git remote-URL parsing is duplicated across four call sites with inconsistent credential handling
## Problem

Four separate functions parse a git remote URL into `owner/repo` or a browsable commit URL:

- `inferProjectId()` / `inferRepoUrl()` in `src/config.ts`
- a parser in `src/index.ts` (~line 1421)
- `repoNameFromRemote()` in `src/metrics/local-session-aggregator.ts`
- `commitUrlFromRemote()` in `src/metrics/local-session-aggregator.ts`

Only one code path consistently strips credentials before the value can reach a log or event field (`config.ts`'s `resolvedRepoUrl` explicitly runs `redactSensitive()` on the raw remote URL); the others extract `owner/repo` or build a commit URL directly from the remote string without that step.

## Expected

A single shared remote-URL parser (with credential stripping applied once, inside the shared function) that all four call sites use, removing the duplication and the inconsistency in credential handling.

## Why this matters

It's also a prerequisite for any future work that stamps `vcs.*`-shaped attributes (matching Claude Code's OTel export — see `docs/ADVANCED.md`'s "Repository identity across both streams" section) on Preflight's own events, since that would need one canonical parser rather than four slightly different ones.

## References

Surfaced during review of #713.

PloneMraz:
## Summary

Closes #716.

Four call sites parsed a git remote independently: `inferProjectId()` in `config.ts`, an inline regex in `index.ts` (~line 1421), and `repoNameFromRemote()` / `commitUrlFromRemote()` in `metrics/local-session-aggregator.ts`. Only `config.ts`'s `resolvedRepoUrl` ran `redactSensitive()` on the raw remote before the value could reach a log or event field.

Adds `src/lib/git-remote.ts`: one `parseGitRemote()` that strips any `user@` / `user:password@` userinfo **first**, then splits host and path. `repoNameFromRemote()` and `commitUrlFromRemote()` are thin wrappers over it, and the aggregator re-exports both names so existing importers (`git-workspace-identity.ts`) and the existing tests are untouched.

### The credential case this fixes

All three "last two path segments" regexes (`/[/:]([^/]+\/[^/]+?)(?:\.git)?$/` and the stricter `[\w.-]` variant in config.ts) capture the credential as the owner when the path has a single segment:

```
https://ghp_token@github.com/widgets.git  →  "ghp_token@github.com/widgets"
```

That value then travels wherever the caller puts it — `projectId`, `repoName` on session summaries, event fields. `git-remote.test.ts` asserts directly that no parse result contains the token or an `@`.

### Behaviour otherwise unchanged

Scheme URLs, scp-like syntax (`git@host:owner/repo.git`), nested GitLab groups (still the last two segments for a name, still the full path for a commit URL) and local paths all parse as before; `commitUrlFromRemote` still returns `null` for a remote with no host. The existing `local-session-aggregator.test.ts` cases pass unmodified.

## Test plan

Verified on Linux in a container (`node:24`, Node v24.21.0, `TZ=UTC`), same `node_modules`, only the changed files swapped between runs:

| | Test suites | Tests |
|---|---|---|
| `main` @ 9859468 | 5 failed, 200 passed | 6 failed, 7093 passed (7103) |
| this branch | 5 failed, 201 passed | 6 failed, 7119 passed (7129) |

The deltas are exactly the 26 new tests (7129 − 7103) and the one new suite; failure counts are identical. Those 6 failures are environmental and present on `main` too (a DST test under `TZ=UTC`; two `plist unreadable` and one `directory not writable` case that root bypasses; `isWsl`, since Docker Desktop is a WSL2 kernel).

- [x] `npx jest -- src/lib/git-remote.test.ts` — 26/26
- [x] `npx jest -- src/metrics/local-session-aggregator.test.ts src/config.test.ts` — unmodified, still green (333 tests with the new file)
- [x] `npm test` — no new failures vs `main`
- [x] `npx tsc -b .`, `npm run lint`, `npx prettier --check .` — clean

Happy to split the credential fix from the deduplication if you would rather review them separately.


chrisdhaan:

PR #776, opened 2026-09-18
Fix #723: don't estimate cost for a session another process owns

### LINKED ISSUE #723: Local dashboard daemon's byte-size cost fallback can inflate a session's estimatedCostUsd above the real cost
## Environment
- @newrelic/preflight v1.56.0
- `--local` dashboard daemon running continuously via launchd (`com.preflight.dashboard`) alongside per-session `--stdio` MCP engines
- macOS, Claude Code

## Summary
A session's persisted `estimatedCostUsd` (surfaced via `nr_observe_get_session_history`, the "Sessions Today" dashboard table via `AiCodingTask`, and any other consumer of the session summary file) can be **inflated far above the real cost**, while `modelBreakdown` (and everything downstream of it — `ModelUsageTracker`, `TurnCostAttributor`/`AiTurnCost`) stays accurate. Observed on session `444bcfe0-3a26-4a5e-a0bc-e2e02e1e4ec6`:

| Source | Value |
|---|---|
| `estimatedCostUsd` (persisted session summary) | **$11.33** |
| `modelBreakdown.claude-sonnet-5.totalCostUsd` (same file) | **$6.80** |
| `AiTurnCost` sum(cost_usd) for this session (NRQL) | $5.61 (151 turns, 91% attribution — healthy) |
| `AiSubagentTurn` sum(usd) for this session (NRQL) | ~$1.00 |
| `AiToolCall` count for this session (NRQL) | 166 — matches local exactly |

`modelBreakdown.totalCostUsd` + subagent cost ($6.80 + $1.00 ≈ $7.80) roughly reconciles with the real per-turn data. `estimatedCostUsd` ($11.33) does not — it's ~45% higher than what the real per-request/per-turn accounting supports.

## Cause

This machine runs a `--local` dashboard daemon (long-lived, `drainAllSessions: true` per `src/index.ts:2059`) *alongside* the `--stdio` MCP engine that actually owns a given Claude Code session's transcript connection. The daemon processes hook-buffer activity for **every** session on the machine, including ones it has no transcript connection to.

1. The daemon never receives a real token/usage event for a session it doesn't own — `costTracker.getMetrics().reportCount` stays `0` forever from its point of view.
2. The byte-size cost fallback (`costTracker.recordEstimatedTokens`, gated on `reportCount === 0` at `src/index.ts:2226-2236`) is designed to fire only for the handful of tool calls before a session's first real usage report. But for a session the daemon will *never* get real data for, that gate never flips — so the heuristic (`chars/4`, no cache-read-discount awareness) fires on **every tool call** the daemon observes for that session, for its entire duration.
3. Both the daemon and the real engine independently call `SessionStore.saveSession()` → `mergeSummaries()` (`src/storage/session-store.ts:375-577`) for the same session id — this is a deliberate, documented dual-writer design (see the comment at `session-store.ts:633-643`): "the engine has the token/model events, the dashboard has the aggregated hook activity... Merge non-destructively instead: counters take the max."
4. But the merge strategy is **inconsistent** between two fields that should stay coherent with each other:
   - `estimatedCostUsd` merges as one whole-session `Math.max(existing, incoming)` (`session-store.ts:540`)
   - `modelBreakdown` merges **per-field, per-model** `Math.max` independently (`session-store.ts:454-471`)

   Since only the real engine ever populates `modelBreakdown` with meaningful data, that field's max keeps the accurate value. But `estimatedCostUsd`'s single-number max lets the daemon's crude, full-session heuristic estimate win outright whenever it happens to be numerically larger than the engine's real total — which is exactly what happened here.

Notably, the codebase already recognizes and fixes this *exact* class of bug for one other field group — see the comment on `mergeEfficiencyScoreTriple` (`session-store.ts:391-399`): "merging the three independently... can attach a thin/unrelated sample count to a completely different score, radically over- or under-weighting it." The same reasoning applies to `modelBreakdown` vs `estimatedCostUsd`, but the fix was never extended there.

## Suggestions
1. Never let the byte-size fallback estimate (`recordEstimatedTokens`) become the *session-level* cost of record for a session this process doesn't own the transcript for. Options:
   - Skip the fallback entirely in `drainAllSessions`/`--local` mode for sessions with an active `--stdio` owner (detectable via the existing session-by-cwd/PID breadcrumb), since that owner's real numbers will always supersede it anyway.
   - Or: track the fallback-estimated total separately from the real per-event total, and never let `mergeSummaries()` treat it as comparable to a real-event-derived `estimatedCostUsd`.
2. At minimum, make the merge strategy for `estimatedCostUsd` consistent with `modelBreakdown`'s: since `modelBreakdown`'s per-field max is trusted as authoritative whenever it's non-empty, `estimatedCostUsd` should probably be derived from `sum(modelBreakdown[*].totalCostUsd) + subagentCostUsd` rather than merged as an independent number that can drift from it.

chrisdhaan:
## Summary

- The `--local` dashboard daemon's byte-size cost-estimate fallback was meant to cover only the handful of tool calls before a session's first real token report, but it kept firing for a session's *entire* duration whenever the daemon had no transcript connection to that session — because `costTracker.reportCount` never left `0` from its point of view, even though the session's real `--stdio` owner was already reporting accurate cost the whole time.
- Adds `shouldApplyCostEstimate()` (`src/metrics/cost-estimate-gate.ts`), a small pure gate that also checks whether another live process already owns the session (via the existing `active-<sessionId>.pid` heartbeat, `LocalStore.getActiveSessionIdsFromHeartbeats()`) before allowing the fallback to fire.
- No change to the merge strategy in `session-store.ts` — this fixes the root cause (the daemon generating a bad estimate in the first place) rather than reconciling around it after the fact.

Fixes #723

## Test plan

- [x] New unit tests in `src/metrics/cost-estimate-gate.test.ts` (written first, watched fail before implementation, per TDD)
- [x] `npx tsc -b .` — clean
- [x] `npm test` — full suite, 207/208 suites passing (1 pre-existing flaky skip), 7120 tests passing
- [x] Version bumped 1.57.0 → 1.57.1 across all 5 manifest files + CHANGELOG entry added


adjohn:
## Automated review at 5f53335

**Verdict:** needs changes
**CI:** verify green at 5f53335
**Scope:** matches issue #723, suggestion 1. Suggestion 2 (making `estimatedCostUsd`'s merge consistent with `modelBreakdown`'s) is deliberately left out, which the PR body states and justifies. The version bump touches all five manifests plus the CHANGELOG.

### Findings

1. **[bug] `isUnscopedProcess` is a startup snapshot, so a provisional `--stdio` session permanently gates off its own cost fallback.** `src/index.ts:2061`. `isProvisional` is true whenever session ID resolution did not complete synchronously (`src/index.ts:1047`). The `const` captures that moment and the `onRecord` closure reads it forever. Resolution then runs `adoptRealSessionId()`, which writes this process's own heartbeat at `src/index.ts:2948` and corrects the drain scoping at `src/index.ts:2958` by passing `drainAllSessions: false`. It does not correct `isUnscopedProcess`.

   Scenario. A `--stdio` engine starts before Claude Code's session breadcrumb is readable, so `sessionTraceId` is `pending-<ts>` and `isUnscopedProcess` is `true`. The id resolves to `S`, `active-S.pid` now holds this process's own PID, and the scoped store drains only `S`, so every subsequent record carries `sessionId === S`. The gate at `src/metrics/cost-estimate-gate.ts:25-28` then sees `isUnscopedSession: true` and `liveOwnedSessionIds.has(S) === true` and returns `false` on every call. The byte-size fallback never fires for that session, which is exactly the pre-first-token-report window the fallback exists to cover. The session's `estimatedCostUsd` stays 0 until a real token report lands, and stays 0 for the whole session on any platform that never reports tokens.

   The test at `src/metrics/cost-estimate-gate.test.ts:59` asserts the protection this wiring defeats. The function is right, the value handed to it is stale.

   Fix: derive the flag from live state at call time rather than from startup, for example a `let isUnscopedProcess` set to `false` alongside the `replaceStore(realLocalStore, false)` call, so the two facts that must agree are written in one place.

2. **[bug] `getActiveSessionIdsFromHeartbeats()` runs on every tool-call record, including when its result is never consulted.** `src/index.ts:2240`. It is an argument expression, so it evaluates before `shouldApplyCostEstimate`'s cheap `estimateBytes <= 0` and `reportCount !== 0` early-outs, and before the `isUnscopedSession` check that decides whether the set matters at all.

   Scenario. The call is a synchronous `readdirSync` of the storage directory plus a `readFileSync` and a liveness `kill` per `active-*.pid` file (`src/storage/local-store.ts:855-878`). With a `--local` daemon draining a machine with 10 live sessions, a poll that yields 100 buffered records performs 100 directory scans, 1000 file reads and 1000 signal checks on the main thread. On the `--stdio` path this blocks the same event loop that serves JSON-RPC. Before this PR the branch did no I/O at all.

   Every other caller runs it once per GC or discovery pass (`src/index.ts:367`, `src/hooks/parent-transcript-watcher.ts:315`). `src/hooks/copilot-usage-watcher.ts:246-249` goes further and skips the scan entirely when the process is scoped. Fix: compute the set only when the process is currently unscoped, or pass a lazy accessor the gate calls last.

3. **[convention] New wiring in `src/index.ts` has no test, and the defect lives in the wiring.** `src/index.ts:2061`. The added tests exercise the pure function only, so they pass with or without finding 1 present. `src/index.test.ts:526-545` already stands up a `LocalStore` double with `getActiveSessionIdsFromHeartbeats` mocked, so a case covering the provisional-then-resolved transition has a harness waiting for it.

### Notes (non-blocking)

- `CostEstimateGateParams.sessionId` is typed `string | null | undefined` (`src/metrics/cost-estimate-gate.ts:17`), but the only caller passes `ToolCallRecord.sessionId`, typed `string | null` (`src/storage/types.ts:295`). The `!== undefined` branch is unreachable. Narrow the field to `string | null` and drop the check.
- Patch bump is the right call here. This is a fix with no new surface, so 1.57.0 to 1.57.1 fits, and no docs under `docs/` needed updating.
- Checked and clean: no edits under `src/shared/`, no `as any` or `: any`, no `eslint-disable`, no stdout writes, `.js` extensions on both new imports, co-located `*.test.ts` present, CHANGELOG entry present.
- Principles applied. **Model the Domain** drove finding 1. Provisional-to-resolved is a lifecycle transition the code already handles at one point, and the PR added a second boolean that has to stay in sync with `drainAllSessions` and does not. **Prove It Works** drove finding 3. The tests verify the unit, not the artifact, so the real failure path is unverified. **Laziness Protocol** drove finding 2, since `copilot-usage-watcher.ts` already establishes the cheaper shape and the diff should follow it rather than pay the scan unconditionally. **Type System Discipline** drove the first note.

_Automated review; a maintainer makes the final call. Push a new commit for a fresh review._

PR #779, opened 2026-09-19
Fix #773: Clear ready_for_review when a PR merges in another session

### LINKED ISSUE #773: Session status "ready for review" never clears when a PR is merged from a different session
## Summary

A session's PR lifecycle status (added in #729 / #693) is derived by matching a `gh pr create` (or MCP PR-create) event against a later `merge` event for the same PR number — but only within that one session's own recorded activity.

If the PR is merged via `gh pr merge` (or an MCP merge call) run from a **different** session than the one that created it — e.g. the user opens the PR in one session, then runs a separate follow-up session later to merge it — the merge event gets attributed to the merging session, not the creating session. The creating session's `create` record is never checked against another session's records, so it keeps reading as `ready_for_review` indefinitely, even though the PR has long since merged.

## Where this lives

`computeSessionStatusAggregate` in `src/dashboard/routes/api-handler.ts` groups PR activity records into `prRecordsBySession` strictly by `record.sessionId`, then `countOpenPrs` only scans for a matching `merge` within that same per-session array.

## Why it matters

This undermines the main point of the `ready_for_review` status: it's meant to disappear once the PR is actually merged, but for any workflow where creation and merge happen in separate sessions, it won't.

Surfaced during review of #729.

dyk1454683243-sudo:
## Summary

- Parse the PR number from a create's tool response (`gh pr create` stdout URL or MCP `html_url`/`number`) and store it on the `create` event. Collector extracts only that number, not raw stdout.
- `computeSessionStatusAggregate` builds one set of merged PR numbers from every `merge` record in the today window, across all sessions. A session's create counts as open only when its number is unknown or absent from that set.
- A create with `prNumber: null` stays `ready_for_review` (conservative — we cannot prove it merged). A merge on a later day is outside the today window and still does not clear the status until the session ages out.

Fixes #773

## Test plan

- [x] Unit: `countOpenPrCreates` — cross-set match, missing number, null conservative, empty merged set (cross-day limitation)
- [x] Unit: `parsePrNumberFromToolResponse` — URL, `owner/repo#N`, MCP payload, ignored lone `number`
- [x] Recorder: create with URL in tool response; create with null `prNumber` stays null; CLI number wins
- [x] Collector: Bash stdout URL → `prNumber`; MCP `html_url` → `prNumber`; unrelated stdout still drops raw output
- [x] Jest focused suites: pr-number, session-status, git-activity-recorder, collector-script, tool-parsers, git-efficiency-tracker, api-handler session-status cases
- [x] Full Jest: 7140 passed; 1 DST concurrency failure is TZ-dependent and passes with `TZ=America/New_York` (CI pins that)
- [x] `npm run lint` and `npm run format:check` clean; `tsc --build` clean; plugin collector bundle rebuilt


dyk1454683243-sudo:

Issue #683, opened 2026-09-14
Extend the toolUseId agent-attribution backfill to ToolSelectionScorer's other call sites


chrisdhaan:
## Problem

PR #680 (#656) added a second `backfillAgentId()` join pass immediately before `AntiPatternDetector.analyze(task.toolCalls)` in `src/index.ts`, since by task-close much more real time has passed for `SubagentWatcher`'s transcript-tailing (2s poll) to have caught up than at hook-buffer intake time (~100ms poll) — meaningfully raising the attribution hit rate for that one call site.

`ToolSelectionScorer.scoreSession()` has several other call sites that read the same class of accumulated `ToolCallRecord[]` and would benefit the same way, but weren't touched:
- `src/metrics/local-session-aggregator.ts`
- `src/storage/session-store.ts`
- `src/dashboard/routes/api-handler.ts`
- `src/tools/extended-analytics-tools.ts`

## Impact

These paths keep the old, lower hit-rate behavior (one-shot backfill at intake only) for tool-selection scoring specifically, even though the anti-pattern detection path (which shares the same underlying `partitionByAgent` grouping) was improved.

## Fix direction

For `local-session-aggregator.ts` (same live process, session-close time): thread the same `toolUseIdToAgentId` map through as a parameter and apply the same `.map(r => backfillAgentId(r, toolUseIdToAgentId))` pass before `scoreSession()`.

For `session-store.ts` / `api-handler.ts` / `extended-analytics-tools.ts`: these read *persisted* `ToolCallRecord[]` data, potentially from a different process than the one that built the in-memory map — the live map may not exist at all by the time these run. Needs a design decision: either persist enough information for a later process to reconstruct/re-derive the join, or accept these paths stay best-effort-at-intake-only (and document that explicitly, rather than leaving it implicit).

## Context

Follow-up from #656 / PR #680, called out explicitly in that PR's final review (Important finding #4) and deliberately scoped out of that fix to keep the review wave focused — see PR #680's review ledger ruling on that finding.

dyk1454683243-sudo:
Design decision before the code lands — live late-join vs persisted intake-only.

[...]

Issue #682, opened 2026-09-14
toolUseIdToAgentId / agentTypeByAgentId maps grow unbounded for process lifetime


chrisdhaan:
## Problem

PR #680 (#656) added `toolUseIdToAgentId`, a `Map<string, string>` in `src/index.ts` populated from `SubagentWatcher`'s `onSubagentTurn` callback — one entry per subagent tool-use id, for the process's entire lifetime, with no eviction. It sits alongside the pre-existing `agentTypeByAgentId` map, which has the same unbounded-growth characteristic.

## Impact

Negligible for a short-lived `--stdio` session (the common case, tied to one Claude Code session). A real, slow memory leak for a long-running `--local` daemon that drains every session's buffer indefinitely — both maps accumulate one entry per subagent tool call / per subagent-type-correlated Agent call across every session it ever observes, forever.

## Fix direction

Some combination of:
- Bound each map's size (LRU-style eviction) and/or
- Prune entries once the owning session/task is known to be closed, and/or
- Time-based expiry (a tool-use id from a session that ended hours ago has no further use)

## Context

Follow-up from #656 / PR #680, flagged by that PR's Task 6 implementer as "worth a follow-up ticket if the controller agrees" and confirmed as a real (if low-urgency) gap during the final review.

dyk1454683243-sudo:
Chosen eviction strategy:

1. **LRU size cap** — `toolUseIdToAgentId` max **10,000**; `agentTypeByAgentId` max **2,048**. `get()` / re-`set()` refresh LRU so in-flight #656 / PR #680 `backfillAgentId()` joins stay resident.
2. **Idle TTL (2 hours)** — unused keys expire. Task analysis closes in tens of seconds; a tool-use id from an ended session has no further join use.
3. **Session prune on local-day rollover** — entries are tagged with `sessionId` / `parentSessionId`. When the live registry rolls yesterday’s ids off “seen today”, those correlations are dropped. `pruneSession()` is also available for an explicit close path.

Together these keep a long-running `--local` daemon bounded without breaking recent in-flight attribution.

PR: https://github.com/newrelic-experimental/preflight/pull/781 (`Fixes #682`).

PR #780, opened 2026-09-19
Fix #681: Backfill ToolCallRecord.agentType from agentTypeByAgentId

### LINKED ISSUE #681: ToolCallRecord.agentType (audit-trail/security-event attribution) still never populates
## Problem

#656 (PR #680) fixed `ToolCallRecord.agentId` via a `toolUseId` join against subagent transcripts, and fixed `CostTracker.subagentCostByAgentType` via a separate correlation on the parent's own `Agent`-tool-call record (`spawnedAgentId` + `subagentType`). Neither of those touches `ToolCallRecord.agentType` itself, which is still sourced only from the hook envelope's `agent_type` field — the same field #656 established never populates in practice.

`src/security/audit-trail.ts` reads `record.agentType` directly (see `AuditTrailManager`'s persisted-record construction, e.g. lines ~533 and ~652), and that flows into `AiAuditEvent`/`SecurityAlert` on the wire (`src/transport/nr-ingest.ts`). So sensitive/destructive action audit records still can't be attributed to which subagent type performed them.

## Impact

Audit trail attribution by subagent type is silently broken for the exact same reason #656's other two mechanisms were — just not covered by that fix's scope.

## Fix direction

Backfill `ToolCallRecord.agentType` the same way `agentId` is now backfilled (`backfillAgentId()` in `src/metrics/agent-partition.ts`), or extend the existing `agentId` backfill to also carry a type where available. Needs a design decision on where the type signal comes from per-call (the parent's own Agent-tool correlation only knows the type at the *call* level, not per individual tool-use — same limitation `agentTypeByAgentId` already has for cost breakdown).

## Context

Follow-up from #656 / PR #680, deliberately left out of that fix's scope (see PR #680's review ledger, "Ruling" on Important finding #3).

dyk1454683243-sudo:
## Summary

- Fixes #681.
- After `backfillAgentId()` resolves `agentId` (hook envelope or `toolUseId` join), sibling `backfillAgentType()` fills `ToolCallRecord.agentType` from the existing `agentTypeByAgentId` map — the parent Agent-tool correlation (`spawnedAgentId` + `subagentType`) CostTracker already uses for cost-by-type.
- Same per-`agentId` limitation as that cost breakdown: not per tool-use. `agentType` stays undefined when there is no type signal (parent call, late id join, unpaired Agent spawn). No invented labels.
- `AuditTrailManager` already copies `record.agentType` onto the persisted audit record, `AiAuditEvent`, and `SecurityAlert`. Those paths now receive the backfilled value because they already read the field.
- Design decision on the issue before the code: https://github.com/newrelic-experimental/preflight/issues/681#issuecomment-5739216383

Out of scope: #682 (map bounding), #683 (scorer call-site coverage), and open PR #779.

## Test plan

- [x] `npx jest -- src/metrics/agent-partition.test.ts src/security/audit-trail.test.ts` (162 passed)
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run format:check`


dyk1454683243-sudo:

PR #748, opened 2026-09-18
Fix #732: Remove stale costPeriod unimplemented warning

dyk1454683243-sudo:
## Summary

- Delete the stale `costPeriod today/week not yet implemented` warning in `loadAlertRulesFromDisk` (`src/index.ts`). Those periods have been wired through `BudgetTracker` since #146, so the warning told users a working rule would never fire.
- Update `AlertSnapshot` and `computeCostWindowValue` JSDoc: keep that cost is period-cumulative (session / today / week), not a rolling N-second window; drop “unimplemented” wording. Also refresh nearby comments that still described the old warning.
- Add a collector + `BudgetTracker` test: a `cost.window` rule with `costPeriod: 'today'` fires when `budgetTracker.getStatus().daily.spentUsd` crosses the threshold. This fails if the `budgetTracker` dep is removed from `AlertSnapshotCollector.readCost()`.

Fixes #732.

## Test plan

- [x] `npx jest -- src/alerts/local-alert-engine.test.ts src/alerts/alert-snapshot-collector.test.ts src/alerts/local-alert-rule.test.ts` (80 passed)
- [x] `npm run lint` / `npm run format:check` (pre-commit)
- [x] `npm run build`
- [x] Repo grep: no log line mentions `costPeriod` being unimplemented
- [ ] `npm test` on CI (Node 24, `TZ=America/New_York`)

PR #771, opened 2026-09-18
Docs #734: Inventory MCP tools and config env vars

dyk1454683243-sudo:
Fixes #734.

## Summary

`scripts/check-docs-inventory.ts` already guards emitted event types and documented metric names. Two other surfaces were drifting with nothing checking them. This PR extends the existing CI check and closes the gaps it reports, in one reviewable change.

- **Checker:** every `nr_observe_*` literal under `src/` (same file set as the event scan: tests, `src/web`, and `src/shared` excluded) must appear in `docs/COMMANDS_TABLE.md`.
- **Checker:** every `NEW_RELIC_AI_*` literal in `src/config.ts` must appear in at least one of `README.md`, `docs/ADVANCED.md`, `SECURITY.md`, or `PRIVACY.md` — the four operator-facing docs the issue named.
- **`docs/COMMANDS_TABLE.md`:** documents the four tools the check found missing — `nr_observe_get_compute_waste` plus the generic-MCP self-report trio (`nr_observe_report_tool_call`, `nr_observe_report_session_start`, `nr_observe_report_session_end`).
- **`docs/ADVANCED.md`:** adds an environment-variable reference for the server's own `NEW_RELIC_AI_*` names that no operator-facing doc named, grouped by server process, identity, budgets/digests, and homelab forwarding, each with its config-file key and default.
- **`PRIVACY.md`:** names the env var for the three privacy settings that were documented under their config-file key only (`NEW_RELIC_AI_MCP_RECORD_CONTENT`, `NEW_RELIC_AI_HIGH_SECURITY`, `NEW_RELIC_AI_RETAIN_SESSIONS_DAYS`).

The check already runs in CI (`npm run check:docs-inventory`); nothing else to wire.

## Test plan

- [x] `npm run check:docs-inventory` passes: `15` event types, `72` metric names, `49` MCP tools, `32` config env vars
- [x] Deleting `nr_observe_get_compute_waste` from `docs/COMMANDS_TABLE.md` fails with that exact name
- [x] Deleting `NEW_RELIC_AI_MCP_STORAGE_PATH` from `docs/ADVANCED.md` fails with that exact name
- [x] Restoring either name makes the check pass again
- [x] `npx prettier --check` clean (repo-wide)
- [x] `npm run lint` — 0 errors, 0 warnings
- [x] `npm run build` — server + web clean
[...]

PR #781, opened 2026-09-19
Fix #682: Bound toolUseIdToAgentId and agentTypeByAgentId maps

### LINKED ISSUE #682: toolUseIdToAgentId / agentTypeByAgentId maps grow unbounded for process lifetime

dyk1454683243-sudo:
## Summary

- Fixes #682.
- Replaces the process-lifetime `Map`s in `src/index.ts` (`toolUseIdToAgentId`, `agentTypeByAgentId`) with `BoundedLruMap`: LRU size cap + idle TTL, with optional session tags.
- **Caps:** 10,000 tool-use joins; 2,048 agent-type correlations.
- **TTL:** 2 hours idle. `get()` / re-`set()` refresh LRU and TTL so recent in-flight `backfillAgentId()` joins from #656 / PR #680 still resolve.
- **Session prune:** entries carry `sessionId` / `parentSessionId`. On local-day rollover, ids that the live registry has rolled off “seen today” are dropped (`retainSessions`). `pruneSession()` covers an explicit close path.
- Strategy noted on the issue: https://github.com/newrelic-experimental/preflight/issues/682#issuecomment-5739269586

Out of scope: open PR #779 (`ready_for_review` / #773), #681 / PR #780 (agentType backfill), and closed #748–#771 work.

## Test plan

- [x] `npx jest -- src/lib/bounded-lru-map.test.ts src/metrics/agent-partition.test.ts` — fill past bound evicts oldest; TTL expiry; session prune / retain; short-session backfill still works; evicted/pruned ids do not backfill
- [x] `npx jest -- src/index.test.ts --testNamePattern="agent-id correlation"` — both maps constructed as `BoundedLruMap` with TTL + day-rollover retain
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] Full Jest under `TZ=America/New_York` (matches CI): 7133 passed; the only remaining failure in this environment was a pre-existing `workflow-watcher` partial-write flake unrelated to these maps


dyk1454683243-sudo:

PR #782, opened 2026-09-19
Fix #683: Late toolUseId backfill before live tool-selection scoring

### LINKED ISSUE #683: Extend the toolUseId agent-attribution backfill to ToolSelectionScorer's other call sites

dyk1454683243-sudo:
## Summary

- Fixes #683.
- Live path: `LocalSessionAggregator.toSummaries()` now takes the process-local `toolUseIdToAgentId` map and applies `.map(r => backfillAgentId(r, toolUseIdToAgentId))` immediately before `scoreSession()`, matching the late join PR #680 added before `AntiPatternDetector.analyze()`.
- `src/index.ts` threads that map through at persist time so session-close scoring sees mappings `SubagentWatcher` added after hook-buffer intake.
- Persisted / out-of-process `scoreSession()` sites stay **best-effort-at-intake-only**, documented next to the call sites (and on the issue):
  - `src/storage/session-store.ts` (`buildSessionSummary`)
  - `src/dashboard/routes/api-handler.ts` (today-aggregate + live session detail)
  - `src/tools/extended-analytics-tools.ts` (`handleGetToolSelectionScore`)
- After #680, `ToolCallRecord` already persists `toolUseId` (join key) and `agentId` (join result when intake succeeded). `ReplayTimelineEntry` does not persist `toolUseId`. This PR does **not** invent a new on-disk join-map schema.
- Design comment on the issue: https://github.com/newrelic-experimental/preflight/issues/683#issuecomment-5739343622

Out of scope: #681 / PR #780 (`agentType`), #682 / PR #781 (map bounding), open PR #779.

## Test plan

- [x] `npx jest -- src/metrics/local-session-aggregator.test.ts` (50 passed) — late-arriving `toolUseId` map improves `scoreSession` attribution vs intake-only
- [x] `npx jest -- src/tools/extended-analytics-tools.test.ts src/storage/session-store.test.ts src/metrics/agent-partition.test.ts` (157 passed)
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run format:check`


dyk1454683243-sudo:

PR #783, opened 2026-09-19
Fix #671: Exclude failed and amend commits from per-session Git Efficiency

dyk1454683243-sudo:
## Summary

- Fixes #671.
- Per-session `GitEfficiencyTracker` counted every `commit`-type event, including failed commits (e.g. pre-commit hook rejection) and `git commit --amend`. The Git tab's weekly/30-day panel already excluded those via `isCountedCommit` (#663/#670), so the two views could disagree on the same session.
- `isCountedCommit` now lives next to `GitEvent` in `git-event-classifier.ts` and is the shared helper for both paths. `git-workspace-report.ts` re-exports it so existing weekly/replay callers stay unchanged.
- Per-session `commitCount`, `commitTimestamps`, and `commitsSinceLastSync` use that filter. Failed and amend events still appear on the git command timeline (they happened); they just no longer inflate commit totals. Hydrated `git log` commits still count.

Out of scope: #672 (chained git/gh error attribution), #755 (timezone/date behavior), and open PRs #776/#779/#780/#781/#782.

## Test plan

- [x] `npx jest -- src/metrics/git-efficiency-tracker.test.ts src/metrics/git-event-classifier.test.ts src/metrics/git-workspace-report.test.ts` (198 passed)
- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run format:check`

PR #774, opened 2026-09-18
Refactor: remove the permanent costByTask: null stub from CostMetrics

PloneMraz:
## Summary

Closes #733.

`CostMetrics.costByTask` was typed `null` with the comment "stub — task boundary detection is not yet implemented". Task boundary detection landed: `TaskDetector` takes a `costTracker` and calls `computeCostDelta()` at task open and close, so per-task cost already lives on the task records it produces. The field was a permanent `null` that every consumer carried and nothing populated.

Removed from:

- `src/metrics/cost-tracker.ts` — the `CostMetrics` field and the `getMetrics()` literal
- `src/web/api/client.ts` — the mirrored type
- fixtures in `cost-tracker.test.ts`, `cost-tools.test.ts`, `session-store.test.ts` (3 occurrences)

`grep -rn costByTask src docs` now returns nothing. Neither `docs/METRICS_TABLE.md` nor `docs/COMMANDS_TABLE.md` mentioned it.

## Test plan

Verified on Linux in a container (`node:24`, Node v24.21.0, `TZ=UTC`), same `node_modules`, only the changed files swapped between runs:

| | Test suites | Tests |
|---|---|---|
| `main` @ 9859468 | 5 failed, 200 passed | 6 failed, 7093 passed (7103) |
| this branch | 5 failed, 200 passed | 6 failed, 7093 passed (7103) |

- [x] `npm test` — identical to `main`; the 6 failures are environmental and present on both sides (a DST test under `TZ=UTC`; two `plist unreadable` and one `directory not writable` case that root bypasses; `isWsl`, since Docker Desktop is a WSL2 kernel)
- [x] `npm run test:web` — **48/48 files, 706/706 tests pass** on Linux
- [x] `npx tsc -b .` and `npx tsc -p tsconfig.web.json --noEmit` — clean
- [x] `npm run lint`, `npx prettier --check .` — clean

Note for anyone reproducing: the first full run against a freshly written bind mount reports ~11 extra failures in `src/index.test.ts` purely from the 15s `testTimeout` on cold I/O — that file passes 82/82 in isolation. The numbers above are from warm runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #631, opened 2026-09-09
Correction detection: remaining false-positive edge cases from #608's regex redesign


chrisdhaan:
## Context

#608 replaced the single anchored \`CORRECTION_RE\` regex in \`src/metrics/transcript-message-tracker.ts\` with a set of targeted patterns (\`isCorrectionMessage()\`), fixing every false positive/negative example named in that issue. While designing the fix, three additional false-positive cases surfaced that weren't part of #608's original examples and aren't covered by its test suite.

## Problem

- **\`TARGETED_UNDO_RE\`** (\`/^(stop|undo|revert|don'?t)\b[^.!?]{0,20}\b(that|it|this)\b/i\`) fires on task instructions/guardrails that happen to contain "that"/"it"/"this" near the trigger verb, not just on genuine undo requests targeting the assistant's prior action. Example: **"Don't push directly to that branch"** — a standing instruction, not a correction — matches because "that" appears within the 20-character window after "don't".

- **\`LEADING_INTERJECTION_RE\`** (\`/^(stop|wait|no|nope|undo|revert)[.,!]/i\`) treats any leading trigger word immediately followed by punctuation as an interjection/correction. Example: **"No, thanks"** or **"No, that's fine"** — a polite decline, not a correction — matches purely on the "No," prefix.

- **\`EMBEDDED_CORRECTION_RE\`**'s \`won'?t work\` clause can fire during ordinary design discussion rather than an actual correction of something the assistant already did. Example: **"That approach won't work for X, let's use Y instead"** — a forward-looking design refinement — reads the same as a correction of prior AI output.

## Impact

Same impact chain as #608: inflates \`userCorrections\`, which feeds \`correctionRate\` (collaboration-profile.ts), the high-priority "high correction rate" recommendation (prompt-feedback.ts), the team-level version (recommendation-engine.ts), the CLAUDE.md-change verdict (claudemd-tracker.ts), and the 'Learning' classification threshold.

## Possible fix

- For \`TARGETED_UNDO_RE\`: narrow the pronoun-proximity window, or require the pronoun immediately follow the verb (no intervening noun phrase) — "don't push directly to **that** branch" has "that" modifying a noun ("branch"), whereas "undo **that**" / "don't do **that**" has the pronoun standing alone as the object.
[...]

PR #784, opened 2026-09-19
Fix #677: Narrow won't work correction false positive

### LINKED ISSUE #677: Correction detection: EMBEDDED_CORRECTION_RE's "won't work" clause false-positives on forward-looking design discussion
## Context

#631 named three false-positive edge cases in `src/metrics/transcript-message-tracker.ts`'s correction-detection regexes, surfaced during review of #608's regex redesign. [PR #676](https://github.com/newrelic-experimental/preflight/pull/676) fixed the first two (`TARGETED_UNDO_RE` firing on standing instructions like "don't push directly to that branch"; `LEADING_INTERJECTION_RE`/`LEADING_NO_RE` firing on polite declines like "No, thanks"). The third — `EMBEDDED_CORRECTION_RE`'s `won't work` clause — was deliberately left unfixed in that PR, per #631's own guidance that it's lower-value and risks reopening the false-positive/recall tension #608 was resolving. Filing it separately so it stays tracked rather than getting silently dropped when #631 is closed.

## Problem

`EMBEDDED_CORRECTION_RE` (`src/metrics/transcript-message-tracker.ts`) includes a `won'?t work` clause intended to catch the user telling the assistant its prior output doesn't work. It can't distinguish that from ordinary forward-looking design discussion that happens to use the same phrase.

Example: **"That approach won't work for X, let's use Y instead"** — a forward-looking design refinement, not a correction of something the assistant already did — matches identically to a genuine correction like "That approach won't work because there's a race condition." (the latter is an existing, must-still-detect test case).

## Impact

Same downstream impact chain as #608/#631: inflates `userCorrections`, which feeds `correctionRate` (`collaboration-profile.ts`), the "high correction rate" recommendation (`prompt-feedback.ts`), the team-level version (`recommendation-engine.ts`), the CLAUDE.md-change verdict (`claudemd-tracker.ts`), and the 'Learning' classification threshold.

## Possible fix

- Require a nearby past-tense/completed-action cue (e.g. "because", "since", a reference to what was just done) before counting `won't work` as a correction, to distinguish it from forward-looking cues ("let's", "instead", "we should").
- Any such heuristic needs to be verified against the existing positive test case ("That approach won't work because there's a race condition.") without breaking it, and against the negative example above.
- Alternatively, accept the residual false-positive rate as a permanent tradeoff — this is fundamentally a harder case than #631's other two (which had a clean structural signal to key off of), and #608/#631 already accepted similar tradeoffs elsewhere.

Filed as a deliberately-deferred follow-up from #631, same pattern as #625 following #607 and #631 following #608.

dyk1454683243-sudo:
## Summary

- Fixes #677.
- `EMBEDDED_CORRECTION_RE`'s `won't work` clause treated forward-looking design talk ("That approach won't work for X, let's use Y instead") as a user correction of prior output. That inflated `userCorrections` → `correctionRate` → recommendations / Learning classification.
- `won't work` is now classified separately. **Chosen rule:** count it as a correction unless it is constraint-framed (`won't work for …`) *and* a forward-looking planning cue (`let's` / `we should` / `instead`) is present, with no causal cue (`because` / `since`). Causal cues always win.
- This keeps the existing positive case ("That approach won't work because there's a race condition.") and a bare rejection ("That won't work.") so recall does not drop. The #631 / #676 undo and polite-decline cases are untouched.
- **Residual tradeoff (accepted, not widened):** constraint-framed `won't work for …` without a planning cue still counts (e.g. "That approach won't work for production."). A genuine correction that uses both `won't work for` and a planning cue without `because` / `since` is dropped — same shape as the issue's negative example. The regex is not widened to chase those leftovers.
- Heuristic documented on the issue: https://github.com/newrelic-experimental/preflight/issues/677#issuecomment-5739485091

Out of scope: open PRs #776 / #779 / #780 / #781 / #782 / #783; closed #748–#771 / #772 / #774 / #775.

## Test plan

- [x] `npx jest -- src/metrics/transcript-message-tracker.test.ts` — 50 passed
- [x] `npm run build` — clean
- [x] Positive `because` case still matches; forward-looking `let's` / `instead` case does not; #631 / #676 cases still pass


dyk1454683243-sudo:

PR #787, opened 2026-09-19
Feat #730: Show per-category dollars in share-table hover

### LINKED ISSUE #730: Show each token category's dollar contribution in the share-table hover, not just the token split
## Summary

The "What's contributing to your spend" share tables' hover breakdown (Skills, Subagents, Plugins) shows input/output/cache-read/cache-write *token* counts per row, but not each category's *dollar* contribution — even though cache-read tokens are priced far below input/output tokens, so the token split alone doesn't tell a reader which category is actually driving the cost.

## Why it's not trivial

The dollar split isn't reconstructable from persisted data today: a bucket spans multiple models, and pricing is per-model, so showing a per-category dollar figure would need either persisting per-model-per-category cost at attribution time, or a different aggregation approach that keeps the model dimension around long enough to price each category correctly.

## Context

#692 originally asked for "each category's dollar contribution" alongside the token breakdown. The token-only breakdown shipped as a first pass; this issue tracks the dollar-contribution part that was deliberately deferred.

Surfaced during review of #728.

dyk1454683243-sudo:
## Why

The Skills / Subagents / Plugins share-table hover already shows input / output / cache-read / cache-write **token** counts (#728), but cache-read tokens are priced far below input and output. A token-only split therefore misleads readers about what is actually driving spend.

A bucket can span models, and pricing is per-model, so category dollars cannot be reconstructed from persisted token totals with a blended rate. This PR prices each category at attribution time with that event's model rates, then sums the already-priced dollars.

Fixes #730.

## Scope

- `TokenCategoryCost` on `TokenBreakdown.cost`: `inputUsd` / `outputUsd` / `cacheReadUsd` / `cacheCreationUsd`. Optional; absent on pre-#730 session files.
- `priceTokenCategories` / `categoryCostFromBreakdown` (`src/metrics/category-cost.ts`) keep the model dimension long enough to call `calculateCost()`, then split and sum. Thinking USD is dropped on purpose.
- `TurnCostAttributor` and `CostTracker.subagentByAgentType` persist those dollars when a token event is attributed.
- `session-store` parse/merge/`buildAttribution` carry `cost` with the same max / one-sided rules as the token fields. Malformed `cost` drops only the dollars.
- `computeUsageInsights` sums category dollars into `UsageShareRow.breakdown.cost`.
- `formatTokenBreakdown` appends `($x.xx)` per category when `cost` is present; otherwise the hover stays token-only. Same plain `title` tooltip as #728.

## Tradeoffs

- Dollars are priced at write time, not re-priced at read time. Later pricing-table edits do not rewrite historical hovers.
- Thinking dollars are omitted. Parent-session token events typically lack thinking tokens (same gap #728 documented). We do not invent a figure the data does not support.
- Mixed windows (some sessions pre-#730) may show dollars for only the sessions that persisted them.

## Blast Radius

Session files gain one optional object on each bucket's `breakdown`. Existing files load unchanged and the hover stays token-only. Share-row JSON grows the same optional field. No endpoint shape change otherwise. ShareTable is not redesigned.

## Test plan

- [x] Unit tests for per-model pricing vs a fake blended rate, cache-read vs input rates, even-split after `calculateCost`, rate multiplier, thinking-dollar omission, unknown model → zeros
- [x] `TurnCostAttributor` / `CostTracker` accumulate per-model category dollars
- [x] `session-store` parse/merge/build: legacy token-only, malformed cost dropped, field-wise max, one-sided keep
- [x] `computeUsageInsights` sums dollars when present and omits them when absent
- [x] `formatTokenBreakdown` + Skills hover: dollars present vs absent
- [x] `npm run build`
- [x] `npm test` (full suite; one unrelated `workflow-watcher` flake on a pre-push run, passed in isolation)
- [x] `npm run test:web`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npx tsc -p tsconfig.web.json --noEmit`


dyk1454683243-sudo:

Issue #555, opened 2026-09-02
Package self-contained native installers (.pkg/.msi) for silent enterprise/MDM deployment


adjohn:
## Problem

Enterprises evaluating fleet-wide rollout (Jamf, Intune, Workspace ONE, etc.) have no artifact that fits their standard silent-deployment model. The only install path today is `npm install -g @newrelic/preflight`, which assumes Node.js v22+ is already provisioned on the target machine (see README Requirements). MDM postinstall/policy scripts run unattended — they can't interactively resolve a missing or wrong Node version the way a developer running `preflight setup` by hand can. That gap, not the MDM tooling itself, is what currently blocks a true "push this to every laptop" rollout.

This is a different distribution mechanism than the Homebrew tap in #516 — Homebrew still assumes a package manager and a manually-provisioned runtime, and doesn't solve unattended fleet push either.

## Proposed approach

1. Produce a self-contained native installer per OS that bundles (or otherwise doesn't depend on a pre-provisioned) Node runtime — a signed `.pkg` for macOS, a signed `.msi`/`.exe` for Windows.
2. Wire the installer to run the existing non-interactive path under the hood: `preflight install --mode cloud --license-key ... --account-id ...`, sourcing the license key/account ID from environment variables — MDM tools already support injecting these per-device (Jamf script parameters, Intune script params/Key Vault-backed variables), so no new config plumbing should be needed there.
3. Document a per-platform deployment recipe (Jamf Policy + script parameters, Intune Win32 app/script deployment) once the artifacts exist.
4. Stretch: a lightweight fleet-verification story — confirming hooks actually activated post-install across N machines. Today's `nr_observe_health`/`doctor` diagnostics are per-machine only; there's no fleet-level rollup or drift detection.

## Priority

Medium — no reported user demand yet, but this is the concrete blocker identified when scoping "how would an enterprise deploy this fleet-wide." Building and signing native installers is real ongoing packaging/CI work (macOS notarization, Windows code-signing cert), worth prioritizing deliberately rather than as a side effect of another release.

Issue #117, opened 2026-07-10
preflight setup needs to auto-create hook files for Kiro, and ADAPTERS.md should document them


ryoryotaro:
### Summary

When setting up Preflight with Amazon Kiro, two things are missing:

1. `preflight setup` does not automatically create hook files for Kiro
2. The Kiro section in ADAPTERS.md does not mention hooks at all

### Current behavior

- `preflight setup` writes hooks to `~/.claude/settings.json` for Claude Code, but does nothing for Kiro's hook system
- The Kiro section in ADAPTERS.md only documents MCP server configuration — no mention of hook files
- Users have to manually create PreToolUse/PostToolUse hook files under `~/.kiro/hooks/`

### Expected behavior

- `preflight setup` (or `preflight install`) should detect Kiro and automatically write the hook files to `~/.kiro/hooks/`
- ADAPTERS.md Kiro section should include the hook file setup steps

### Environment

- Preflight version: 1.4.3
- Platform: Amazon Kiro (macOS)
- Node.js: v22.23.1

PR #788, opened 2026-09-19
Feat: preflight install configures every detected assistant

### LINKED ISSUE #721: preflight install should configure every detected AI assistant, not only Claude Code
## Problem

A developer who uses more than one AI coding assistant (Claude Code, Cursor, Kiro, Windsurf) has to configure Preflight once per assistant, and only one of those configurations is automated.

`preflight install` writes hooks into `~/.claude/settings.json` and registers the MCP server for Claude Code. It accepts `--license-key`, `--account-id`, `--mode`, `--project`, and the two WSL flags. It has no way to target any other assistant (`src/install/cli.ts`, the `install` command; the platform resolver at lines 73 to 134 only distinguishes native from WSL).

Every adapter already implements `getHookInstallInstructions()`, but no CLI command calls it. ADAPTERS.md says so in its maintenance note at the top, and `preflight doctor --platform <x>` tells the user to verify manually.

This came up as a public question on a LinkedIn post about Preflight: "If a single developer uses 3 different paid AI subscriptions (e.g., Claude Code, Kiro, Cursor), can a single non-interactive command automatically register hooks across all tools at once, or do we need separate installers/configuration flows?" The honest answer today is separate flows.

## Current state per assistant

| Assistant | What `preflight install` does | What the user does by hand |
|---|---|---|
| Claude Code | Merges hooks into `settings.json`, registers MCP server. Idempotent. | Nothing |
| VS Code Copilot Chat | Nothing, but VS Code reads the Claude-format hooks file by default, so the Claude Code install is picked up | Nothing, unless they also use `~/.copilot/hooks` (double capture) |
| Cursor | Nothing | Create `~/.cursor/hooks.json` with six events pointing at `preflight-collector`, add MCP server entry |
| Windsurf | Nothing | Create `~/.windsurf/hooks.json`, add MCP server entry |
| Kiro | Nothing | Add MCP server entry to `~/.kiro/settings/mcp.json`; hook files under `~/.kiro/hooks/` are undocumented (#117) |
| Amazon Q Developer CLI | Nothing | Per ADAPTERS.md |

All of these feed the same `preflight-collector` binary and the same storage directory, and one Preflight instance drains every buffer there. The runtime already treats several assistants on one machine as one developer. Only the install step does not.

## Proposal

1. `preflight install` detects which assistants are present on the machine (config directory exists: `~/.cursor`, `~/.windsurf`, `~/.kiro`, `~/.aws/amazonq`, and so on) and writes the hook file and MCP entry for each one, using the same merge-and-filter approach `mergeSettings()` and `mergeMcpConfig()` use for Claude Code, so re-running is idempotent and never clobbers a user's other hooks.
2. Add `--assistants <list>` (and `--assistants all`) to override detection, so an unattended install can pre-configure an assistant that is not installed yet. This is what a fleet script or the #555 installer would call.
3. `preflight uninstall` mirrors it, removing only Preflight's own entries from each file.
4. `preflight doctor` reports per-assistant status from the same table instead of asking the user to verify manually.
5. Move the per-assistant setup text in ADAPTERS.md to point at the command output, per that file's own maintenance note.

The hook and MCP JSON for each assistant is already written down in ADAPTERS.md, so the main work is the detection table, the file writers, and tests around the merge.

## Relationship to open issues

- #117 asks for auto-created Kiro hook files. That is one row of this table and should land as part of it, or first.
- #555 asks for signed `.pkg` and `.msi` installers for MDM rollout. Those wrap the non-interactive install command, so this issue decides what that command configures.

dyk1454683243-sudo:
## Summary

- `preflight install` now detects which AI coding assistants are present (config directory exists: `~/.cursor`, `~/.windsurf` / `~/.codeium/windsurf`, `~/.kiro`, `~/.aws/amazonq`, `~/.copilot`, `~/.factory`, `~/.gemini`, `~/.codex`) and writes hook + MCP entries for each one, not only Claude Code.
- Each file-based writer uses the same merge-and-filter semantics as Claude Code's `mergeSettings()` / `mergeMcpConfig()`: strip our previous entries, append the current ones, never clobber unrelated user hooks. Re-running is idempotent.
- `--assistants <list>` and `--assistants all` override detection so an unattended / fleet install can pre-configure an assistant that is not installed yet.
- `preflight uninstall` mirrors the table and removes only Preflight's own entries (including leftover `--assistants all` files). `--assistants cursor` restricts cleanup.
- `preflight doctor` reports per-assistant hook/MCP status from the same table (`ok` / `partial` / `missing` / `malformed`) instead of “verify manually” for those platforms.
- `docs/ADAPTERS.md` setup sections now point at the command output; installer JSON stays in each adapter's `getHookInstallInstructions()`.
- Claude Code (WSL path targeting) and Copilot (VS Code collision fix) keep their dedicated writers. Plugin-based and mcp-tools-only platforms are unchanged. No signed `.pkg`/`.msi` (#555).

Fixes #721. Includes Kiro hook-file install from #117 as one row of the table.

## Test plan

- [x] `npx jest -- src/install/assistant-install.test.ts src/install/cli.test.ts src/install/diagnostics.test.ts src/install/setup-wizard.test.ts` (358 passed)
- [x] Detection via config dirs; `--assistants` / `--assistants all` override; unknown names rejected
- [x] Merge idempotency and uninstall cleanup in temp dirs (no real `$HOME` writes)
- [x] `npm run build`
- [x] `npm run lint` / Prettier on changed files
- [ ] CI on this PR


dyk1454683243-sudo:

Issue #671, opened 2026-09-12
Per-session Git Efficiency tab still counts failed and amended commits


chrisdhaan:
## Summary

The Git tab's weekly/30-day panel (fixed in #663/#670) now excludes failed and amend commits from its commit count via `isCountedCommit`. The per-session `GitEfficiencyTracker`, which powers the per-session Git Efficiency view, was not updated to match — it still counts every `commit`-type event regardless of `record.success` or an `--amend` flag.

## Background

From #663's own PR description (Tradeoffs section):

> The per-session `GitEfficiencyTracker` now classifies per segment too, but keeps its own commit counting rule, so a failed or amended commit still counts there. It hydrates from git log for today with its own 5-second dedup, and changing that is a separate fix.

So a session that ran a failing `git commit` (e.g. a pre-commit hook rejection) or an `--amend` will show an inflated commit count on the per-session tab, while the weekly/30-day view for the same activity now correctly excludes it. The two views can disagree on the same underlying data.

## Proposed fix

Apply the same `isCountedCommit`-style filter (success + not-amend) to `GitEfficiencyTracker`'s commit counting in `src/metrics/git-efficiency-tracker.ts`, so per-session and weekly/30-day commit counts stay consistent.

Issue #486, opened 2026-08-27
Engaged/active time metric — collector silently ignores lifecycle hook events


adjohn:
## Problem
`ai.session.duration_ms` is pure wall clock from MCP-process construction (`src/metrics/session-tracker.ts:158,288,332`) — a session left open overnight counts every hour. Claude Code's OTel exports `active_time.total` (active vs. idle), which is the number managers actually want for engagement.

The ingredients exist but aren't wired: `collector-script.ts`'s dispatch chain handles only tool-shaped events and silently ignores `UserPromptSubmit`/`SessionStart`/`SessionEnd`/`Stop` (unknown-event fallthrough at the end of the chain), and the installer wires only the two tool hooks. Gap machinery exists only at tool granularity (`TurnTracker` 2s bursts, `src/metrics/turn-tracker.ts:32-35`; `TaskDetector` idle timeout; `LiveSessionRegistry` 180s liveness).

## Proposed fix
- Collector branches + installer wiring for `UserPromptSubmit`, `SessionStart`, `SessionEnd`, `Stop`.
- A gap-coalescing accumulator (prompt→Stop spans unioned with tool-call bursts, capped idle gaps) producing e.g. `ai.session.engaged_ms` alongside wall-clock duration.

## Honest ceiling
Claude Code's definition includes "actively typing, reading" — keystroke/focus states hooks can never see. This should ship as a distinct *engaged time* metric, documented as an approximation, not claimed as parity.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #789, opened 2026-09-19
Fix #672: Don't attribute chained gh failures to last git segment

### LINKED ISSUE #672: Chained git/gh command errors are attributed to the wrong segment when the failure isn't the last git/gh segment
## Summary

When a shell command chains multiple git/gh segments with `&&`, error output from the tool call is attributed to the *last* git segment in the chain — but the failure isn't always in the last segment. A command like `git push && gh pr create` that fails on the `gh pr create` step still hands the error text to the `push` segment, since that's the last *git* segment `classifyGitSegments` sees (the `gh` command is classified separately).

## Background

From #663's PR description (Tradeoffs section):

> Error text is attributed to the last git segment. A chain like `git push && gh pr create` that fails on the create would still hand the error to the push. The hook payload carries no per-segment exit status, so a segment rule cannot close that gap.

This means a conflict/rejection indicator or generic failure text from a later `gh` step in a chain can be misclassified as belonging to an earlier `git` step (e.g. incorrectly flagging a successful `git push` as `push_rejected`, or missing that `gh pr create` actually failed for `git-activity-recorder.ts`'s own `record.success` check, which is command-level, not segment-level).

## Root cause

The hook payload (`ToolCallRecord`) carries one `success`/`error` pair for the entire shell command, not per shell segment. There is currently no way to know which specific segment in a chain produced the observed error, so `classifyGitSegments` (`src/metrics/git-event-classifier.ts`) approximates by handing the error to the last segment.

## Proposed fix

This likely requires either:
- Capturing per-segment exit status in the hook collector (`src/hooks/collector-script.ts`) so each segment's success/error can be attributed correctly, or
- A safer heuristic than "always the last git segment" for the case where a non-git segment (like `gh`) follows the last git segment in the chain.

Filed for tracking since the current PR (#663) explicitly left this as a known gap rather than attempting a fix.

dyk1454683243-sudo:
## Summary

- Fixes #672.
- `classifyGitSegments` used to hand the command-level `record.error` to the last *git* segment. A chain like `git push && gh pr create` that fails on create therefore classified the successful push as `push_rejected` whenever the `gh` error text looked like a rejection.
- Safer heuristic: when the command-level record is a failure **and** a later `gh` segment follows the last git segment, withhold that error from every git segment. Pure-git chains still attribute conflict/rejection text to the last git segment (`&&` stops at the first failure).
- No collector change. `ToolCallRecord` still has one success/error pair; this only changes how that pair is applied. Per-segment exit status would be more precise but is not needed for the filed case.

Out of scope: #486 (engaged/active time), #671, and open PRs #776 / #779–#788.

## Test plan

- [x] `npx jest -- src/metrics/git-event-classifier.test.ts src/metrics/git-activity-recorder.test.ts` (72 passed)
  - `git push && gh pr create` failure-on-create does **not** classify as `push_rejected`
  - `git fetch && git push` rejection still lands on the last git segment
- [x] `npm run build`
- [x] `npm test` — 7121 passed. Two failures are pre-existing and outside this diff: `api-handler` DST bucket (passes with `TZ=America/New_York`, as CI pins) and a `workflow-watcher` partial-write flake.
- [x] `npm run lint`
- [x] `npm run format:check`


dyk1454683243-sudo:

